PR #97, opened 2022-04-13
Remove tests that fail in the new brms version

paul-buerkner:
Hi all,

I am in the process of submitting a new brms version to CRAN in which I, among others, have reworked the prior bound interface. As a result, four of your tests are now failing with the new brms version, which prevents (or at least delays) the new brms release to CRAN. 

In this PR have commented out these four tests, which should make sure that your package will not have test failurs in both the current and future CRAN version of brms. 

If this PR is okay with you, would it be possible for you to make a new CRAN release soon? Please apologize for the inconvenience caused.

Paul


dbarneche:
Hi Paul, my sincerest apologies for missing this PR. Luckily we ended up fixing this issue independently, so I think all is good now. Thanks very much for taking the time to warn us about the change, and apologies again. All the best, D

paul-buerkner:
No worries! Thanks for fixing it!

PR #99, opened 2022-09-30
Explore priors

beckyfisher:


beckyfisher:
This branch makes some minor edits to the priors, making them slightly less informative. It also has some major updates to the article.

PR #137, opened 2023-03-14
fix rhat generic-method conflict

paul-buerkner:
This PR fixes an R check warning that will occur once brms 2.19 is on CRAN, which will tentatively happen in the next few days.


beckyfisher:
Hi Paul, thanks for being so proactive with this required change. We have updated the package generic methods in commit #24b0e50 and submitted to cran just now.

PR #155, opened 2024-03-20
fix test setup to work with the latest brms

paul-buerkner:
In the current brms github version (soon to be on CRAN), your test setup fails because it uses the internal `step` function of brms in one of its formulas. Due to a change in how non-linear formulas search for their variables, `brms:::step` is no longer in the search path and I have yet to find a good and safe approach to include it there. 

This PR provides a little fix for the resulting problem in your test setup.


beckyfisher:
This is now handled at the package level: step() was added to the toolkit and exported (commit 89edef72, in 2.1.2.0) for brms non-linear formula evaluation, so the test-setup workaround is no longer needed. Closing as superseded.

PR #158, opened 2025-08-07
Ignore order of attributes.

kalibera:
This fixes the tests to ignore the ordering of attributes. In principle, the ordering can be changed by various R functions and no particular order should be expected. The patch fixes problems found in tests using a recent update of na.omit.data.frame() that impacts also the ordering of attributes, but similar changes in R can happen at any time. This is also why identical() in base R by default ignores the ordering of attributes.


beckyfisher:
Superseded. It changes test-model.frame.R to use expect_setequal() so attribute order doesn't matter. master already solves the exact same problem via expect_identical(sort(...), sort(...)) (commit 6f6bd25d, "fixed flaky model frame test to accommodate idiosyncrasies among linux devel versions"). Same intent — order-independent attribute checks — just a different idiom. Closing it won't reintroduce the CRAN/R-devel failure it was meant to fix.

PR #163, opened 2026-06-30
Fix ecnsec

beckyfisher:
### Problem
`ecnsec()` errored for any model whose prediction newdata has more than one
column — i.e. binomial and beta-binomial models, which carry a `trials`
column. For example:

```r
ecnsec(aa.fit, nsec = 3.6)
#> Error in `names(x) <- value`:
#> ! 'names' attribute [2] must be the same length as the vector [1]
```

### Cause
The NSEC prediction point was built by hand as a single-column data frame and
then renamed with the model's full set of predictor column names:

```r
newdat_nsec <- data.frame(x = c(max(x_vec), nsec[1]))
colnames(newdat_nsec) <- x_name   # x_name has length 2 for trials models
```

This only works when newdata has exactly one column and when `xform` is
`identity` (the raw `nsec[1]` was used, ignoring any `xform`).

### Fix
Build the NSEC prediction newdata via the existing `newdata_eval()` helper
(already used for the main prediction grid in the same function), and apply
`xform()` to the supplied `nsec` value before locating the point:

```r
nsec_use   <- xform(nsec)
newdat_nsec <- newdata_eval(object, resolution = 2,
                            x_range = c(nsec_use, nsec_use))
```

This handles multi-column newdata correctly and honours `xform`.

### Tests
`ecnsec()` previously had no test coverage, which is why this regression went
unnoticed.

- `tests/testthat/test-ecnsec.R` — behaviour (return types, posterior) and
  argument validation (`prob_vals`, `hormesis_def`, `xform`).
- `tests/local/test-ecnsec.R` — regression tests on binomial, logit-link
  binomial and beta-binomial (`trials`) `bayesmanecfit` fixtures, which
  reproduce the original error.

### Other
Fixed the roxygen example, which called `ecnsec()` without the required
`nsec` argument.

### Verification
On a real binomial fit, `ecnsec(aa.fit, nsec = 3.6)` errors on `dev` and
returns sensible values on this branch (median ≈ 1.65%). The fast test file
passes (11/11).


beckyfisher:
Hi Diego,
I'm trying to clean a few things up in bayesnec - if you want to review in detail feel free otherwise just have a quick check over? Note that some things are failing - I'll work through that separately. This branch was just to fix the ecnsec (I needed it for another project).

Issue #173, opened 2026-07-31
Guidance against normalising to control before dose-response analysis (Ritz et al. 2026)


beckyfisher:
## Summary

Ritz, Gerhard & Streibig (2026), *Better alternatives than normalizing to control: case studies with algae toxicity and dose-response analysis*, Environmental and Ecological Statistics 33:35–55, [doi:10.1007/s10651-025-00698-y](https://doi.org/10.1007/s10651-025-00698-y), shows that the standard ecotoxicology practice of normalising a response to the control mean before dose-response analysis produces **biased effective-dose estimates and standard errors that are too small**.

`bayesnec` already implements the paper's recommended alternative, but nowhere says so, and nothing warns a user who arrives with pre-normalised data. This issue proposes closing that documentation gap.

## The problem

Growth inhibition is conventionally derived as `y0i = 1 - yi / ȳcon`, where `ȳcon` is the mean of 3–6 control replicates. Because every observation is divided by the *same random quantity*, the normalised values are positively correlated. Standard analysis ignores that.

Two consequences:

1. **Too-small standard errors**, because the uncertainty in `ȳcon` is discarded.
2. **Biased effective doses.** By Jensen's inequality applied to `y → 1/y`, `E(1 - yi/ȳcon) ≤ 1 - E(yi)/E(ȳcon)`, so the inhibition trend is biased downwards and the resulting ED estimates are biased **upwards**.

Their simulation (Table 4, ED10, 6 controls):

| approach | bias | CV | coverage of nominal 95% CI |
|---|---|---|---|
| normalisation | 6.8% | 26.4% | 0.90 |
| normalisation + modelled correlation | 4.7% | 23.6% | 0.92 |
| **growth-rate based** | **2.1%** | **12.7%** | **0.95** |

Roughly half the CV is an artefact of normalising, and nominal 95% intervals cover at 88–91%.

The key result is that **nothing is lost by not normalising**: the ED for a p% increase in inhibition is identical to the ED for a p% decrease in growth rate.

## bayesnec already does the right thing

The paper's recommended estimand (their eq. 8) is

```
f(EDp, β) = (1 - p/100) · f(0, β)
```

a decline relative to the *fitted* control value. That is exactly what `ecx(type = "absolute")` computes (`R/ecx.R:295`):

```r
ecx_x_absolute <- function(y, ecx_val, x_vec) {
  range_y <- c(0, max(y, na.rm = TRUE))
[...]

PR #174, opened 2026-07-31
Add the zero_inflated_beta family

beckyfisher:
**Closes:** #183

**Stacked on #172** (which is stacked on #169 → #168). Review that stack first; the diff here is only the `zero_inflated_beta` work.

Extends the two-block machinery from #172 to proportional responses on (0, 1) that also contain exact zeros — the algal growth-rate case, where growth is expressed against a ceiling and some replicates fail outright.

---

## The short version

`zero_inflated_beta` is structurally the same model as `hurdle_gamma`. Zero-inflation differs from a hurdle only when the base distribution can itself emit zeros, and neither the Gamma nor the Beta can — brms's own generated Stan for `zero_inflated_beta` is the hurdle form, with no `log_sum_exp`. So both routes from #172 carry over unchanged:

```r
# Route A - joint
bnec(y ~ crf(log_x, "nec3param"), data = dat, family = "zero_inflated_beta")

# Route B - factorised
bnec_hurdle(y ~ crf(log_x, "all"), data = dat)   # picks Beta automatically
```

The only real difference is that brms names the second block `zi` rather than `hu`, giving `zitop`, `zinec` and so on. That name was hardcoded in eight places; it is now carried through from a lookup:

```r
hurdle_fams    <- c(hurdle_gamma = "hu", zero_inflated_beta = "zi")
hurdle_mu_fams <- c(hurdle_gamma = "Gamma", zero_inflated_beta = "beta")
```

## Three places the families genuinely differ

Beyond the renaming:

- **`check_models()`** — both blocks are 0-1 bounded here, so `nechormepwr01` is **kept**; it is the equation designed for that range. `hurdle_gamma` drops it via its zero-bounded mu block. Tested both ways.
- **`check_data()`** — zeros are preserved as the signal, but **ones still need nudging**, since Beta's support is open at 1. Easy to miss and it would have failed at sampling time.
- **`bnec_hurdle()`** — `family_growth` now defaults to `NULL` and is chosen from the **non-zero subset**, so proportional data gets Beta rather than the previously hardcoded Gamma. Choosing from the whole response would read the zeros as part of a continuous distribution rather than as the hurdle they are.

## Verification

Simulated algal-style data (84 observations, 7 concentrations, 27 zeros), both routes:

| | Route A (joint) | Route B (factorised) | true |
|---|---|---|---|
| NEC, mu block | 1.64 [0.99, 2.74] | 1.65 [0.98, 2.92] | 1 |
| NEC, zi block | 4.67 [3.03, 9.14] | 5.26 [3.05, 9.35] | 3 |
| `top` | 0.742 | — | 0.75 |

The two routes agree, `top` recovers well, and `posterior_epred()` equals `mu * (1 - zi)` exactly (max abs diff 0). `ne_type` is `NEC`, confirming the `extract_pars` anchoring fix from #172 works with the `zi` prefix too, and `bnec_hurdle()` correctly reports "Fitting the growth component (57 survivors of 84) with a beta distribution".

**Honest caveat on the simulation:** both thresholds are biased upward with the true values sitting at the lower credible bound, and the fits produced 62 and 70 divergent transitions at `adapt_delta = 0.99`. That is a property of the toy design — 84 observations, a threshold model, and zeros confined to the top two concentrations give the threshold parameters very little to work with — rather than evidence about the implementation, which is exercised by the unit tests. But it should not be read as a demonstration of good recovery.

## Tests

657 passing, 0 failures, 0 errors. 37 new tests in `test-hurdle_family.R` (93 total in that file) covering the `zi` naming throughout, the `link_zi` guard, the `nechormepwr01` divergence between the two families, zeros-kept-but-ones-nudged in `check_data()`, and `zi`-prefixed initial values.

## Documentation

`?bnec` now describes both families together, explains when to reach for each, and states the constraint that matters for this one: **where a proportional response was obtained by dividing through by a maximum, that divisor must be a constant fixed in advance — a design ceiling or a historical value — not one computed from the dataset being analysed.** Dividing by an estimated quantity induces correlation across all observations, biases effective doses upward and understates their uncertainty (Ritz et al. 2026, cited in `@references`). See #173, which covers that criticism in full and is deliberately left out of scope here.

## Not included

- The `bnec_hurdle` name is now slightly off for the zero-inflated case. Left alone rather than churning a function introduced one PR ago; worth revisiting if a third family arrives.
- `hurdle_lognormal`, which needs its own handling because lognormal's `mu` is on the log scale despite brms reporting an identity link.


beckyfisher:
Adds the full method set for `bayesnechurdlefit`, plus the `Route A`/`Route B` rename to `joint`/`factorised`.

## Why

An audit of every package method against a fitted `bayesnechurdlefit` found **4 of 28 working**. Three of the failures were silent rather than errors:

| method | before |
|---|---|
| `summary()` | fell through to `summary.default`, returning a table of the list structure |
| `fitted()` | returned `NULL` |
| `pull_out()` | returned the object unchanged, with "Returning original object" |

## What changed

Methods split into two groups.

**Curve methods** compute from both components and take `which = "combined"` (default), `"growth"` or `"survival"`; `plot()` and `autoplot()` also accept `"all"`. Combined is the default throughout because it is what a hurdle model exists to estimate — a user wanting only the components would fit them separately with `bnec()` and not reach for the wrapper.

**Fit methods** — `rhat()`, `check_chains()`, `check_priors()`, `pull_brmsfit()`, `pull_out()`, `dispersion()`, `model.frame()` — have no combined analogue and return one result per component.

`amend()`, `update()`, `c()` and `+` apply to both components so the pair stays in step. `update(newdata =)` re-splits the data, or zeros would reach the growth component as small responses rather than as the hurdle.

## Three implementation notes

- **`posterior_predict(which = "combined")` simulates the process**, not the mean: a survival outcome is drawn per posterior draw, so non-survivors contribute exact zeros and the draws carry the spike the data have. `posterior_epred()` deliberately does not, being a mean.
- **The survival panel plots observed proportions**, not raw points — individual 0/1 values against a probability curve are unreadable.
- **The three curves are panelled, not overlaid.** Growth is in response units, survival is a probability; one pair of axes would be a dual-axis chart and misrepresent relative magnitude. Each panel then carries one series, so no colour encoding or legend is needed.

## A trap worth recording

`pull_out`, `dispersion` and `compare_fitted` are **plain functions, not generics**. An S3 method for them registers in `NAMESPACE` and silently never dispatches. My first pass added `pull_out.bayesnechurdlefit`, which registered, was reported "OK" by the audit, and did nothing — reintroducing the exact silent failure it was meant to fix. All three now branch internally, and a test pins the distinction so a future S3 method for them cannot register inert.

## State

- **28 of 29 methods pass** on both the factorised and joint objects
- The one remaining error is `amend()`, which affects ordinary single-model `bayesnecfit` objects identically — filed separately as #176
- **890 tests passing, 0 failures, 0 skips** under `NOT_CRAN=true`; 58 new structural tests, none requiring sampling

Related: #175 (vignette, blocked on a case study), #176 (`amend()` on single fits).

Issue #178, opened 2026-08-01
Vignette figures are PDFs embedded in HTML articles


beckyfisher:
## Summary

Every vignette is authored with `dev = "pdf"` but rendered as `rmarkdown::html_vignette`.
Knitr therefore writes each figure as a PDF and `pandoc` embeds it with

```html
<embed src="vignette-fig-exmp1-binom-1.pdf" title="plot of chunk exmp1-binom"
       width="60%" type="application/pdf" />
```

Browsers render an `<embed>` of a PDF by loading their built-in PDF viewer, so each plot
appears inside a dark viewer panel with its own page-number box, zoom controls and
annotation toolbar, at whatever size the viewer chooses rather than the requested `width`.

This affects the **published documentation**, not just locally built vignettes.

## Evidence

Counted from the live site:

| article | figures embedded as PDF |
|---|---|
| `example1` | 6 |
| `example2` | 6 |
| `example3` | 1 |
| `example4` | 5 |
| `example2b` | 0 — but see below |
| `example5` | 0 (no figures) |

The PDF files themselves are served correctly (`200`, `application/pdf`), so nothing is
missing; the content is fine and only the presentation is wrong.

## A second, worse variant in `example2b`

`example2b` bypasses knitr for its two theoretical-curve figures and hand-writes the HTML
(`vignettes/example2b.Rmd.orig` lines 98 and 248):

```html
<img src="vignette-fig-exmp2b-theoretical_ecx_curves.pdf" title="" alt="" width="100%"/>
```

An `<img>` element pointing at a PDF is not something browsers render at all, so these two
figures are likely blank rather than merely awkward. Worth confirming visually — it is the
one case here that loses the figure entirely rather than wrapping it in a viewer.

## Cause

`dev = "pdf"` is the right choice for a PDF vignette and the wrong one for an HTML vignette.
It is set identically in five of the six `.Rmd.orig` files:

```r
knitr::opts_chunk$set(
  ...
  dev = "pdf",
  eval = identical(Sys.getenv("NOT_CRAN"), "true")
)
```

## Fix

Switch the device to a raster or vector format the browser renders inline:

```r
  dev = "png",
  dpi = 150,
```

`html_vignette` base64-embeds the result, so the figure arrives as
`<img src="data:image/png;base64,...">` and needs no separate file at serve time. `svg` is
the alternative if scalability matters more than file size.

Two things have to change together:

[...]

PR #186, opened 2026-08-06
Guidance and detection for responses normalised to the control

beckyfisher:
Closes items 1–3 of #173.

Ritz, Gerhard & Streibig (2026) show that normalising a response to the control
mean before dose-response analysis divides every observation by the same random
quantity: the values become correlated, the uncertainty in the divisor is
discarded, and by Jensen's inequality the effective doses come out biased
upwards. `bayesnec` already implements their recommended alternative —
`ecx(type = "absolute")` measures the decline relative to the *fitted* control
value, evaluated within each posterior draw — but said so nowhere, and nothing
warned a user who arrived with pre-normalised data.

## 1–2. Documentation

- **`?bnec`** carries the guidance: a "supply the raw response" section, why
  nothing is lost by not normalising (`ecx(type = "absolute")` is already the
  recommended estimand), why dividing by the observed maximum is worse than
  dividing by the control mean, and the Beta-family divisor rule. The existing
  `zero_inflated_beta` paragraph defers to it rather than repeating it.

  `?ecx` and `?nsec` deliberately carry **none** of this. Both take a fitted
  object, so by the time either is called the response has already been chosen
  and advice about preparing it is unactionable. `bnec()` is where the data
  enters.
- **`vignette("example1")`** gains a *Preparing the response* section covering
  the same ground with the Jensen's-inequality argument and the estimand
  written out. Prose only, no new chunks, so `.Rmd` and `.Rmd.orig` stay in
  sync without a re-knit.

## 3. Detection

`bnec()` now checks both signatures, once per call rather than once per model:

| practice | trace |
|---|---|
| divided by observed maximum | `max(y) == 1` attained by exactly one observation |
| divided by control mean | `mean(y[x == min(x)]) == 1` |

The issue asked for this to be prototyped against real data first, on the
grounds that genuine proportion data often sits at `1`. That turned out to
matter a great deal. **Unguarded, the divided-by-maximum check fires on up to
38% of simulated fertilisation assays**, purely because one replicate recorded
every individual as alive.

The fix is `on_rational_grid()`: count-derived proportions are `k/n` for a small
`n`, a continuous response divided by a continuous maximum is not. With it:

- **0 false positives** across 2000-draw simulations of eight genuine-data
  scenarios, and 100% detection of continuous responses divided by their maximum
  or by the control mean;
- **0 false positives** in a sweep of 1269 columns and column pairs from 33 real
  datasets — while correctly flagging four that *are* normalised
  (`lum_plates.csv`, `CoralColour.csv`, `Monique_microtox.csv`,
  `tmp_test_dat.csv`), each verified by hand.

Two deliberate misses are documented: `y / (max(y) * 1.01)` leaves no trace at
all, and a count proportion divided by its own maximum stays on a rational grid.
Both fail silent, which is the safe direction for a diagnostic message.

Evidence and both reproduction scripts are in `notes/normalisation_detection.md`
and `notes/scripts/`.

Also confirmed, since it is the obvious question: the Beta example in
`example1` uses genuine maximum quantum yield (values ≈ 0.57), not a normalised
rate. The issue's claim that `bayesnec` does not do this anywhere holds.

## Not in scope

The `beta_ub` custom-family plan in the later issue comments is a separate,
multi-phase piece of work gated on a standalone Stan prototype. Nothing here
touches it.

## Checks

`devtools::test()` — 787 pass, 0 fail, 0 warn.
`R CMD check` — 0 errors, 0 warnings, 1 note (`.git`, an artefact of checking
from a git worktree, where `.git` is a file rather than a directory).

Based on `hurdle-vignette` and targeted at it, following the existing stack.


beckyfisher:
Rebased off `bdd6c748` (the #178 work, which moved to its own branch
`issue-178-vignette-figs` and out of `hurdle-vignette`). The PR now carries only
its own commit: 16 files against `hurdle-vignette`, version 2.1.3.8.

Re-verified after the rebase — `devtools::test()` 787 pass / 0 fail / 0 warn,
`R CMD check` 0 errors / 0 warnings / 1 note (`.git`, an artefact of checking
from a git worktree).

PR #187, opened 2026-08-06
beta_ub: an upper-bounded beta family with an estimated ceiling

beckyfisher:
Implements the `beta_ub` family proposed in #173, through Phases 0–9 of the plan
in that issue's comments. One commit per phase, so any stage can be rewound to
independently.

Based on `hurdle-gamma-impl`, as the descoping comment says it can be: `delta`
is a scalar nuisance parameter, not a block with its own curve, so none of the
two-block machinery on `zi-beta-impl` is needed.

## Why

Growth-rate and cell-yield endpoints are compressed against an experimental
ceiling near the control, so an unbounded family misstates the control variance.
The conventional fix — divide by `max(y)`, fit a Beta — buys a bounded variance
function at the cost of dividing every observation by an extreme order statistic.
This family gives the ceiling its own parameter instead:

```
mu(x)      the existing bayesnec equation, natural scale
y          ~ U * Beta(mu/U * phi, (1 - mu/U) * phi)
Var[y | x] = mu (U - mu) / (1 + phi)
```

`U = ymax + delta` with `delta > 0`, because `brms::custom_family()` cannot
express a data-dependent lower bound — but **the prior goes on `U`**, not on
`delta`. The map is a pure location shift, so `U ~ normal(U_loc, U_scale)`
truncated to `U > ymax` is exactly `delta ~ normal(U_loc - ymax, U_scale)` with
`lb = 0`. A prior placed directly on `delta` would centre the ceiling just above
the sample maximum, where it moves with `n` and with the noise in one extreme
value.

The response reaches `brms` untouched — no divisor anywhere.

## The phases

| | |
|---|---|
| 0 | Standalone Stan prototype. Three decisions settled with numbers |
| 1 | `beta_ub()` custom family, Stan lpdf/rng, brms post-processing methods |
| 2 | Registration, data checks, model-set routing (Phase 3 dropped per the descoping comment) |
| 4 | Ceiling prior, `U_loc`/`U_scale` on `bnec()`, initial values |
| 5 | `bnec()` end to end |
| 6 | Predictions, ECx, NSEC, plotting |
| 7 | Model averaging and QC |
| 8 | SBC and the coverage study |
| 9 | Documentation, NEWS, packaging |

## Validation

**SBC passes.** 256 replicates, 99 thinned draws each: ranks uniform on all five
parameters (chi-square p 0.287–0.975), 0 divergences across 255 fits.

**Coverage study**, mirroring the Ritz et al. (2026) design, 150 replicates per
cell, three arms differing only in the ceiling — `max(y) * 1.01`, the true value
as a constant, or estimated.

The predicted pattern appears, but at `top` rather than at ECx:

| control n | arm | `top` bias % | width | coverage |
|---|---|---|---|---|
| 3 | max(y) | +0.91 | 0.0717 | **0.880** |
| 3 | fixed | 0.00 | 0.0810 | 0.947 |
| 3 | estimated | −0.02 | 0.0830 | 0.960 |

Nominal 0.95, MC SE 0.018. Coverage of EC10, EC50 and NEC is fine in **all three**
arms (0.913–0.980), which is more specific than the plan expected. ECx is a
relative decline from the fitted `top`, so a mis-stated ceiling largely cancels
out of it; `top` is what it actually distorts. The gain here is an honest
interval at the top of the curve, not a different toxicity estimate.

## Three limits, documented rather than papered over

- **The data only inform `U` once the control reaches ~70% of it.** At
  `top/U = 0.2` the posterior for `U` equals its prior to three decimal places.
  Hence the refusal to default `U_loc`/`U_scale` silently.
- **A response that decays past the floating-point floor cannot be fitted.**
  `mu * phi` underflows and `beta_lpdf` goes degenerate. It presents as a
  failure to find initial values rather than a clear message, so `?beta_ub`
  says so.
- **Supplied inits are load-bearing.** Random initialisation aborted 6 of the
  first 16 SBC replicates, and the ones it aborts depend on the parameters.

## Two bugs worth flagging

**`brms` reports every custom family as `family$family == "custom"`**, keeping
the real name in `family$name`. Every existing switch on the tag would have
fallen through. New `family_tag()` resolves custom families to their name and
returns `family$family` otherwise — no behaviour change for the built-ins.

**`Roxygen: list(markdown = TRUE)` makes `\%` wrong in roxygen comments.** It
emits `\\%` into the Rd, and in the middle of a long `@details` block it
silently dropped the entire section: `bnec.Rd` came out with an empty
`\details{}` and `R CMD check` reported only "Dropping empty section". Caught by
reading the generated Rd rather than trusting the check summary. `R/data.R`
carries the same latent issue from an earlier commit (`nassarius.Rd` line 69),
left alone here.

## Checks

`devtools::test()` — 701 pass, 0 fail, 0 warn (fast suite; 71 further fitting
tests in `tests/local/test-beta_ub.R`).
`R CMD check` — 0 errors, 0 warnings, 1 note (`.git`, an artefact of checking
from a git worktree).

Evidence: `notes/beta_ub_phase0.md`, `notes/beta_ub_phase8.md`, and the
reproduction scripts in `dev/`.


beckyfisher:
## Closing: `beta_ub` does not earn its place

Testing the premise behind this family against real data rather than simulation, **it does not
hold**, and the gain it does deliver on its best case comes from somewhere else. Closing without
merging. The branch `beta-ub-impl` stays; nothing is deleted.

Follow-up: #191, which subsumes the motivation.

### The premise was testable, and it failed

`beta_ub` was proposed on the grounds that growth-rate and cell-yield endpoints are *"compressed
against an experimental ceiling near the control"*. Under this family the ceiling is not a free
extra parameter — fixing it fixes the *shape* of the residual distribution. With `m = top/U`,
`CV² = (1−m)/(m(1+φ))`, so a measured control CV pins `φ` once `m` is chosen and the control
skew follows with no freedom left:

```
skew(Beta(a,b)) = 2(b−a)√(a+b+1) / ((a+b+2)√(ab)),   a = mφ,  b = (1−m)φ
```

Tested on four algal specific growth-rate tests, five *Hydra* population growth-rate tests,
twenty algal cell-yield tests and the shipped `herbicide` PAM data, with parametric-bootstrap
nulls reproducing the within-group centring exactly:

| endpoint | n control | observed skew | 5% null cutoff | compatible ceiling |
|---|---|---|---|---|
| Specific growth rate | 60 | −0.315 | −0.497 | U ≥ 1.13 × top |
| Population growth rate | 20 | +0.040 | −0.665 | U ≥ 1.23 × top |
| Cell yield | 160 | −0.238 | −0.309 | U ≥ 1.24 × top |
| PAM Fv/Fm | 83 | −0.274 | −0.404 | U ≥ 1.13 × top |
| *Number of Young* (positive control) | 100 | *−1.526* | *−0.396* | *U ≥ 1.64 × top* |

The last row is a genuinely skewed count endpoint, detected at −1.53 — the test has power;
growth rate simply does not trigger it.

**For growth rate this closes decisively.** OECD TG 201 Annex 5 states the response is
referenced to controls *"showing maximum response for the conditions imposed — with light and
temperature as primary determining factors"*. The control **is** grown at µ_max, so any
biologically honest `U` is ~1.00–1.05 × top. The data require U ≥ 1.13 × top. **The
biologically meaningful ceiling is precisely the one the data exclude**, and the ceilings that
survive (1.13–1.43 × top, the upper bound being this family's own `top/U ≥ 0.70` identifiability
floor) have no biological referent — at ≥ 4 control SD away the Beta is numerically Gaussian.

Separately, specific growth rate is on the real line (observed range −1.99 to +1.49 across these
tests), so no bounded family applies to it at all. The `alga` data only ever looked Beta-like
because zeros had been substituted for negative growth and the response divided by `max(SGR)`.

### On its best case, the gain is a variance function — not the ceiling

PAM Fv/Fm is the one endpoint here where a ceiling is defensible: inverting the control skew
gives a best-matching U = 0.793 (95% lower bound 0.720), bracketing the physiological Fv/Fm
maximum of 0.75–0.83. So the family was given its most favourable case, with `U_loc = 0.80`,
`U_scale = 0.05` elicited from the literature rather than from the data.

Four models, identical mean curve and curve priors, differing only in dispersion:

| | diuron (n=72) | tebuthiuron (n=108) |
|---|---|---|
| M2 Beta, `log(φ) = c0 + c1·log(µ) + c2·log(1−µ)` | **0.0** | −4.5 (se 2.6) |
| M1 Beta, `φ ~ log₁₀(dose)` | −1.8 (se 1.3) | −9.2 (se 3.2) |
| M3 `beta_ub`, elicited ceiling | −2.4 (se 1.9) | **0.0** |
| M0 Beta, constant `φ` | −3.8 (se 2.8) | −15.6 (se 4.7) |

`beta_ub` beats a **constant-φ** Beta on tebuthiuron (15.6, ~3.3 SE) but is **indistinguishable
from a Beta with mean-dependent dispersion** (4.5, ~1.7 SE). On diuron it places third with
every difference inside noise. Give the Beta a variance function and the advantage disappears.

Not surprising in hindsight: **`beta_ub` is a hard-coded variance function.** `Var = µ(U−µ)/(1+φ)`
asserts `sd ∝ √(µ(U−µ))`. Estimating the variance function directly gets the same flexibility
without a custom Stan family or an elicited ceiling prior — that is #191.

### The implementation is sound; it just isn't needed

I suspected the family might be exploiting its data-dependent support boundary
(`U = ymax + delta`, with a `−log(U)` Jacobian rewarding small `U`). **It is not.** On data
simulated from a plain Beta with no ceiling whatsoever:

```
        elpd_diff se_diff
Beta     0.0       0.0
beta_ub -0.9       0.7
```

Beta wins slightly, as it must, and the posterior `U` stayed at its prior (0.797) rather than
collapsing onto `ymax` (0.626). Pareto-k ≤ 0.26. The SBC and coverage work in this PR stands —
the family does what it says. No endpoint tested needs it.

### Practical limits, for the record

- **Zeros.** 3 of the 7 `herbicide` datasets (ametryn 12, irgarol 10, hexazinone 1) contain
  zeros and `beta_ub` rejects them outright, so it can fit only 4 of 7 without `beta_ub_lcdf`.
- **ECx barely moves.** diuron EC50 2.40 → 2.26; tebuthiuron 100.2 → 101.2 — consistent with
  this PR's own coverage study, which found the benefit confined to `top` intervals rather than
  toxicity estimates.

### One methodological note for anyone repeating this

`brms::Beta()` defaults to `link = "logit"`, while `bnec()` forces `link = "identity"` on every
family via `validate_family()` so that `top`, `bot` and `nec` stay on the natural response
scale. A hand-rolled `brms` comparison must pass the link explicitly, or the concentration-
response curve is fitted to the *logit* of the mean and the comparison is meaningless. The
symptom is an implausibly large `elpd_diff` — tens to hundreds of units on a small dataset —
that still samples cleanly with healthy R-hat. Every number above uses identity links throughout.

### Stack

`beta-ub-impl` is the tip of the stack and nothing is built on it — verified by ancestry:
`cens-impl`, `issue-173-guidance` and `issue-178-vignette-figs` are all *contained in*
`beta-ub-impl`, and `beta-ub-impl` is contained in nothing. Closing this needs no rebasing.
#186 becomes the top of the stack; branch from `issue-173-guidance` next.

Issue #188, opened 2026-08-06
bnec_hurdle() rejects all aterms, so a censored response cannot be fitted as a hurdle


beckyfisher:
`bnec_hurdle()` refuses any `aterm` on the response. This blocks the combination
that #181 was raised for: a growth endpoint that is both zero-bounded with
structural zeros (deaths) and left-censored at the recording resolution
(survivors measured below the resolution limit).

```r
library(bayesnec)   # 2.1.3.8, branch cens-impl
d <- data.frame(x = rep(c(0, 1, 2, 3), each = 10),
                y = c(rep(2, 20), rep(0.5, 10), rep(0, 10)))
d$y[c(3, 7)] <- 0.05
d$cens <- ifelse(d$y == 0.05, "left", "none")

bnec_hurdle(y | cens(cens) ~ crf(x, "nec3param"), data = d)
#> Error : bnec_hurdle requires a plain, untransformed response on the left of
#> the formula, because zero values in it are used to identify deaths.
#> You supplied "y | cens(cens)".
```

It is not specific to `cens()` — `hurdle_response_var()` tests the whole LHS, so
every aterm fails the same way:

```r
d$w <- 1
bnec_hurdle(y | weights(w) ~ crf(x, "nec3param"), data = d)
#> Error : bnec_hurdle requires a plain, untransformed response ...
#> You supplied "y | weights(w)".
```

## Why this is the case #181 exists for

The second comment on #181 argues the hurdle interaction is the important one,
and this is where it lands. A dead individual is a structural zero belonging to
the Bernoulli component; a living individual measured below the resolution limit
is a censored observation belonging to the response component. Both are zeros or
near-zeros in the same column, and only a two-part model with a censored response
component can tell them apart. That model is currently unreachable through
`bnec_hurdle()`.

The workaround is to make the two `bnec()` calls directly and assemble a
`bayesnechurdlefit` by hand. That works — `ecx()`, `nec()` and
`posterior_epred()` all behave, for `which = "growth"`, `"survival"` and
`"combined"` — which is itself evidence that nothing but the guard is in the way.

## Suggested change

The guard is stricter than the constraint it protects. What
`bnec_hurdle()` actually needs is that the *response variable* be a bare,
untransformed name, because its zeros identify deaths. An aterm alongside it does
not threaten that.

1. Have `hurdle_response_var()` extract the response from the LHS and validate
   only that part, rather than requiring the whole LHS to deparse to a bare name.
[...]

PR #189, opened 2026-08-06
Support the cens() aterm for censored responses

### LINKED ISSUE #181: Support the cens() aterm, for responses left-censored at the recording resolution
## Summary

Responses that are rounded or truncated at the recording resolution are
left-censored, not zero. `brms` expresses this with the `cens()` aterm, but
`bayesnecformula` validates only `trials()` and `weights()`, and
`model.frame()` drops everything else — so the censoring information never
reaches the fit. The current fallback is a point substitution that stands in
for an interval.

`?bayesnecformula` already says:

> **NB:** `aterms` other than `trials()` and `weights()` are currently omitted
> from `model.frame` output. If you need other `aterms` as part of that output
> please raise an issue on our GitHub page.

This is that issue, for `cens()` specifically.

## Motivating case

The shipped `herbicide` dataset [@jones2003], used as a counter-example in
`vignette("example6")`. `fvfm` contains 23 zeros that look like candidates for
zero-inflation but are not:

```r
nonzero <- herbicide$fvfm[herbicide$fvfm > 0]
c(resolution = min(diff(sort(unique(nonzero)))),
  smallest_nonzero = min(nonzero))
#>       resolution smallest_nonzero
#>             0.01             0.01
```

Every value is a multiple of the recording resolution and the smallest non-zero
value is exactly one unit of it, so the gap ratio is 1.0: a recorded `0.00`
means the reading fell below the rounding bound, not that a distinct process
occurred. Zeros appear only in the compounds whose curves drop below that
floor — diuron, which levels off above it, records none.

The correct model is a left-censored likelihood at the rounding bound rather
than either a hurdle or a substitution. Note the bound depends on the recording
convention: `[0, 0.005)` if values were rounded to nearest, `[0, 0.01)` if
truncated. `cens()` would make that assumption explicit instead of leaving it
implicit in a nudge.

## Current behaviour (verified)

`bnf()` accepts the aterm and warns, but the information is then dropped:

```r
library(bayesnec)
data(herbicide)
d <- subset(herbicide, herbicide == "diuron")
d$y <- d$fvfm; d$x <- d$concentration

f <- bnf(y | cens("left", 0.005) ~ crf(x, "nec3param"))
#> Warning: You have specified brms special aterms other than trials and
#> weights. bnec may yield unexpected model fits, proceed at your own risk.

names(model.frame(f, data = d))
#> [1] "y" "x"                      # the censoring term is gone
names(attr(model.frame(f, data = d), "bnec_pop"))
#> [1] "y_var" "x_var"              # and is not registered as a population var
```

So the fit silently becomes an ordinary uncensored one.

## Why the existing fallback is not equivalent

`check_data()` shifts zeros to `min(y[y > 0]) / 10` for Gamma and beta
responses (`R/check_data.R`, lines ~57 and ~70). For `herbicide` that is 0.001,
which does fall inside `[0, 0.005)`, so it is defensible in this particular
case. But:

* it asserts a point where the data support only an interval, and the fitted
  curve is pulled toward that point;
* landing inside the true interval is a coincidence of `min(y[y>0])/10` here,
  not a property of the rule — with a coarser resolution or a different
  smallest non-zero value it need not;
* it gives no way to state which rounding convention was assumed.

Related: #93 (auto-correcting the shifts `check_data()` applies at 0 and 1).

## What would be needed

1. Carry `cens()` through `bayesnecformula` validation and into
   `model.frame()` output, registering the censoring variable in the
   `bnec_pop` attribute alongside `trials_var` so it reaches `brm()` with the
   data.
2. Decide the interaction with `check_data()`: a censored response should not
   also be nudged off zero, in the same way the hurdle families are already
   exempted from that shift.
3. Confirm downstream behaviour — `posterior_epred()` for a censored fit, and
   whether `ecx`/`nsec` need anything, since both read off the fitted mean
   curve rather than the likelihood.

## Scope note

Raised while writing `vignette("example6")` (#175), where the `herbicide` data
is the counter-example for *not* reaching for a hurdle model. The vignette
currently states that this route is untested and that the existing nudge is
crude but defensible for these data; that text can be updated if `cens()`
becomes supported.

beckyfisher:
Closes #181.

Adds support for the `brms` `cens()` aterm, so a response that is rounded at the
recording resolution, reported below a limit of detection, or deliberately
coarsened at a boundary can be fitted as censored rather than substituted.
Previously `cens()` was accepted with a warning and then silently dropped,
leaving an ordinary uncensored fit.

## What changed

**The aterm is carried through to the fit.** Both `cens(indicator)` and the
interval form `cens(indicator, upper_bound)` now pass through `check_formula()`
into `model.frame()` output, and are registered in the `bnec_pop` attribute as
`cens_var` and `cens_y2_var` alongside `trials_var`.

**Censored rows are exempt from boundary shifting.** `bnec()` shifts zeros in
Gamma and Beta responses, and ones in Beta responses, away from the boundary.
The recorded value on a censored row is a declared bound, not a boundary
artefact, so shifting it would restate the bound the user chose. Uncensored rows
shift exactly as before. Where a row is censored at a value the family excludes
— left-censored at 0 under Gamma or Beta, right-censored at 1 under Beta —
`bnec()` now stops with an explanation, since the censored likelihood
contribution there is `F(0) = 0` and Stan would otherwise fail at initialisation
with nothing to point at.

**Bug fix in aterm parsing.** `bayesnecformula` lost any two-argument aterm
written on its own. The aterm chain was split by treating every length-three
call as an `a + b` pair, which destructured the aterm call itself, so
`y | cens(indicator, upper_bound)` parsed as though no aterm had been given.
Only formulas with a single two-argument aterm were affected.

**`check_formula()`** no longer reports `cens()` as an unvalidated aterm, and now
warns when a `cens()` term contains no variable — as in `cens("left")`, which
`brms` recycles into a declaration that every row of the response is censored.

## New dataset: `alga`

Growth inhibition tests on *Cladocopium proliferum* (7-day) and *Rhodomonas
salina* (3-day) against each of two contaminants, consolidated from four tests.
Contaminant identities are anonymised as A and B and exposure units are not
stated; nominal dose values preserve relative spacing, so toxicity estimates are
unaffected while the identity of the tests is not disclosed.

Two features are retained rather than cleaned away, because they are the point:

1. Cell density is counted to a resolution of 10, so a recorded density of 0
   means "below the counting limit", not "no cells". Growth rate is left-censored
   at the value implied by a density of 10, not undefined.
2. Where density was recorded as 0 the source substituted a growth rate of 0 —
   placing total loss of the population mid-range, above every genuinely negative
   value, and turning a monotonic concentration-response into a non-monotonic
   one. `sgr` is left exactly as supplied and `sgr_source` marks those rows.

Serves both #181 and #173.

## New vignette

*Censored responses* (`vignette("example7")`) — when a recorded value is a bound
rather than a measurement, why substitution biases the fit in a
concentration-dependent direction rather than only locally, how to encode
censoring so the response column carries the bound, and the distinction between
a censored zero and a structural one that `vignette("example6")` deals with.

## Notes

- Based on `hurdle-vignette`, so it sits alongside #185 and #186 in the stack.
- `tests/testthat/test-cens.R` adds coverage for the formula parsing, the
  `bnec_pop` registration, the boundary-shift exemption, and the excluded-value
  error.
- #188 (`bnec_hurdle()` rejects all aterms) is the follow-on this work exposed:
  a censored response cannot currently be fitted as a hurdle.


beckyfisher:
## Plan: rework the censoring material, and repurpose `example7`

Reviewing the rendered vignette, `example7` carries too much of the reasoning that
produced it rather than the reasoning a reader needs. Three specific problems, then a
plan.

### What is wrong with it now

**1. It generalises from the case that motivated it.**

> Because censored observations cluster at the high-concentration end, that upward pull
> is concentrated exactly where the response is changing fastest.

Only true for a declining response censored at a lower detection limit. An instrument
that saturates censors the *controls*; an increasing response such as mortality inverts
it entirely. We have no idea where a reader's censored values sit, and the vignette
should not assume.

**2. Case A does not motivate the feature.** The `herbicide` zeros are recording
artefacts — `fvfm` rounded to two decimal places, so a recorded `0.00` means the truth
fell in `[0, 0.005)`. That is formally interval censoring and is not *wrong*, but it is
23 of 580 values with an interval of width 0.005 on a response spanning 0–0.73. It
changes nothing. An example where the careful treatment and the crude one agree is an
argument for the crude one. `check_data()`'s existing nudge to `min(y[y > 0]) / 10` is
the right answer there and the vignette should say so plainly, as a *when not to
bother*, rather than presenting it as a worked case for censoring.

**3. One passage is right but reads as wrong.**

> Left-censoring at the bound then asserts that the truth lies in `(0, bound]` — the
> family supplies the lower limit and the censoring supplies the upper one.

This is correct — in left-censoring the bound is a ceiling on the unobserved truth, so
the interval really is `(0, bound]`. But "left-censoring at the bound" reads as though
the bound were a floor, and the sentence does nothing to dispel that. Needs rewriting
for clarity, not correcting.

**Case B — coarsening at a substantive boundary — is the only materially useful case**,
and it is where the algal data belongs.

---

### The plan

#### Step 1 — `example1`: a short *Censoring* section (on #186, **not** this PR)

**Correction to an earlier version of this comment:** this cannot land in #189. The section
is positioned after *Preparing the response*, which exists only on `issue-173-guidance`
(#186) — a branch *above* `cens-impl` in the stack. It has to go in #186 or higher.

Immediately after the *Preparing the response* material from #186, add a section called
simply **Censoring**. Content is essentially the current *How censoring is expressed*:
what the `cens()` aterm is, the three forms (`"left"`, `"right"`, `"interval"`), that the
response column carries the *bound* on censored rows rather than a substituted value, and
the one-line contrast between `F(c)` saturating and `f(c)` penalising a descending curve.

Short, mechanical, no case study. A reader who has censored data needs to know the syntax
exists and where to read more; that is all this belongs in `example1` to do.

#### Step 2 — land the dispersion sub-model (#191) — **before** the rewrite

The rewritten `example7` should demonstrate #191 rather than describe it, so #191 goes
first. This ordering is deliberate: writing the vignette against a feature that does not
exist yet would fix its shape before we know what the interface looks like.

#### Step 3 — repurpose `example7` as a growth-data case study

New scope: **modelling algal growth data**, using the four `alga` tests as case studies
and working through the issues in the order a practitioner meets them. This replaces the
current vignette rather than extending it; the useful parts of Case B survive as the
flooring section.

Proposed structure:

1. **The endpoint.** OECD TG 201 defines two response variables — average specific growth
   rate `µ = (ln X_j − ln X_i)/(t_j − t_i)` giving `ErCx`, and yield giving `EyCx` — and
   states growth rate is scientifically preferred because it does not depend on the
   control's absolute level or on test duration. Note that `sgr` and `density` in `alga`
   are the *same measurement* on two scales, not two endpoints, so choosing between them
   is choosing a variance model.

2. **Do not floor the negatives.** The `alga` growth rates go to −1.99; the source
   substituted `0` wherever density fell below the counting limit, which puts total
   population loss *above* every genuinely declining value and turns a monotone
   concentration–response non-monotone. TG 201 Annex 5 is explicit:

   > It is not recommended to assign a zero or small positive value to such negative
   > values because this distorts the error distribution.

   Three fits — substituted as supplied, censored rows dropped, censored properly with
   `cens()` — showing that substitution biases in a concentration-dependent direction.
   This is Case B from the current vignette, on real data instead of simulation.

3. **Do not normalise to the control mean or the observed maximum.** Cross-reference
   `example1` rather than repeating it; add only what is specific to growth rate — that
   `ecx(type = "absolute")` *is* `ErCx` algebraically, since `ErCx` is defined as
   `µ_T = (1 − x/100)·µ_C`. Note `type = "relative"` is **not** the ecotoxicological
   definition when `bot` is negative.

4. **Model choice.** A response on the real line needs `gaussian()`, and the model set
   must exclude the `zero_bounded` group. `bnec()` already drops those automatically, so
   `model = "all"` is safe — no need to hand-pick. Worth stating explicitly that this is
   *not* a requirement for a `bot` parameter: `neclin`, `neclinhorme` and `ecxlin` are
   `bot_free` yet unbounded below, and a linear decay is a plausible shape for a growth
   rate crossing into decline.

5. **Dispersion (#191).** The payoff. Variance is not constant across these curves —
   within-group SD on the algal tests runs from 0.005 near the control to 0.09 where the
   population collapses. Fit with and without a dispersion sub-model and compare, showing
   what it changes and, honestly, what it does not: ECx is defined on `mu`, so the point
   estimates move little and the *intervals* are what improve.

6. **When censoring is not worth it.** The `herbicide` rounding case, demoted to a short
   closing note: formally censored, immaterial in practice, existing nudge is fine. This
   keeps the honest negative result without letting it lead.

#### What is dropped

The bulk of the current *Substitution versus censoring* derivation, the `herbicide` Case A
worked fits, and the general claims about where censored values sit. The mechanics move to
`example1`; the argument survives in the flooring section attached to a case where it
actually bites.

---

### Sequencing

| | branch | |
|---|---|---|
| 1 | `issue-173-guidance` (#186) | `example1` *Censoring* section |
| 2 | new branch off `issue-173-guidance` | #191 dispersion sub-model |
| 3 | after #191 lands | `example7` rewrite |

Steps 2 and 3 need not block this PR. If we want #189 to merge on its current schedule,
the `cens()` support in `R/` is shippable on its own and `example7` can be removed from
this PR and rewritten against #191 in a follow-up.

Note that #191 has to be cut from the top of the stack rather than from `dev`: #172 and
#174 already modify `R/bayesnecformula.R`, `R/bnec.R`, `R/fit_bayesnec.R` and
`data-raw/sysdata.R`, which is precisely the set a dispersion sub-model touches — and
`R/sysdata.rda` is a binary, so a conflict there cannot be merge-resolved, only
regenerated.

Issue #177, opened 2026-08-01
nechormepwr fails to initialise on bounded (bernoulli) responses, including the hu block of hurdle_gamma


beckyfisher:
## Summary

On the `zi-beta-impl` branch, `nechormepwr` fails to initialise under
`family = "hurdle_gamma"`. `bnec()` reports that it could not find initial
values within the range of the response, falls back to Stan's default
initialisation, and then aborts with `Initialization failed`.

The failure is specific to the **combination**: the same data fit under
`hurdle_gamma` with a sibling hormesis model, and `nechormepwr` fits under
`Gamma` once the zeros are removed.

In practice this means a `model = "zero_bounded"` call under `hurdle_gamma`
silently averages over 9 equations rather than 11 — `nechormepwr01` is dropped
by design (not valid for the identity link), and `nechormepwr` drops out through
this failure. I hit it on all 24 tests of a real series before reducing it to
the reprex below.

## Reproducible example

```r
.libPaths(c("lib-bayesnec-zi", .libPaths()))
library(bayesnec)
set.seed(42)

## a hurdle response: mortality rises with concentration (response exactly 0),
## survivors' growth declines with it
n_per <- 60
conc  <- rep(c(0, 2.5, 5, 8, 12, 15, 25, 35, 50), each = n_per)
p_die <- 0.02 + 0.5 * (conc / 50)^2
mu    <- 16 * exp(-0.03 * conc)
alive <- rbinom(length(conc), 1, 1 - p_die)
resp  <- ifelse(alive == 1, rgamma(length(conc), shape = 25, rate = 25 / mu), 0)
dat   <- data.frame(conc = conc, resp = resp)
# n = 540, zeros = 69 (13%)

fit_one <- function(model, family, data) {
  form <- bnf(sprintf('resp ~ crf(conc, model = "%s")', model))
  try(bnec(form, data = data, family = family,
           iter = 2000, warmup = 1000, chains = 2, seed = 42,
           refresh = 0, open_progress = FALSE), silent = TRUE)
}

fit_one("nechormepwr", "hurdle_gamma", dat)                  # case 1: FAILS
fit_one("nechorme",    "hurdle_gamma", dat)                  # case 2: ok, 137 s
fit_one("nechormepwr", "Gamma",        dat[dat$resp > 0, ])  # case 3: ok, 107 s
```

## Observed

**Case 1 — `nechormepwr` / `hurdle_gamma`:**

```
[...]

PR #199, opened 2026-08-14
Allow a cens() aterm in bnec_hurdle() (#188)

### LINKED ISSUE #188: bnec_hurdle() rejects all aterms, so a censored response cannot be fitted as a hurdle

beckyfisher:
Closes #188.

## What changed

`bnec_hurdle()` accepts a `cens()` aterm on the response. Everything else on the
left of the formula keeps erroring, but now by name and with the reason.

- **`hurdle_response_var()` validates the response, not the whole left-hand
  side.** It previously required the entire LHS to deparse to a bare name, which
  is stricter than the constraint it protects: what the zero-as-death convention
  needs is that the *response variable* be a bare, untransformed name, and an
  aterm alongside it does not threaten that. A new `hurdle_lhs_parts()` splits
  the LHS into response and aterms; a transformed response is still rejected,
  and the message now names the response rather than the whole LHS.
- **`check_hurdle_aterms()`** accepts `cens()` and refuses the rest by name.
  `trials()` — neither component takes a trial count, and an aggregated count
  could not be split into survivors and deaths row by row. `weights()` — whether
  a weight applies to the growth component, the survival component or both is a
  modelling decision, and guessing it would silently change the model. Anything
  unrecognised gets a generic reason but is still named.
- **Routing.** The growth component gets the formula unchanged, aterms included:
  the censoring indicator is an ordinary data column, so the existing
  `data[y > 0, , drop = FALSE]` subset carries it along. The survival component
  drops it, because the alive/dead response is observed exactly. That was
  already what `swap_response()` did by rebuilding from `rhs()` alone; it is now
  deliberate, commented and tested rather than incidental.
- **`check_hurdle_cens()`** refuses any row that is both zero and censored. A
  structural zero and a censored observation are the two things a hurdle model
  exists to separate, and a row claiming to be both cannot be assigned to either
  block. The recycled-literal form `cens("left")` is caught by the same check.
- `?bnec_hurdle` gains a *Censoring, and which aterms are allowed* section,
  including the caveat that under a Gamma growth component the censoring bound
  cannot be `0` (`bnec()` already rejects that, correctly) — the bound has to be
  the resolution limit, so the encoding is "at most the smallest resolvable
  value" rather than "at most zero".

## How it was verified

**The issue's reprex fits.** It returns a `bayesnechurdlefit`.

**Simulated data where the answer is known** — 150 individuals, `nec3param`
growth with `top = 10`, `nec = 2`, `beta = log(0.6)`, mortality rising with
concentration, and survivors below a recording limit of 1.5 recoded as
left-censored at that limit:

```
survivors in data: 130   rows in growth fit: 130
alive coded in survival fit: 130
cens column present in growth data: TRUE
growth formula:   y | cens(cens) ~ top * exp(-exp(beta) * (x - nec) * step(x - nec))
survival formula: .alive ~ top * exp(-exp(beta) * (x - nec) * step(x - nec))
cens reached Stan (growth):   TRUE
cens reached Stan (survival): FALSE

growth parameters (truth: top 10, nec 2, beta log(0.6) = -0.51)
                 Estimate    Q2.5   Q97.5
top_Intercept      10.062   9.518  10.724
beta_Intercept     -0.516  -0.604  -0.417
nec_Intercept       1.985   1.787   2.199
```

This is the hazard the queue entry names: a left-censored survivor is an
observation of the growth component, not a structural zero. All 130 survivors
are in the growth fit and all 130 are coded alive in the survival fit, so no
censored survivor was routed to the hurdle block. The censoring declaration
reaches Stan for the growth block and not for the survival block.

**Tests.** `devtools::test()` on `test-bnec_hurdle.R`: 53 passing; on
`test-cens.R`, `test-bayesnechurdlefit-methods.R`, `test-bayesnecformula.R`,
`test-check_formula.R`, `test-make_brmsformula.R` together: 148 passing.
New tests cover the accepted case (`cens()` alone, with an interval upper bound,
and reading the response past an aterm), the rejected cases (`weights()`,
`trials()`, an unrecognised aterm, and a rejected aterm alongside an accepted
one), both directions of the zero-versus-censored check, that `swap_response()`
drops the aterm, and an end-to-end fit asserting the routing above.

## Observed while verifying, not fixed here

The reprex takes several minutes before it starts sampling, all of it in
`make_good_inits()`. Its growth data has three distinct response values across
30 rows, no candidate draw satisfies `check_init_predictions()`, and the search
runs its full `n_trials = 1e4` — about 5 minutes at ~30 ms a trial — before
falling back to Stan's random initialisation. Nothing to do with this change;
it is the same failure mode as #177, which is next in the queue, and worth
noting there that the cost is paid in full even when the search is hopeless.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -113,6 +113,32 @@
   nothing to return. Both methods now share one implementation, so the single-
   and multi-model paths cannot drift apart. See
   [#176](https://github.com/open-AIMS/bayesnec/issues/176).
+- `bnec_hurdle()` now accepts a `cens()` aterm on the response, which unblocks
+  the combination the censoring work was raised for: a growth endpoint that is
+  both zero-bounded with structural zeros (deaths) and left-censored at the
+  recording resolution (survivors measured below the limit). Only a two-part
+  model with a censored response component can tell those two zeros apart. The
+  guard was stricter than the constraint it protected — `hurdle_response_var()`
+  required the whole left-hand side to deparse to a bare name, when what the
+  zero-as-death convention needs is a bare *response*; an aterm alongside it
+  does not threaten that.
+- The declaration is routed rather than merely allowed. It is passed to the
+  growth component, where the censoring indicator travels as an ordinary data
+  column and is subset with everything else, and dropped from the survival
+  component, whose alive/dead response is observed exactly. That was already the
+  behaviour of `swap_response()`, incidentally; it is now deliberate and tested.
+- A row that is both zero and censored is refused. A structural zero and a
+  censored observation are the two things a hurdle model exists to separate, and
+  accepting a row claiming to be both would reproduce the confusion
+  `vignette("example6")` warns about. Under a Gamma growth component the
+  censoring bound cannot be `0` — `bnec()` already rejects that — so the
+  encoding is "at most the smallest resolvable value", which `?bnec_hurdle` now
+  says.
+- Other aterms keep erroring, but by name and with the reason. `trials()` has no
+  meaning for either component; `weights()` is refused because whether a weight
+  applies to the growth component, the survival component or both is a modelling
+  decision `bnec_hurdle()` will not make on the user's behalf. See
+  [#188](https://github.com/open-AIMS/bayesnec/issues/188).
 
 # bayesnec 2.1.3.6
 
--- a/R/bayesnechurdlefit-methods.R
+++ b/R/bayesnechurdlefit-methods.R
@@ -578,7 +578,9 @@ amend.bayesnechurdlefit <- function(object, drop, add, loo_controls,
 #' @param object An object of class \code{\link{bayesnechurdlefit}}.
 #' @param newdata Optional new data. Split into the two component datasets the
 #' same way \code{\link{bnec_hurdle}} does, so zeros continue to denote the
-#' hurdle rather than being modelled as small responses.
+#' hurdle rather than being modelled as small responses, and checked the same
+#' way: a row that is both zero and censored is refused here as it is at
+#' construction.
 #'
 #' @return An object of class \code{\link{bayesnechurdlefit}}.
 #'
@@ -597,6 +599,13 @@ update.bayesnechurdlefit <- function(object, newdata = NULL, ...) {
     stop("\"newdata\" must contain the response column \"", object$y_var,
          "\".", call. = FALSE)
   }
+  # The zero-vs-censored invariant belongs to the data, not to the fit, so it
+  # has to be re-checked against newdata. Without this a row that is both zero
+  # and censored is dropped from the growth refit and coded as a death in the
+  # survival refit, silently -- the confusion the construction-time check in
+  # bnec_hurdle() exists to prevent.
+  check_hurdle_cens(check_hurdle_aterms(object$formula), newdata, y,
+                    object$y_var)
   surv_data <- newdata
   surv_data[[".alive"]] <- as.integer(y > 0)
   out <- hurdle_rewrap(
--- a/R/bnec_hurdle.R
+++ b/R/bnec_hurdle.R
@@ -77,8 +77,12 @@
 #'
 #' \bold{Censoring, and which aterms are allowed}
 #'
-#' \code{cens()} is the one \pkg{brms} aterm accepted on the response, because
-#' it is the one a hurdle model needs. A growth endpoint can be both
+#' \code{cens()} is the one aterm accepted on the response. \code{\link{bnec}}
+#' itself carries three -- \code{trials()}, \code{weights()} and
+#' \code{cens()} -- and of those \code{cens()} is the only one whose meaning
+#' stays unambiguous once the response is split across two models. It is also
+#' the one with a use here that nothing else covers. A growth endpoint can be
+#' both
 #' zero-bounded with structural zeros and left-censored at the recording
 #' resolution, and only a two-part model with a censored response component can
 #' tell the two apart: a death is a structural zero belonging to the Bernoulli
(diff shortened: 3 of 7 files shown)


beckyfisher:
## Review follow-up (f8122bc)

Three low-severity findings from a code review, all peripheral to the core routing logic. The central mechanism was verified as correct: on the growth subset `make_standata()` receives `cens` with the expected censored/uncensored split, the `|`-bearing LHS still parses so `hurdle_xform_x()`/`hurdle_raw_data()` are unaffected, and the change is a strict relaxation of the old guard (for a formula with no `|`, `hurdle_lhs_parts()` returns the LHS unchanged, so `hurdle_response_var()` is byte-for-byte equivalent on every previously accepted formula).

**1. `update()` bypassed the zero-vs-censored invariant** — `R/bayesnechurdlefit-methods.R`

`update.bayesnechurdlefit()` re-split `newdata` with the same `y > 0` / `.alive` logic as `bnec_hurdle()` but only checked `is.null(y)`. A row that was both zero and censored was dropped from the growth refit and coded as a death in the survival refit, silently — exactly the confusion the construction-time check exists to prevent. The invariant is a property of the data, not of the fit, so `check_hurdle_cens()` now runs against `newdata` as well.

(Adjacent and left alone: `update()` also skips the `NA` / negative / all-zero response checks. That gap predates this PR.)

**2. Empty `cens()` raised an internal error** — `R/bnec_hurdle.R`

`cens()` with no arguments hit `subscript out of bounds` from inside a validation helper rather than a named diagnostic. Guarded, with a message pointing at the missing indicator column.

**3. `brms::cens()` was refused as an unrecognised aterm** — `R/bnec_hurdle.R`

`hurdle_lhs_parts()` named aterms with `deparse1(tt[[1]])`, giving `"brms::cens"`, so the namespace-qualified form was rejected with the misleading "no validated behaviour in a two-block fit" reason — even though `bnec()` accepts it (the rest of the package matches on `grepl("cens(", ..., fixed = TRUE)`). Any `::`/`:::` prefix is now stripped when deriving the name, so the qualified form of a *rejected* aterm also gets its proper reason.

### Noted, not changed

`cens(y2 = ub, cc)` — a positional argument after a named `y2` — makes the check inspect the wrong variable. `split_calls()` mis-parses the same form identically, so the fitted model is already wrong; resolving named arguments only in the check would make it disagree with the term `brms` is actually given, which is worse than agreeing and being wrong together. Left as-is with a comment, as a pre-existing `split_calls()` issue.

### Tests

`tests/testthat/test-bnec_hurdle.R` 56 pass / 0 fail with `NOT_CRAN=true` (end-to-end censored fit included), `test-bayesnechurdlefit-methods.R` 60 pass, and `test-cens.R` / `test-bayesnecformula.R` / `test-check_formula.R` clean. New cases cover the `update()` invariant, empty `cens()`, and the namespace-qualified form.

Issue #190, opened 2026-08-06
Re-run precompile.R before the next release: vignette output predates the current model code


beckyfisher:
Blocked by: #319, #340, #310

#319 stops `example8` rendering: the precompile errors at its first fit because
`crf()` resolves a variable model set in the global environment only. #340 means
a render cannot be attributed to the checkout it was run from, so a re-run
establishes nothing about the branch it was run on. #310 leaves the rendered
output different on every run, so what is committed is not yet a stable record.

## Summary

The precompiled vignettes carry output generated in January 2026. The current PR stack (#168 → #169 → #172 → #174 → #179 → #185) has since changed code that some of that output depends on, so parts of the published articles now describe behaviour the package no longer has. `precompile.R` should be re-run before the next CRAN release.

This is a release blocker in the weak sense: shipping is not *broken*, but it ships documentation that contradicts the code.

## What is specifically stale

**`example1`, the beta_binomial section.** `exp_1b` (`example1.Rmd.orig:161`) fits `beta_binomial` on the default `identity` link. #168 fixed the init failure for exactly that combination, so the fit shown was produced by the pre-fix initialisation path.

**`example2b`, the model-suitability prose and `models()` output.** The vignette prints `models()` at `.orig:46` and states that linear-hormesis models "are not suitable for modelling responses that are 0, 1-bounded ... however they are currently allowed". That is the subject of #170, and `check_models.R` and `R/sysdata.rda` both changed across the stack. The prose needs re-reading against the current exclusion rules, not just re-knitting.

**Figure vintage.** `example1`–`example4` figures were drawn under an older ggplot2; `example6` and the two `example2b` theoretical-curve figures are ggplot2 4.0.3. A full re-run brings them onto one version.

## What is *not* stale, as far as I can tell

`example3` (priors). The changes to `define_prior()` in the stack are purely additive — a new early-return branch for hurdle families and a new `define_hurdle_prior()` — so the non-hurdle defaults the vignette documents are unchanged.

## Cost, and why it has not been done

A full `precompile.R` run refits every model in every vignette. #185 deliberately avoided triggering it: that PR fixed the figure *device* (pdf → png in an html_vignette) and converted the existing figures rather than regenerating them, precisely so a presentation fix would not churn every reported number in the prose. The `.orig` files there are already correct, so a re-run needs no further preparation — it will produce PNGs natively.

[...]

PR #202, opened 2026-08-14
get_priors(): priors from a fit or from a formula and data (#141)

### LINKED ISSUE #141: Function get priors
Modify the get_priors function, or generate a new function to get priors from bayesnec that can be re-used as priors for bayesnec

beckyfisher:
Closes #141. Implements **D9**: both entry points, one function.

## What changed

**`get_priors()`**, a new exported generic with two entry points.

- **Given a fit** — a `bayesnecfit`, `bayesmanecfit` or `bayesnechurdlefit` —
  it returns the priors that fit actually used, user overrides included, so
  `bnec(formula, data = data, prior = get_priors(fit))` reproduces the model.
- **Given a formula and data** — as a `bayesnecformula`, a plain `formula` or a
  character string — it returns the priors `bayesnec` *would* generate, without
  fitting anything, so they can be inspected and edited before the first run.

A single model returns a `brmsprior`; a model set returns a named list of them,
named the way `bnec()` expects a named prior list. A `bayesnechurdlefit` returns
one per component, because the two are independent fits with independent model
sets.

The formula entry point runs the same sequence `bnec()` runs, through the same
functions — `model.frame()`, `retrieve_valid_family()`, `check_models()`,
`check_data()`, `define_prior()`. It picks the family from the data if none is
given and drops invalid models with the same message. A parallel implementation
would be the wrong answer the moment either changed; `check_data()` in
particular can move the data a prior is derived from (nudging zeros off a
boundary), and skipping it would give the wrong priors on exactly the datasets
where it matters.

## Two things had to be fixed for the round trip to work

The round trip is the whole point of the issue, and it did not work. Feeding
`pull_prior(fit)` back to `bnec()` fails, in two different ways:

1. **Extra rows.** A fit's `prior` slot holds more than was supplied: a
   duplicated `coef = "Intercept"` row per population-level prior and `brms`'s
   own `sigma` default. `make_inits()` matches prior names against the model's
   parameters, so the `sigma` row makes it error with *"user-specific parameter
   prior names (beta, bot, nec, top, sigma) do not match expectation (beta, bot,
   nec, top)"*. `get_priors()` returns the supplied rows only, which the
   `source` column identifies.

2. **Blank bounds.** `brms` records an absent bound as `""` in a fit's own
   `prior` slot, where `define_prior()` and `brms::prior()` use `NA`. All mean
   unbounded, but the bound-respecting redraw in `make_inits()` tested with
   `is.na()`, so `""` was read as a bound, coerced to `NA` by `as.numeric()`,
   and evaluated as `while (NA)` — *"missing value where TRUE/FALSE needed"*.
   Blank bounds are now normalised in `make_inits()` and `make_good_inits()`, so
   any prior in that form works, and `get_priors()` returns `NA` so its output
   compares equal to a generated prior.

`pull_prior()` is unchanged. It returns the whole `brmsprior` a fit carries and
is for looking at a fit; `?pull_prior` now says so and points at `get_priors()`.

## How it was verified

Fitting `nec4param` on `nec_data` with defaults, 2 chains:

```
entry points agree with no override:  TRUE
round trip class:                     bayesnecfit bnecfit
round trip priors identical:          TRUE
override recorded on the fit:         TRUE
override differs from generated:      TRUE
override round trips:                 TRUE
model set names:                      nec4param, ecx4param
model set fit class:                  bayesmanecfit bnecfit
model set round trips again:          TRUE
```

So: the two entry points agree when nothing is overridden; a fit's priors are
accepted by `bnec(prior = )` and come back identical from the refit; a user
override is recorded on the fit, makes the two entry points disagree — the case
D9 asks to be documented — and itself round trips; and a model set round trips
as a named list.

`tests/testthat/test-get_priors.R`, 40 assertions: the shape of the returned
object (no `sigma` row, no duplicated `Intercept` rows, `NA` rather than `""`
bounds, bounds that were set preserved); the named-list form for a model set;
the three ways of naming a formula agreeing; the family being chosen from the
data; the `bayesnechurdlefit` split; the errors for missing `data`, a wrong
class, and a model invalid for the family; `blank_bounds_to_na()` on blank,
present and already-`NA` bounds, and `make_inits()` succeeding on a prior in the
stored form; and the two slow tests above — the round trip, and the override
making the two entry points disagree.

Targeted suites all pass locally; the full suite runs on CI with this PR.

## What is returned, and what is left to `brms`

What comes back is the population-level priors on the parameters of the model's
own non-linear formula — the ones `bnec()` accepts through `prior`. Where a
route B `disp` term is in the formula (`disp("power")` and the other named
variance functions), its parameters `c0`, `c1`, … belong to that formula and
come back with the curve parameters, from both entry points. They have to:
`bnec()` takes a supplied prior whole, so a set handed back without them leaves
`brms` to put a flat prior on each, which is a different model and the state
`define_disp_prior()` exists to prevent. Verified — a `nec4param + disp("power")`
fit round trips to the identical six-row set.

What is left to `brms` and its defaults is the family's own dispersion parameter
(`sigma`, `shape`, `phi`) where dispersion is constant, and the linear predictor
of a `disp(~x)` sub-model. Neither can be supplied through `bnec(prior = )`
today, since `make_inits()` rejects any prior naming a parameter outside the
model's own set, so neither is returned. `?get_priors` says so. Lifting that
restriction is #207.

## Review follow-ups

- The dispersion paragraph above replaces an earlier note that claimed no
  dispersion priors were returned at all, which was wrong for route B. Two tests
  were added with it: the formula entry point returning `c0`/`c1` for route B and
  nothing extra for route A, and `usable_prior()` keeping the disp rows while
  dropping the vectorized duplicates and the `brms` `sigma` default. Both assert
  on the stored prior form, so no disp model is compiled for them.
- #207 opened: `bnec(prior = )` cannot accept a prior on the dispersion
  parameter, and does not notice when a supplied prior omits parameters the
  formula introduces.
- The Priors vignette (`example3`) should gain a `get_priors()` section — it
  still teaches fit-first, `pull_prior()`, hand-write. Not done here, since the
  protocol forbids touching vignettes or running `precompile.R`; noted on #190.


beckyfisher:
## Review

The implementation does what #141 asked and the design decision I'd have worried about most was avoided: `get_priors.formula()` runs the same sequence `bnec()` runs (`model.frame()` → `retrieve_valid_family()` → `check_models()` → `check_data()` → `define_prior()`) rather than a parallel implementation. I checked the behaviours that could have drifted from `bnec()` and none have — group model names expand identically (`crf(x, "ecx")` → all 12 equations), the hurdle path returns both blocks with their `hu*` names, and an unsupported family/link raises the same error. The two blocking bugs it fixes (the extra `sigma` row, the `""` bounds) are real and correctly diagnosed.

Three things came out of the review. Two are fixed on the branch; the third is now #207.

### 1. The dispersion note was wrong for `disp("power")`

The Note said only the curve parameters are returned and "a prior on the dispersion parameter is left to `brms` and its default". That is not what happens where a route B `disp` term is in the formula. `define_prior()` takes `disp_spec` and appends `define_disp_prior()`, and `get_priors.formula()` passes it:

```
=== disp("power") ===
 normal(0, 5)         b   beta
 normal(0.93, 0.55)   b   top
 normal(0.38, 0.55)   b   bot
 gamma(5, 2.28)       b   nec
 normal(-1.515, 2)    b   c0     <- dispersion sub-model
 normal(0, 2)         b   c1     <- dispersion sub-model
```

The fit entry point returns them too — fitting `nec4param + disp("power")` and calling `get_priors(fit)` gives the same six rows. They survive `usable_prior()` because `brms` stores them as `class = "b"`, `coef = ""`, `source = "user"` (it matches them by `nlpar`; `dpar` stays blank even in the fit's own slot).

This is the correct behaviour, not a bug — `bnec()` takes a supplied prior whole, so a set returned *without* those rows would leave `brms` to put a flat prior on each, which `define_disp_prior()` exists to prevent. The round trip holds: `bnec(y ~ crf(x, "nec4param") + disp("power"), prior = get_priors(...))` refits and returns the identical six-row set. Only the documentation was wrong.

What *is* left to `brms` is the family's own dispersion parameter under constant dispersion, and the linear predictor of a `disp(~x)` sub-model. The docs, the `usable_prior()` note and the NEWS entry now say that instead, and there are two new tests: the formula entry point returning `c0`/`c1` for route B and nothing extra for route A, and `usable_prior()` keeping the disp rows while dropping the vectorized duplicates and the `brms` `sigma` default. Both are assertions on the stored prior form rather than on a fit, so no disp model has to be compiled and sampled for them.

### 2. `bnec(prior = )` cannot take a dispersion prior at all — #207

Confirmed: `make_inits()` tests `identical(sort(par_names), sort(fct_args))`, so a `class = "sigma"` row errors with "user-specific parameter prior names (top, beta, bot, nec, sigma) do not match expectation". The fix looks small — filter to the model's own `class == "b"` rows before the name check, the way `add_brm_defaults()` already does for `disp_pars()`, and let Stan random-init the dispersion parameter, which is what happens today anyway.

While checking that, a second defect turned up: a prior set that omits the `c0`/`c1` rows is accepted silently by a formula that *does* carry `disp("power")`, and the fit proceeds on `(flat)` priors for both. `validate_priors()` takes any `brmsprior` wholesale and never checks it covers the model. Editing a returned prior set and handing it back is exactly the workflow this PR invites, so it becomes easier to hit. Both are written up in #207.

### 3. The Priors vignette

`example3` still teaches fit-first → `pull_prior()` → hand-write a complete set with `prior_string()`, and states that a user supplying priors "requires them to specify priors for **all** parameters" — which is the pain `get_priors(formula, data)` removes. It should gain a `get_priors()` section.

Not done here: `notes/implementation/00_protocol.md` forbids touching vignettes or running `precompile.R`, and `.Rmd.orig`/`.Rmd` have to move together. Noted on #190 instead, whose body currently says `example3` is not stale — true of the prior *defaults* it documents, no longer true of the workflow.

### Minor

- Base is `issue-133-failed-models`, so this stacks on #201 — merge in order.
- `if (Sys.getenv("NOT_CRAN") == "") skip_on_cran()` was equivalent to a bare `skip_on_cran()`; simplified to match the rest of the suite.

Verified with R 4.6.1, brms 2.23.0. The non-fitting tests in `test-get_priors.R` pass; the two `skip_on_cran()` round-trip tests are left to CI.

Issue #183, opened 2026-08-05
Support the zero_inflated_beta family for proportional responses containing exact zeros


beckyfisher:
## Summary

Extends the two-block machinery to proportional responses on (0, 1) that also
contain exact zeros — the algal growth-rate case, where growth is expressed
against a ceiling and some replicates fail outright.

`zero_inflated_beta` is structurally the same model as `hurdle_gamma`.
Zero-inflation differs from a hurdle only when the base distribution can itself
emit zeros, and neither the Gamma nor the Beta can — `brms`'s own generated Stan
for `zero_inflated_beta` is the hurdle form, with no `log_sum_exp`. So both
routes carry over unchanged:

```r
# joint
bnec(y ~ crf(log_x, "nec3param"), data = dat, family = "zero_inflated_beta")

# factorised
bnec_hurdle(y ~ crf(log_x, "all"), data = dat)   # picks Beta automatically
```

## What needs handling

The main mechanical difference is that `brms` names the second block `zi` rather
than `hu`, giving `zitop`, `zinec` and so on. That name should be carried through
from a lookup rather than hardcoded.

Three places the two families genuinely differ:

- **`check_models()`** — both blocks are 0, 1 bounded here, so `nechormepwr01`
  should be **kept**; it is the equation designed for that range.
  `hurdle_gamma` drops it via its zero-bounded mu block.
- **`check_data()`** — zeros are preserved as the signal, but **ones still need
  nudging**, since the Beta's support is open at 1. Easy to miss, and it would
  fail at sampling time rather than at validation.
- **`bnec_hurdle()`** — the survivor-component family must be chosen from the
  **non-zero subset**, so proportional data gets a Beta rather than a Gamma.
  Choosing from the whole response would read the zeros as part of a continuous
  distribution rather than as the hurdle they are.

## Constraint worth documenting

Where a proportional response was obtained by dividing through by a maximum,
that divisor must be a constant fixed in advance — a design ceiling or a
historical value — not one computed from the dataset being analysed. Dividing by
an estimated quantity induces correlation across all observations, biases
effective doses upward and understates their uncertainty. See #173, which covers
that criticism in full.

## Status

Implemented in #174, stacked on #172. Filed retrospectively so the work is
findable from the issue list.

Follow-ups deliberately left out of that PR:

- `hurdle_lognormal`, which needs its own handling because lognormal's `mu` is
  on the log scale despite `brms` reporting an identity link.
[...]

Issue #148, opened 2023-08-24
Add more diagnostic methods and tests for bayesnecfit objects


beckyfisher:
At the moment most diagnostic tests only check for model convergence. Give the sensitivity of NSEC on estimates of variance, diagnostic tests to ensure that the model simulated variance is representative of the input data would provide surety that the estimates are valid.

beckyfisher:
IN the first instance we could implement a pp_check method for the bayesnecfit class.
This include some examples in the single model usage vignette and/or it may be worth building a dedicated model checking vignette.

beckyfisher:
## Scoping this out

The issue as written asks for "diagnostic tests to ensure that the model simulated variance is representative of the input data". That is the right target, but it is not yet a specification, so this comment tries to turn it into one: what exists, where the actual hole is, what would fill it, and what has to be decided before any of it can be written.

Two things are settled already and are assumed here rather than re-argued. First, the split recorded in D6: sensitivity to the *estimator's* own choices (`sig_val`, the reference definition, `resolution`) belongs to `toxval`, because it applies equally to `nsec.drc` and `nsec.brmsfit`; sensitivity to what the *fit* supplies belongs here. Nothing below re-opens that. Second, `pp_check` was already proposed in the comment above, and it is part of the answer — but only part, and not the interesting part.

---

### 1. What a `bayesnec` fit currently offers as a diagnostic

| | what it checks |
|---|---|
| `check_chains()` | mixing, autocorrelation |
| `rhat()` | convergence |
| `check_priors()` | prior vs posterior overlap |
| `loo` / `waic` in `mod_stats` | relative predictive fit between candidate models |
| `dispersion()` | observed vs simulated Pearson SS — **Poisson and Binomial only** |
| `plot()` / `autoplot()` | the curve, by eye |

Everything in that list except `dispersion()` is about the sampler or about *relative* model ranking. Nothing asks whether the fitted model's implied variability matches the data's — which is what the issue is about, and the one assumption that `nsec` rests on most heavily.

### 2. Why this lands on NSEC specifically

`nsec` sets its reference at `R/nsec.R:155`:

```r
reference <- quantile(p_samples[, 1], sig_val)
```

[...]

PR #204, opened 2026-08-14
Zero-inflated Poisson and negative binomial families (#104)

### LINKED ISSUE #104: add zero-inflated Poisson/negative binomial distributions
I'm interested in using zero-inflated distributions for count data. These are available in brms. I'd be happy to contribute code but I could use some pointers to get started.

beckyfisher:
Closes #104. Zero-inflated counts take the joint single-block path, `bnec_hurdle()`
refuses them, and the reasoning is in the documentation.

## What changed

`"zero_inflated_poisson"` and `"zero_inflated_negbinomial"` are now accepted
families. They take the **ordinary single-block path**: the
concentration-response curve is fitted to `mu`, and `brms` estimates a single
constant `zi` alongside it. No second bayesnec equation, no `model_survival`, no
`zi` curve.

- `mod_fams` gains the two tags. They are deliberately **absent** from
  `hurdle_fams`, and that single omission is the whole design decision:
  `is_hurdle_family()` stays `FALSE`, so `define_hurdle_prior()`,
  `make_good_hurdle_inits()`, `add_hu_block()`, `check_model_survival()` and the
  hurdle branch of `expand_nec()` are all bypassed.
- `check_models()` treats them as count families: the mu block is a count mean,
  so the same four models go (`neclin`, `neclinhorme`, `ecxlin`,
  `nechormepwr01`).
- `define_prior()` maps the tag onto its base count family rather than
  duplicating eight table entries — the mu block *is* a poisson or negbinomial
  mean.
- **`bnec_hurdle()` refuses them** as `family_growth`, with the reason and a
  pointer to `bnec(family = )` for the mixture. It also now refuses a two-block
  `family_growth`, which was previously accepted and would have nested one
  two-part model inside another.
- **`disp()` is refused**, before the generic `has_disp_par()` check, whose
  message ("no free dispersion parameter") would be wrong for
  `zero_inflated_negbinomial` — it does have a `shape`. The real reason is
  different: that parameter describes the count component while the response is
  the mixture.
- `?bnec` gains a *Zero-inflated counts are not two-block families* section
  carrying the argument below.

## Why `zi` is held constant

The temptation is to reuse the `zero_inflated_beta` path, since the name
matches. It is not the same model.

Zero-inflation differs from a hurdle **only when the base distribution can
itself produce a zero**. Neither Gamma nor Beta can, so for those the two
coincide: every zero must have come from the inflation component, the likelihood
separates exactly into a Bernoulli term over all observations and a
positive-response term over the rest, and the Stan density `brms` generates for
`zero_inflated_beta` is the hurdle form with no mixture at zero. Two blocks is
then an accurate description, and #183 was right to treat it that way.

Poisson and negative binomial **can** produce a zero, so the equivalence fails.
An observed zero is evidence about both components at once, the likelihood
carries a `log_sum_exp` over them and does not factorise. That rules out
`bnec_hurdle()`, which is exactly that factorisation performed as two separate
fits, and it refuses these families accordingly.

It does **not**, on its own, rule out a joint fit carrying a curve on `zi`.
`brms` treats `zi` as a distributional parameter like any other and will model
it on a predictor; the `log_sum_exp` is internal to the density. `zi` is held
constant here for two reasons that outlast the factorisation argument:

1. **Identifiability degrades where it matters.** `zi` and `mu` trade off — a
   low count rate is either a small `mu` or a large `zi` — and the data separate
   them only through the shape of the positive counts. In a concentration-response
   design `mu` runs from large at the control to near zero at the top
   concentration, so a `zi` curve would be well informed at the low
   concentrations and almost unidentified at the high ones that set the
   threshold. A hurdle has no such problem: every zero belongs to the second
   block by construction.
2. **`zi` is a latent class, not an observable.** Nothing in the data records
   which zeros are structural, so a threshold estimated from a `zi` curve would
   describe something that was never measured. In the survival case the deaths
   were observed, and that is what makes the second block interpretable.

Which gives the rule for choosing between them: **if you can tell which zeros
are structural, you have a hurdle, not zero-inflation.**

## How it was verified

Simulated zero-inflated counts: `nec3param` with `top = 40`, `nec = 2`,
`beta = log(0.5)`, `zi = 0.25` applied independently of dose, 150 observations,
36 zeros.

```
zero_inflated_poisson: class bayesnecfit bnecfit in 69s
  family in fit: zero_inflated_poisson | link: identity
  brms formula:  y ~ top * exp(-exp(beta) * (x - nec) * step(x - nec))
  fixef rows:    top_Intercept, beta_Intercept, nec_Intercept
  no bayesnec zi curve: TRUE
  zi estimate (truth 0.25): 0.243 [0.178, 0.314]
  nec (truth 2): 2.196
  ecx10 ok: TRUE | predict rows: 150 | summary ok: TRUE

zero_inflated_negbinomial: class bayesnecfit bnecfit in 76s
  ... same shape ...
  zi estimate (truth 0.25): 0.244 [0.183, 0.314]
  nec (truth 2): 2.191
```

The curve carries only `top`, `beta` and `nec`; `zi` is a single estimated
constant and recovers its true value; `nec` recovers close to truth. `ecx()`,
`nec()`, `predict()` and `summary()` all work.

`tests/testthat/test-zero-inflated-counts.R`: the families validate with an
identity link on `mu`; `is_hurdle_family()` is `FALSE` for them and `TRUE` for
`zero_inflated_beta` and `hurdle_gamma`; `check_models()` gives the same set as
plain `poisson`; the priors are identical to the base family's; `model_survival`
is refused; `bnec_hurdle()` refuses both, with the message saying *why* and
pointing somewhere useful, and still accepts a plain count family; `disp()` is
refused with the mixture reason and not the generic one; and the end-to-end fit
asserting the curve is on `mu` alone and `zi` is present as a constant. The
`disp()` refusal was also checked end to end through `bnec()`.

## Known limitations, recorded rather than fixed

- **A hurdle on counts is not available** — #209. Its positive part is
  zero-truncated, and bayesnec has no zero-truncated count family.
  `bnec_hurdle()` accepts a count `family_growth` but fits it *untruncated* to
  the non-zero subset, estimating `mu / (1 - exp(-mu))` rather than `mu` —
  negligible for large `mu`, material as `mu` falls towards zero. `?bnec_hurdle`
  now documents this, and the refusal message no longer recommends that route.
- **The prior tables degrade with many zeros** — #210. `top` and `bot` are built
  from quantiles of the raw response, so once a quarter of it is zero the `bot`
  prior collapses to a mean near zero, and past three quarters the `top` prior
  becomes `gamma(2, Inf)`. Pre-existing for any zero-heavy `poisson` or `Gamma`
  response, but these families make it the normal case. The misleading comment
  at the call site is corrected here.
- **Predictions are on the mixture scale.** `posterior_epred()` returns
  `mu * (1 - zi)`, so `predict()`, `fitted()`, `autoplot()` and `pred_vals` sit a
  factor `1 - zi` below the `top` and `bot` that `summary()` reports. `ecx()` and
  the threshold are unaffected — a constant factor cancels from both. Documented
  under `?bnec`.
- **Not selected automatically.** A count response is still read as `poisson`;
  the zero-inflated forms have to be asked for. Choosing between a count family
  and its zero-inflated form from the data alone is a modelling decision.
- **`dispersion()` does not apply.** It covers `poisson` and `binomial` only, so
  a zero-inflated fit reports `NA` for it, as `Gamma` and `negbinomial` already
  do. For `zero_inflated_poisson` a dedicated over-dispersion diagnostic would
  have to account for the mixture, which is #148 territory.
- **`R/sysdata.rda` is regenerated** from `data-raw/sysdata.R`, so the diff
  includes a binary file. Verified object by object against the parent commit:
  only `mod_fams` changed.


beckyfisher:
## Review

The implementation is clean and the tests are thorough. Two substantive points
below — one about the justification rather than the code, one about a prior that
degrades in exactly the regime these families are for — plus some smaller
things. Verified along the way:

- **`R/sysdata.rda` is honest.** Loaded the new blob and the parent commit's
  side by side and compared every object: only `mod_fams` differs. `hurdle_fams`
  still holds `hurdle_gamma` and `zero_inflated_beta` alone, which is the whole
  design decision.
- **No regressions.** `test-zero-inflated-counts.R` passes (41 assertions
  off-CRAN, the end-to-end fit skipped), as do `test-bnec_hurdle.R`,
  `test-check_models.R`, `test-define_prior.R`, `test-disp_model.R`,
  `test-hurdle_family.R` and `test-dispersion.R`.
- **The `define_prior()` prefix strip is correctly ordered** — it sits after the
  `logit`/`log` branch that rewrites `fam_tag` to `"gaussian"`, so a
  non-identity link still gets the gaussian priors rather than falling through
  to poisson.
- **The D4 premise checks out.** `brms:::posterior_epred_zero_inflated_poisson`
  and its negbinomial twin are both `mu * (1 - zi)` — a genuine mixture, not the
  hurdle form. The contrast drawn with `zero_inflated_beta` is accurate.

---

### 1. The argument for D4 proves the wrong thing

`?bnec`, the `bnec_hurdle()` refusal and the `hurdle_fams` comment all rest on
"the likelihood carries a `log_sum_exp` and does not factorise". That correctly
rules out `bnec_hurdle()`, which is a factorised **two-fit** procedure and
genuinely cannot represent this model.

It does not rule out the **two-block joint fit**, which is what
`bnec(family = "hurdle_gamma")` does and which never relied on factorisation.
brms lists `zi` as a predictable distributional parameter:

```
zero_inflated_poisson     | dpars: mu, zi
zero_inflated_negbinomial | dpars: mu, shape, zi
```

so `zi ~ <bayesnec equation>` fitted jointly with `mu ~ <bayesnec equation>` is
a well-defined model; the `log_sum_exp` is internal to the density and is
indifferent to a covariate driving `zi`. As written, `?bnec` says such a fit
"would give a different model from the one asked for" — it would give a
different model that is also valid, and which bayesnec is choosing not to offer.

The decision to keep `zi` constant is still the right one, but for different
reasons that are worth writing down instead:

1. **Identifiability degrades exactly where it matters.** `zi` and `mu` trade
   off — a low count rate is either small `mu` or large `zi` — and the data
   separate them only through the shape of the positive counts. In a
   concentration-response design `mu` runs from large at the control to near
   zero at the top dose, so a `zi` curve would be well informed at low
   concentration and almost unidentified at high concentration. A hurdle has no
   such problem: every zero is structural, so the second block is informed at
   every dose.
2. **`zi` is a latent class, not an observable.** Nothing in a zero-inflated
   dataset says which zeros are structural, so a threshold on the `zi` curve
   would be a threshold on something never measured. In the survival case the
   deaths were observed, which is what makes that second block interpretable.

Which gives the decision rule the docs should be offering: **if you can tell
which zeros are structural, you have a hurdle, not zero-inflation.**

### 2. The `bot` prior collapses in the regime these families exist for — #210

`R/define_prior.R:40` maps the tag onto the base count family, and the new
comment says taking quantiles over the whole response "places `top` a little
low" and that the priors are "weakly informative either way". That is wrong on
both the parameter and the magnitude. Reproduced:

```r
set.seed(1)
x  <- as.numeric(rep(1:10, each = 15))
mu <- 40 * exp(-0.35 * pmax(x - 2, 0)) + 5          # true bot = 5
y  <- as.numeric(rpois(length(x), mu) * rbinom(length(x), 1, 0.7))   # zi = 0.30
define_prior("nec4param", validate_family("zero_inflated_poisson"), x, y)
#>  gamma(2, 0.0833)   top
#>  gamma(2, 66.67)    bot     <- prior mean 0.03, true bot = 5
```

Once `zi >= 0.25` — the regime you reach for these families in —
`quantile(response, 0.25)` is exactly `0` and the `bot` rate is set entirely by
the `min(positive)/100` fudge term. Under `prior_type = "regularizing"` the
denominator uses `quantile(response, 0)`, which is `0` for *any* zero-inflated
response, so `bot` is pinned near 0.1 unconditionally. Past 75% zeros the `top`
prior becomes the literal string `gamma(2, Inf)`.

This is not new — a plain `poisson` response with many zeros has always hit it —
but #104 makes it the normal case rather than an accident. Filed as #210; the
misleading comment is corrected on this branch.

### 3. `bnec_hurdle()`'s refusal points at a mis-specified fit

The message says: *"leave `family_growth` unset — bnec_hurdle will choose a
poisson or negbinomial growth component from the non-zero counts, which is the
hurdle you want."* Three problems:

- The positive part of a hurdle Poisson is **zero-truncated**. `bnec_hurdle()`
  fits an untruncated `Poisson(mu)` to `data[y > 0, ]`, which estimates
  `mu / (1 - exp(-mu))`. Negligible for large `mu`, material as `mu` falls
  towards zero — the high-concentration end that determines the NEC and ECx.
  For `hurdle_gamma` the analogous concern does not arise, which is why the
  existing design is right there and not here.
- `set_distribution()` never returns `negbinomial`, so "poisson or negbinomial"
  overstates what the auto path can pick.
- It returns `"Gamma"`, not `"poisson"`, whenever the counts are stored as
  `numeric` rather than `integer` (`R/set_distribution.R:29-36`), so the
  promised count hurdle can silently become a Gamma fit.

The family that would actually serve this case is `hurdle_poisson` /
`hurdle_negbinomial` — filed as #209. Message reworded on this branch to stop
recommending the untruncated route.

### 4. Predictions are not on the `mu` scale, and `?bnec` doesn't say so

`posterior_epred()` returns `mu * (1 - zi)`, so `fit$pred_vals`, `predict()`,
`fitted()` and `autoplot()` describe the mixture mean while `top` and `bot` in
`summary()` are on the `mu` scale. With this PR's own verification example
(`top ~ 40`, `zi ~ 0.24`) the plotted plateau sits at ~30 against a reported
`top` of ~40. `ecx()` is unaffected — both `"relative"` and `"absolute"` reduce
to ratios in which the constant factor cancels — and because `zi` is constant
the combined curve leaves its plateau at the same `nec`, so the threshold is
fine too. But the gap between the plot and the summary needs a sentence. The
hurdle path carries exactly this caveat at `R/expand_classes.R:79-83`. Added on
this branch.

### 5. `man/bnec_hurdle.Rd` not regenerated

`@param family_growth` still documents only the Gamma/Beta auto-selection and
says nothing about the two new refusals. `family_growth = "hurdle_gamma"` was
previously accepted and is now an error, with no record of that anywhere in the
help. Fixed on this branch.

### Minor

`test-zero-inflated-counts.R:41` draws `rpois(80, 12)` without a `set.seed()`,
unlike the end-to-end test below it. Not a flake risk — `zi_pr` and `base_pr`
are computed from the same `y`, so they are equal by construction — but it is
the repo convention. Seeded on this branch.

beckyfisher:
## Changes made in response to the review

Pushed as b74f681. Base also retargeted from `issue-177-nechormepwr-bounded` to
`dev`, since #203 merged that branch this morning — this PR is now exactly one
commit ahead of `dev`. The description above has been updated to match.

**1. The D4 rationale, rewritten** (`?bnec`, `?bnec_hurdle`, `NEWS.md`,
`data-raw/sysdata.R`). The factorisation argument is kept where it applies — it
does rule out `bnec_hurdle()` — but no longer does duty for the joint case,
which it never settled. `zi` is still held constant; the stated reasons are now
identifiability (`zi` and `mu` are weakly separated exactly where `mu` is small,
which is the end of the range that sets the threshold) and interpretation (`zi`
is a latent class, not something the experiment observed). `?bnec` now ends on
the rule that follows: if you can tell which zeros are structural, you have a
hurdle, not zero-inflation.

**2. `bnec_hurdle()` no longer recommends a mis-specified fit** (#209). The
refusal message pointed at "leave `family_growth` unset ... which is the hurdle
you want". It isn't: the positive part of a count hurdle is zero-truncated, and
`bnec_hurdle()` fits an untruncated poisson to the non-zero subset. The message
now says the count hurdle is unavailable and why, rather than sending users to
the wrong model, and `@param family_growth` documents both the new refusals and
the truncation caveat. Filed #209 to add `hurdle_poisson` / `hurdle_negbinomial`
as two-block families — that is the family this case actually wants, and most of
the machinery is already in place (`check_models.R:124` has anticipated a
poisson or negbinomial mu block all along).

**3. The prediction scale is documented** (`?bnec`, `NEWS.md`).
`posterior_epred()` returns `mu * (1 - zi)`, so the plotted curve sits a factor
`1 - zi` below the reported `top`. Noted along with why `ecx()` and the
threshold are unaffected.

**4. The `define_prior()` comment is corrected** (#210). It claimed the
whole-response quantiles put `top` "a little low" and were weakly informative
either way. It is `bot` that collapses, to a prior mean near zero, once a
quarter of the response is zero. The comment now states what actually happens
and why conditioning on the non-zeros — the fix `define_hurdle_prior()` uses —
is not transferable here. The guard itself is #210; left out of this PR because
it touches the shared poisson/Gamma prior path and so reaches beyond #104.

**5. `man/bnec_hurdle.Rd` regenerated**, and `set.seed()` added to the prior
test.

Test file now at 42 assertions off-CRAN, all passing, along with
`test-bnec_hurdle.R`, `test-check_models.R`, `test-define_prior.R`,
`test-disp_model.R` and `test-hurdle_family.R`. No behavioural change in this
commit beyond the reworded error message, which its test now pins.

PR #205, opened 2026-08-14
Stop caching the posterior prediction matrix on fitted objects (#180)

### LINKED ISSUE #180: bnec() caches a posterior prediction matrix that is ~25x the size of the fit itself
## Summary

The dominant memory cost of a `bayesnecfit` is not the fitted model but the
cached posterior prediction matrix `pred_vals$posterior`, built by
`expand_nec()`. It scales as `n_draws x resolution` and is roughly **25 times
larger than the `brmsfit` it accompanies**.

## Measured

Single `nec3param` fit, Gamma, 140 observations, `chains = 4, iter = 2000,
warmup = 1000` (4000 retained draws):

```
resolution = 1000                    resolution = 200
  whole bayesnecfit    31.8 MB         whole bayesnecfit     7.4 MB
    $fit (brmsfit)      1.2 MB           $fit (brmsfit)      1.2 MB
      @sim (draws)      1.0 MB             @sim (draws)      1.0 MB
      $criteria (loo)   0.0 MB             $criteria (loo)   0.0 MB
    $pred_vals$posterior              $pred_vals$posterior
                       30.5 MB                              6.1 MB
                  [4000 x 1000]                        [4000 x 200]
    $ne_posterior       0.0 MB           $ne_posterior       0.0 MB
    $predicted_y        0.0 MB           $predicted_y        0.0 MB
```

96% of the object is the prediction matrix, and it is exactly
`n_draws * resolution * 8` bytes.

## Why it matters

The cost multiplies with the model set and with `iter`. At the package
defaults (`iter = 1e4`, so 8000 retained draws, `resolution = 1000`) a single
model stores ~64 MB, so:

* `model = "all"` (23 models) retains roughly **1.5 GB** in one `bnec()` call;
* a `bayesmanecfit` additionally stores `w_pred_vals$posterior`, the weighted
  matrix, on top of the per-model ones;
* `bnec_hurdle()` doubles it again, since it holds two component fits.

This makes multi-fit workflows expensive in a way that is not visible from the
model being fitted, and it is why `resolution` — documented as controlling the
precision of `ecx`/`nsec` — is in practice also the main memory dial.

## Reproducer

```r
library(bayesnec)
set.seed(11)
nec3 <- function(x, top, beta, nec) top * exp(-exp(beta) * (x - nec) * (x > nec))
conc <- rep(c(0, 0.5, 1, 2, 3, 4, 5), each = 20)
mu <- nec3(conc, 25, log(0.55), 1)
d <- data.frame(x = conc, y = rgamma(length(conc), 12, 12 / mu))

f <- bnec(y ~ crf(x, model = "nec3param"), data = d, family = "Gamma",
          chains = 4, iter = 2000, warmup = 1000, seed = 11, refresh = 0)

format(object.size(f), units = "MB")                     # ~31.8 Mb
format(object.size(f$fit), units = "MB")                 # ~1.2 Mb
format(object.size(f$pred_vals$posterior), units = "MB") # ~30.5 Mb
dim(f$pred_vals$posterior)                               # 4000 x 1000
```

## Possible directions

1. Recompute predictions on demand rather than caching them. They are a
   deterministic function of the fit and the grid, and most methods
   (`ecx.bayesnechurdlefit`, the hurdle plot methods) already recompute at the
   resolution the caller asks for rather than reading the cache.
2. Store at a lower default resolution than is used for reporting, given that
   the cached matrix mainly feeds the NSEC for `ecx`-type models, and that NSEC
   is located by interpolation whose precision saturates near 100 (see #166 and
   the note in #39).
3. Failing either, document that `resolution` governs memory as well as
   precision, and that it is the dial to reach for on large model sets.

## Where this came from

Rendering `vignette("example6")` peaked at 13.4 GB RSS: 48 hurdle model fits
across four contaminants, held one contaminant at a time. That is transient
high-water rather than retention, and it is a separate matter from this issue.

Two things I want to explicitly rule out, having tested for both, so nobody
chases them:

* **No forked-worker leak.** Six sequential `bnec()` fits at `cores = 4`, and
  four sequential six-model `bnec_hurdle()` fits, both leave exactly one child
  process between fits and none at exit.
* **No unbounded accumulation.** Across four sequential six-model hurdle fits
  the post-`gc()` floor was 1.53, 2.12, 2.93, 2.67 GB — it plateaus rather than
  growing without limit, consistent with R holding a high-water mark.

beckyfisher:
Closes #180, #211, #213, #214. Implements **D2**: drop the cache, compute on demand.

## What changed

A fitted object no longer stores any of the `n_draws x resolution` posterior
matrices it used to keep. The small summaries the plot methods use are unchanged.

**`pred_vals$posterior` on every `bayesnecfit`** (#180). The one reader was the
model-averaging path in `expand_manec()`. Nothing else touched it: `predict()`
did not, and the plot methods use `pred_vals$data`.

**`w_pred_vals$posterior` on every `bayesmanecfit`** (#213). Once the per-model
matrices were gone this single weighted matrix was what remained — 77% of a
five-model set. Nothing in the package read it either. `w_pred_vals$data`, which
`plot()` and `autoplot()` do read, is unchanged.

**`hurdle$mu_pred` and `hurdle$hu_pred` on every two-block fit** (#214). Written
on every hurdle fit and never read anywhere — not by the plot methods, not by
`ecx()` / `nsec()`, not by the `bayesnechurdlefit` methods, not by the tests, and
not documented as slots. They are no longer computed at all, which removes two
`posterior_epred()` calls per two-block fit and four per `bnec_hurdle()` pair.

### How the model-averaged draws are built now

`expand_manec()` needs each model's posterior on the grid to assemble the
weighted draws. It computes each one and **immediately thins it** to the
`round(sample_size * wi)` rows the weighting keeps, one model at a time, in
`w_grid_pred_calc()`. Accumulating them into a list first would hold every
model's full matrix simultaneously — the same peak #180 is about, merely not
retained afterwards. The model weights do not depend on the posteriors, so
nothing forces them to be built together. Thinning goes through the existing
`w_pred_list_calc()`, so this and `posterior_epred.bayesmanecfit()` cannot
sample differently. The weighted draws are unchanged.

### One prediction grid (#211)

`bnec_newdata()`, `expand_nec()` and `posterior_on_grid()` each built the same
grid from their own copy of the same code. The copies had already diverged:
`bnec_newdata()` tested `any(is.na(x_range))` while the other two tested
`any(is.na(x_range[1]))`, so a partially specified range was silently ignored by
one and turned into `seq(NA, NA)` by the others — and even that differed by
which end was missing. There is now a single internal `prediction_grid()` that
all three delegate to, and `check_args_newdata()` rejects a partially specified
`x_range` with a clear error rather than guessing.

## What it cost

**It is a net saving in computation, not a trade.** `expand_nec()` used to
compute the matrix for every model whether anything needed it or not. Now:

| | before | after |
|---|---|---|
| single threshold model | 1 `posterior_epred` | **0** |
| single smooth model | 1 | 1 |
| single two-block fit | 3 | **0–1** |
| model set, per threshold model | 1 | 1 |
| model set, per smooth model | 1 | 2 |

Two-block fits gain the most: the two `dpar` calls that built the stored
component curves are gone, and the combined curve is computed only when at
least one block is smooth. The one case that pays is a smooth model inside a
model set, which computes the matrix twice — once in `expand_nec()` for its NSEC
and once in `expand_manec()` for the weighted draws:

```
posterior_on_grid at 100 draws x 1000 grid: 0.205 s (median of 5)
```

Against fitting times measured in minutes, that is not a trade worth avoiding.

## Measured

The reproducer from the issue — `nec3param`, Gamma, 140 observations,
`chains = 4, iter = 2000, warmup = 1000`, `resolution = 1000`:

```
                        issue     this branch
whole bayesnecfit      31.8 MB      1.28 MB
  $fit (brmsfit)        1.2 MB      1.20 MB
  $pred_vals           30.5 MB      0.03 MB
  pred_vals$posterior  present      absent
  pred_vals$data       1000 x 4     1000 x 4
```

96% of the object is gone. At `model = "all"` and the package defaults that is
roughly 1.5 GB per `bnec()` call that is no longer retained, and the transient
peak inside `expand_manec()` goes with it rather than merely not being kept.

On the shipped `manec_example`, which was built before this change and still
carries the weighted matrix, dropping it alone takes the object from **1.335 MB
to 0.535 MB — 60%**, on only 100 draws. The earlier five-model measurement
(`nec3param`, `nec4param`, `ecx4param`, `ecxwb1`, `ecxll4`; 2 chains, 999
retained draws, `resolution = 1000`) put `w_pred_vals` at 7.65 MB of a 9.90 MB
object; removing it leaves roughly 2.25 MB.

## Breaking change, and the migration path

`x$pred_vals$posterior` and `x$w_pred_vals$posterior` now return `NULL` rather
than erroring, so downstream code fails a step later with something unhelpful
like `NULL[i, ]`. NEWS and both class docs now name the replacement:

```r
posterior_epred(x, newdata = bnec_newdata(x, resolution = 1000), re_formula = NA)
```

which works on both classes and, for a model set, draws from the component
models in the same weighted proportions.

Objects saved before this change still carry the matrices. Nothing needs them
and nothing trips over them, which is tested on both classes — the
`bayesmanecfit` case uses the shipped `manec_example`, a genuinely old
serialised object rather than a synthesised one.

## Downstream impact

Checked because of the silent-`NULL` failure mode:

- **`toxval`** — unaffected. Every live path calls `posterior_epred(object,
  newdata = ...)` fresh; its only slot reads are `$pred_vals$data$x` and
  `$w_pred_vals$data$x`, both of which survive.
- **`msPAF`** — unaffected. No reads of any `pred_vals` / `posterior` slot; it
  uses `bnec_newdata()` + `posterior_epred()`.
- **Vignettes** — unaffected. The only slot access across all six is
  `example2.Rmd:284`, `exp_5_nec$w_pred_vals$data`. `example4` is built on
  `compare_posterior()`, which computes fresh via `compare_fitted()`.

## How it was verified

- `test-expand_classes.R`: assertions on `pred_vals$posterior` and
  `w_pred_vals$posterior` replaced by assertions that both hold `data` alone;
  the model-averaged draws now checked by **count and membership** — each model
  contributes exactly `round(sample_size * wi)` rows, and every row it
  contributes is a row of that model's own posterior — rather than by a range
  check that any subsetting would have passed; backward-compatibility tests for
  old-style objects of both classes; and a test that a partially specified
  `x_range` is rejected at either end.
- **525 assertions passing, 0 failing** across `test-expand_classes.R`,
  `test-bayesmanec_methods.R`, `test-bayesnec_methods.R`, `test-helpers.R`,
  `test-compare_fitted.R`, `test-compare_posterior.R`, `test-compare_estimates.R`,
  `test-ecx.R`, `test-nsec.R`, `test-ecnsec.R`, `test-nec.R`,
  `test-average_estimates.R`, `test-pull_out.R`, `test-amend.R`,
  `test-bayesnechurdlefit-methods.R`, `test-bnec_hurdle.R`,
  `test-hurdle_family.R`.
- `bnec_newdata()` and `prediction_grid()` verified to return `identical()`
  grids.
- Full suite and R CMD check run on CI.

## Left undone, deliberately

- **#212** — `ggbnec_data.bayesnecfit()` rebuilds the fitted curve via
  `conditional_effects()` instead of reading `pred_vals$data`, which its
  `bayesmanecfit` sibling does. Pre-existing and untouched by this PR; changing
  it moves the plotted line (brms's 100-point grid and mean, against the bnec
  grid at `resolution` and a median), so it regenerates the vignette figures and
  belongs with #190.
- The shipped `manec_example` still carries `w_pred_vals$posterior`, since
  regenerating package data means refitting. It is harmless, is what the
  backward-compatibility test uses, and is a candidate for #190.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -280,9 +280,10 @@
   `pred_vals$data`, the small summary that `plot()` and `autoplot()` use, is
   unchanged. See [#180](https://github.com/open-AIMS/bayesnec/issues/180).
 - The matrix had exactly one reader in the package, the model-averaging path in
-  `expand_manec()`, which now builds its own for the duration of that call and
-  discards it. The weighted draws are the same. Nothing else read it:
-  `predict()` did not, and the plot methods use the summary.
+  `expand_manec()`, which now builds each model's posterior itself and thins it
+  to the weighted draws one model at a time. The weighted draws are the same.
+  Nothing else read it: `predict()` did not, and the plot methods use the
+  summary.
 - This is a net saving in computation, not a trade. `expand_nec()` used to
   compute the matrix for every model whether or not anything needed it; it now
   computes it only where it does — for smooth (`ecx`-type) models, whose NSEC is
@@ -290,11 +291,32 @@
   on its own no longer computes it at all. The one place it is computed twice is
   a model set containing smooth models, at about 0.2 s per model against fitting
   times measured in minutes.
-- Objects saved before this change still carry the matrix and are unaffected;
-  every accessor works on both, which is tested. Two things that were not
-  touched and remain: a `bayesmanecfit` still stores `w_pred_vals$posterior`,
-  the single weighted matrix, and a two-block fit still stores its two component
-  curves in `hurdle$mu_pred` and `hurdle$hu_pred`.
+- **This is a breaking change for code that read the matrix off a fitted
+  object.** `x$pred_vals$posterior` now returns `NULL` rather than erroring, so
+  downstream code will fail a step later. Replace it with
+  `posterior_epred(x, newdata = bnec_newdata(x, resolution = 1000), re_formula = NA)`,
+  which returns the same thing computed on demand. Objects saved before this
+  change still carry the matrix and are unaffected; every accessor works on
+  both, which is tested.
+- A `bayesmanecfit` no longer stores `w_pred_vals$posterior` either, the single
+  `sample_size x resolution` matrix of model-weighted draws. Once the per-model
+  matrices were gone this was 77% of what remained — 7.65 MB of a 9.90 MB
+  five-model set, and roughly 64 MB on its own at the package defaults. Nothing
+  in the package read it. `w_pred_vals$data` is unchanged, and the same
+  `posterior_epred()` call above reproduces the draws, sampling the component
+  models in the same weighted proportions. See
+  [#213](https://github.com/open-AIMS/bayesnec/issues/213).
+- A two-block fit no longer stores `hurdle$mu_pred` and `hurdle$hu_pred`, the
+  two component curves. They were written on every hurdle fit and never read
+  anywhere, so they are no longer computed at all — removing two
+  `posterior_epred()` calls per two-block fit, and four per `bnec_hurdle()`
+  pair. See [#214](https://github.com/open-AIMS/bayesnec/issues/214).
+- The prediction grid is now built in exactly one place. `bnec_newdata()`,
+  `expand_nec()` and the internal `posterior_on_grid()` had three separate
+  copies of the same code, which disagreed on a partially specified `x_range`:
+  `bnec_newdata()` silently ignored it, the other two produced `seq(NA, NA)`.
+  A partially specified `x_range` is now rejected with a clear error. See
+  [#211](https://github.com/open-AIMS/bayesnec/issues/211).
 
 # bayesnec 2.1.3.6
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -243,6 +243,30 @@
   parameter", which would be wrong for `zero_inflated_negbinomial`.
 - Neither family is selected automatically. A count response is still read as
   `poisson`; the zero-inflated forms have to be asked for.
+- A fitted object no longer stores `pred_vals$posterior`, the
+  `n_draws x resolution` matrix of posterior predictions over the plotting grid.
+  It dominated the size of every fit — **31.8 MB against a 1.2 MB `brmsfit`** at
+  the package defaults, 96% of the object — and the cost multiplied with the
+  model set and with `iter`: `model = "all"` retained roughly 1.5 GB in a single
+  `bnec()` call. The measured reproducer from the issue now returns **1.28 MB**.
+  `pred_vals$data`, the small summary that `plot()` and `autoplot()` use, is
+  unchanged. See [#180](https://github.com/open-AIMS/bayesnec/issues/180).
+- The matrix had exactly one reader in the package, the model-averaging path in
+  `expand_manec()`, which now builds its own for the duration of that call and
+  discards it. The weighted draws are the same. Nothing else read it:
+  `predict()` did not, and the plot methods use the summary.
+- This is a net saving in computation, not a trade. `expand_nec()` used to
+  compute the matrix for every model whether or not anything needed it; it now
+  computes it only where it does — for smooth (`ecx`-type) models, whose NSEC is
+  read off the curve, and for the two-block families. A threshold model fitted
+  on its own no longer computes it at all. The one place it is computed twice is
+  a model set containing smooth models, at about 0.2 s per model against fitting
+  times measured in minutes.
+- Objects saved before this change still carry the matrix and are unaffected;
+  every accessor works on both, which is tested. Two things that were not
+  touched and remain: a `bayesmanecfit` still stores `w_pred_vals$posterior`,
+  the single weighted matrix, and a two-block fit still stores its two component
+  curves in `hurdle$mu_pred` and `hurdle$hu_pred`.
 
 # bayesnec 2.1.3.6
 
--- a/R/bayesmanecfit-class.R
+++ b/R/bayesmanecfit-class.R
@@ -30,9 +30,17 @@
 #' data.
 #' @slot w_residuals Model-weighted residual values
 #' (i.e. observed - w_predicted_y).
-#' @slot w_pred_vals A \code{\link[base]{list}} containing model-weighted
-#' posterior predicted values based on the supplied \code{resolution} and
-#' \code{x_range}.
+#' @slot w_pred_vals A \code{\link[base]{list}} with a single element
+#' \code{data}, a \code{\link[base]{data.frame}} of summary model-weighted
+#' posterior predicted values over a grid set by \code{resolution} and
+#' \code{x_range}. Up to and including version 2.1.3.6 it also held
+#' \code{posterior}, the full \code{sample_size x resolution} matrix of
+#' weighted draws those summaries came from. Once the per-model matrices were
+#' dropped that one matrix dominated what remained, so it is no longer stored.
+#' To obtain it directly, use \code{posterior_epred(x, newdata =
+#' bnec_newdata(x, resolution = 1000), re_formula = NA)}, which draws from the
+#' component models in the same weighted proportions. Objects saved before the
+#' change still carry it and are unaffected.
 #' @slot w_ne The summary stats (median and 95% credibility intervals) of
 #' w_ne_posterior.
 #' @slot ne_type A \code{\link[base]{character}} vector indicating the type of
--- a/R/bayesnecfit-class.R
+++ b/R/bayesnecfit-class.R
@@ -24,11 +24,13 @@
 #' @slot pred_vals A \code{\link[base]{list}} with a single element
 #' \code{data}, a \code{\link[base]{data.frame}} of summary posterior
 #' predicted values over a grid set by \code{resolution} and \code{x_range}.
-#' Up to version 2.1.3.7 it also held \code{posterior}, the full
+#' Up to and including version 2.1.3.6 it also held \code{posterior}, the full
 #' \code{n_draws x resolution} matrix those summaries came from. That matrix
 #' dominated the size of a fitted object and is no longer stored; it is
-#' recomputed from \code{fit} where it is needed. Objects saved before the
-#' change still carry it and are unaffected.
+#' recomputed from \code{fit} where it is needed. To obtain it directly, use
+#' \code{posterior_epred(x, newdata = bnec_newdata(x, resolution = 1000),
+#' re_formula = NA)}. Objects saved before the change still carry it and are
+#' unaffected.
 #' @slot top The estimate for parameter "top" in the fitted model.
 #' @slot beta The estimate for parameter "beta" in the fitted model.
 #' @slot ne The estimated NEC.
--- a/R/bayesnecfit-class.R
+++ b/R/bayesnecfit-class.R
@@ -21,10 +21,14 @@
 #' to fit the model.
 #' @slot bayesnecformula An object of class \code{\link{bayesnecformula}} and
 #' \code{\link[stats]{formula}}.
-#' @slot pred_vals A \code{\link[base]{list}} containing a
-#' \code{\link[base]{data.frame}} of summary posterior predicted values
-#' and a vector containing based on the supplied \code{resolution} and
-#' \code{x_range}.
+#' @slot pred_vals A \code{\link[base]{list}} with a single element
+#' \code{data}, a \code{\link[base]{data.frame}} of summary posterior
+#' predicted values over a grid set by \code{resolution} and \code{x_range}.
+#' Up to version 2.1.3.7 it also held \code{posterior}, the full
+#' \code{n_draws x resolution} matrix those summaries came from. That matrix
+#' dominated the size of a fitted object and is no longer stored; it is
+#' recomputed from \code{fit} where it is needed. Objects saved before the
+#' change still carry it and are unaffected.
 #' @slot top The estimate for parameter "top" in the fitted model.
 #' @slot beta The estimate for parameter "beta" in the fitted model.
 #' @slot ne The estimated NEC.
--- a/R/bnec_newdata.R
+++ b/R/bnec_newdata.R
@@ -40,24 +40,12 @@ bnec_newdata <- function(x, resolution = 100, x_range = NA) {
 #' @export
 bnec_newdata.bayesnecfit <- function(x, resolution = 100, x_range = NA) {
   check_args_newdata(resolution, x_range)
-  data <- model.frame(x$bayesnecformula, data = x$fit$data)
-  x_var <- attr(data, "bnec_pop")[["x_var"]]
-  fit <- x$fit
-  x_vec <- fit$data[[x_var]]
-  if (any(is.na(x_range))) {
-    x_seq <- seq(min(x_vec), max(x_vec), length = resolution)
-  } else {
-    x_seq <- seq(min(x_range), max(x_range), length = resolution)
-  }
-  newdata <- data.frame(x_seq)
-  names(newdata) <- x_var
-  fam_tag <- fit$family$family
-  custom_name <- check_custom_name(fit$family)
-  if (fam_tag == "binomial" || fam_tag == "beta_binomial") {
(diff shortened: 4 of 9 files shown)


beckyfisher:
Retargeted from `issue-104-zero-inflated-counts` to `dev`, now that #204 has merged and that branch is deleted.

Merging `dev` in conflicted on `NEWS.md` — both sides appended bullets to the same point in the 2.1.3.7 section. Resolved in 18bd8c8 by keeping both, with the #104 bullets (including the \#209/#210 limitations note added during that review) ahead of the #180 ones. Additive on both sides; no text was dropped or reworded.

The diff against `dev` is unchanged in substance — still the same 6 files, +175/−25. `test-expand_classes.R`, `test-zero-inflated-counts.R`, `test-bnec_hurdle.R`, `test-define_prior.R` and `test-hurdle_family.R` all pass on the merged tree locally.

beckyfisher:
## Review notes

Claims in the description were checked against the branch source rather than taken at face value.

### Verified

- **"Exactly one reader" holds.** `pred_vals$posterior` was read only at `R/helpers.R:105` (`w_post_pred_calc`). `plot.R`, `autoplot.R`, `predict()`, `ecx()`, `nsec()`, `compare_posterior()` and `compare_fitted()` all build their own posterior or read `pred_vals$data`. Nothing else touched it.
- **`posterior_on_grid()` reproduces `expand_nec()`'s grid exactly** — same `is.na(x_range[1])` branch, same `x_var` extraction, same `trials = 1` for `binomial` / `beta_binomial`.
- **`pred_list` naming is safe.** `object <- object[success_models]` runs before the loop, so lengths and order match.
- **The memoised closure earns its keep.** A smooth two-block fit reaches both `mod_class == "ecx"` (L85) and the hurdle `else` branch (L124); without memoisation that is two full `posterior_epred` calls.

### Change requests

**1. Peak memory in `expand_manec()` is unchanged, and this is the one worth fixing.**

`pred_list` holds every model's full `n_draws x resolution` matrix simultaneously until `post_pred` is built. At `model = "all"` that is the same transient peak #180 complains about — no longer *retained*, which is the win, but the peak is still there and need not be.

Neither `mod_stats$wi` (from `loo_model_weights()` over `object`) nor `sample_size` (from `extract_simdat(object[[1]])`) depends on `pred_list`. So the `posterior_on_grid()` call can move out of the `expand_nec()` loop to the point where `post_pred` is built, computing and subsampling one model at a time — only `round(sample_size * wi)` rows per model ever held. Peak drops from sum-of-all-matrices to one matrix plus the result. Keeping it at the `do_wrapper()` call site also preserves RNG stream order, so the weighted draws stay bit-identical.

**2. The description overstates what `posterior_on_grid()` unifies.**

> New internal `posterior_on_grid()` holds the grid construction, so `expand_manec()` and `expand_nec()` cannot build different grids.

It does not. `expand_nec()` still builds `x_seq` / `new_dat` inline (`R/expand_classes.R:32-46`) and never calls `posterior_on_grid()` — it cannot, because it needs `new_dat` for `fitted()` and `x_seq` for `pred_data`.

Worse, the exported `bnec_newdata.bayesnecfit()` (`R/bnec_newdata.R:41-61`) already builds exactly this grid. So `posterior_on_grid()` is a **third** copy, not a consolidation — and the copies already disagree: `bnec_newdata()` tests `any(is.na(x_range))` while `expand_nec()` and `posterior_on_grid()` test `any(is.na(x_range[1]))`, which differ for `x_range = c(1, NA)`. The divergence pre-dates this PR but is now triplicated. Filed as #211; at minimum the claim in the description should be corrected.

**3. Version reference is off by one.** `R/bayesnecfit-class.R:27` and `man/bayesnecfit-class.Rd` say "Up to version 2.1.3.7 it also held `posterior`", but `DESCRIPTION` is `2.1.3.7` and the change lands in it. Should read 2.1.3.6.

**4. The migration path should be named.** The break is silent — `fit$pred_vals$posterior` now returns `NULL` rather than erroring, so downstream code fails a step later with something unhelpful like `NULL[i, ]`. `expand_nec()` is exported and this was a documented slot. NEWS and the class doc should give the replacement:

```r
posterior_epred(fit, newdata = bnec_newdata(fit, resolution = 1000), re_formula = NA)
```

**5. The new tests are weaker than they look.**

- `expect_gte(min(weighted), min(components))` passes for *any* subsetting of the component matrices, including one with wrong weights. It cannot fail the thing it is meant to protect.
- `expected_rows <- sum(round(n * tt5$mod_stats$wi))` reimplements the line under test.
- A sharper assertion: every row of `w_pred_vals$posterior` matches a row of exactly one component matrix, with `round(n * wi)` rows drawn from each. That actually pins "same draws, same proportions".

The old-object test is fine, though note it constructs the old style by calling `posterior_on_grid()` — it tests "an object carrying a matrix", not a genuinely old serialised object. Adequate for what it claims.

**6. Worth checking, not asserting.** In the hurdle branch `mu_curve` and `hu_curve` are already computed, and the comment above asserts `posterior_epred()` returns `mu * (1 - hu)`. If exact, the third call in `nsec_off_curve(get_pred_posterior())` is redundant. But that identity holds only for the continuous hurdle families — `hurdle_poisson` / `hurdle_negbinomial` use the zero-truncated mean `mu / (1 - exp(-mu)) * (1 - hu)` — so do not act on it blindly. It does mean the cost table under-reports two-block smooth models: three `posterior_epred` calls in `expand_nec()`, not one.

### Compatibility assessment

Checked because `toxval` and `msPAF` consume `bayesnec` objects, and because of the silent-`NULL` failure mode.

**What a fitted object still carries:**

| slot | before | after |
|---|---|---|
| `bayesmanecfit$w_pred_vals$posterior` | present | **still present** |
| `bayesmanecfit$w_pred_vals$data` | present | still present |
| `bayesnecfit$pred_vals$posterior` | present | **gone** |
| `bayesnecfit$pred_vals$data` | present | still present |
| `bayesmanecfit$mod_fits[[i]]$pred_vals$posterior` | present | **gone** |

**`toxval`** — no runtime impact. It carries its own copy of `w_post_pred_calc` at `R/helpers.R:95`, but it is dead code: nothing in the repo calls it (checked `R/`, tests, vignettes, `.Rd`). Every live path — `ecx.R:113/312/334`, `nsec.R:101/207/238`, `nsec_multi.R:88` — calls `posterior_epred(object, newdata = ...)` fresh. Its only other slot access, `return_x_range()` at `helpers.R:352-354`, reads `$pred_vals$data$x` / `$w_pred_vals$data$x`, both of which survive. Cleanup filed at beckyfisher/toxval.

**`msPAF`** — unaffected. Zero reads of any `bayesnec` `$pred_vals` / `$posterior` slot. Its touchpoints are `bnec_newdata()` + `posterior_epred()` in `hp_fun.R:41,53,56`, computed fresh. The `pred_vals` identifiers in `additive_predict.R` / `additive_hc.R` are its own local SSD variables — same name, unrelated object.

**`example4`** — unaffected. Built entirely on `compare_posterior()`, which for `comparison = "fitted"` routes to `compare_fitted()` and calls `posterior_epred` fresh at `R/compare_fitted.R:63` via `newdata_eval_fitted()`. It never reads `pred_vals`. Across all six vignettes the only slot access is `example2.Rmd:284`, `exp_5_nec$w_pred_vals$data` — a `bayesmanecfit` summary slot this PR does not change.

**Wider scan** — a tree-wide search for `pred_vals$posterior` across all local repos returned 25 hits, every one either a `bayesnec` branch clone, the dead `toxval` copy, or an archived copy. No live analysis script, `.Rmd` or `.qmd` anywhere reads it.

One consequence worth registering: eight `bayesnec-*` branch clones each carry `w_post_pred_calc` and the four `dim(pred_vals$posterior)` assertions, so each will conflict in `R/helpers.R` and `tests/testthat/test-expand_classes.R` when rebased onto `dev`. An argument for landing this while the outstanding branch stack is small.

### On the design intent behind the cache

Worth recording, since it bears on whether this is the right cut. The cache had two halves with very different value for money:

- `pred_vals$data` — about 30 KB, read by three of the four plotting paths (`plot.bayesnecfit`, `plot.bayesmanecfit`, `ggbnec_data.bayesmanecfit`). An excellent trade, and untouched here.
- `pred_vals$posterior` — about 30 MB, read by nothing that plots; one consumer, once, at fit time.

This PR keeps the good half and drops the bad half, which is the right cut. The fourth plotting path, `ggbnec_data.bayesnecfit`, rebuilds via `conditional_effects()` rather than reading the cached summary its `bayesmanecfit` sibling uses — pre-existing, unchanged by this PR, and filed separately as #212.

### Follow-up issues filed

- #211 — prediction-grid construction exists in three places and the copies disagree
- #212 — `ggbnec_data.bayesnecfit()` rebuilds instead of reading `pred_vals$data`
- #213 — `bayesmanecfit` still stores `w_pred_vals$posterior` (now 77% of what remains)
- #214 — `hurdle$mu_pred` / `hurdle$hu_pred` are written and never read

### Verdict

Approve after (1) and (3). (2) is a description-vs-code mismatch that should be resolved either way; (4) and (5) are cheap and improve it. CI was still pending at time of review.

beckyfisher:
Review points addressed in 6a7bf99d, and #211, #213 and #214 folded in. The description above has been rewritten to the current state.

**Change requests**

1. **Peak memory** — `expand_manec()` no longer accumulates `pred_list`. `w_grid_pred_calc()` computes each model's posterior and thins it to its `round(sample_size * wi)` rows in the same step, so only one model's full matrix exists at a time. Thinning still goes through `w_pred_list_calc()`, and the call sits at the same point in the sequence, so the draws are unchanged.
2. **Grid consolidation** — resolved properly rather than by correcting the claim. `prediction_grid()` is now the single definition; `bnec_newdata()`, `expand_nec()` and `posterior_on_grid()` all delegate to it. Verified `identical()` between the old `bnec_newdata()` output and the new one.
3. **Version reference** — now "up to and including version 2.1.3.6".
4. **Migration path** — `posterior_epred(x, newdata = bnec_newdata(x, resolution = 1000), re_formula = NA)` documented in NEWS and both class docs, with a test that it works on both classes.
5. **Tests** — the range assertion is gone. The weighted draws are now checked by count and membership: each model contributes exactly `round(sample_size * wi)` rows, and every row it contributes is a row of that model's own posterior. The `bayesmanecfit` backward-compatibility test uses the shipped `manec_example`, which is a genuinely old serialised object rather than a synthesised one.
6. **`mu * (1 - hu)`** — not acted on, deliberately. The identity holds only for the continuous hurdle families; `hurdle_poisson` / `hurdle_negbinomial` (#209) use the zero-truncated mean. #214 makes it moot anyway: the two `dpar` calls are gone entirely, so a two-block fit now makes at most one `posterior_epred` call on the grid rather than three.

**Folded-in issues**

- **#211** — one `prediction_grid()`; a partially specified `x_range` is now rejected at either end instead of being silently ignored by one code path and turned into `seq(NA, NA)` by another.
- **#213** — `w_pred_vals$posterior` dropped. On the shipped `manec_example` that alone is 1.335 MB → 0.535 MB (60%) at only 100 draws.
- **#214** — `hurdle$mu_pred` / `hu_pred` dropped, and no longer computed.

**Left open: #212.** `ggbnec_data.bayesnecfit()` rebuilding via `conditional_effects()` is pre-existing and untouched here. Changing it moves the plotted line — brms's 100-point grid and mean, against the bnec grid at `resolution` and a median — so it regenerates the vignette figures and belongs with #190.

**Verification.** 525 assertions passing, 0 failing, across 17 test files locally (R 4.6.1). CI pending on this push.

One loose end worth flagging: the shipped `manec_example` still carries `w_pred_vals$posterior`, since regenerating package data means refitting. Harmless, and it is what the backward-compatibility test uses, but it is a candidate for #190.

Issue #45, opened 2021-02-01
Add number of divergent transitions to summary()-output or $mod_stats


MNordborg:
I think it would be quite useful if it was possible to get a summary/overview of the number of divergent transitions for each candidate model, either when using summary() or in the $mod_stats, for use in assessing which candidate models are ok to include or not. Especially if someone (like me for example) are running model fitting for several MANECs overnight (eg both on the binomial and binomial distribution across a couple of data sets) and then doing the assessments the next day.

beckyfisher:
In the revised version these are now returned with summary().

Issue #33, opened 2020-09-11
Add capacity for a factor covariate


MNordborg:
At this time bnec does not allow inclusion of an interaction with a fixed factor. Including an interaction term within each of the non-linear models implemented in bayesnec is relatively straightforward, and may be introduced in future releases. However, in many cases the functional form of the response may change with different levels of a given factor. The substantial complexity of defining all possible non-linear model combinations at each factor level means it unlikely this could be feasibly implemented in bayesnec in the short term. In the meantime the greatest flexibility in the functional form of individual model fits can be readily obtained using models fitted independently to data within each factor level. The compare_posterior function allows these individual models fits to be compared.

beckyfisher:
At this time `bnec` does not allow inclusion of an interaction with a fixed factor. Including an interaction term within each of the non-linear models implemented in `bayesnec` is relatively straightforward, and may be introduced in future releases. However, in many cases the functional form of the response may change with different levels of a given factor. The substantial complexity of defining all possible non-linear model combinations at each factor level means it unlikely this could be feasibly implemented in `bayesnec` in the short term. In the meantime the greatest flexibility in the functional form of individual model fits can be readily obtained using models fitted independently to data within each factor level. The compare_posterior function allows these individual models fits to be compared.

beckyfisher:
## Feasibility of a `bnec_hurdle`-shaped route to a factor covariate

Assessment of whether the factorised two-fit structure now used by `bnec_hurdle()` /
`crossed_weights()` / `bnec_joint()` can be reused for a fixed factor: fit each factor level
independently (model-averaged within level), read crossed model weights off the per-level
weights, and refit the favoured combination jointly.

**Summary.** Both halves carry over. The likelihood factorisation and the outer-product weight
identity generalise from two blocks to *G* levels unchanged. The joint refit is expressible by
dummy-coding the levels and stringing the chosen per-level equations into a single non-linear
formula, which — unlike the alternative of putting the factor on each parameter as a fixed
effect — needs no new formula grammar and leaves most of the existing post-processing intact.

### 1. Factorised per-level fits and crossed weights

The argument `bnec_hurdle()` rests on holds verbatim. Factor levels partition the data disjointly
and share no parameters, so

```
log p(y) = Σ_g log p(y_g | θ_g)      ⇒      elpd(m_1,…,m_G) = Σ_g elpd_g(m_g)
```

[...]

Issue #19, opened 2020-06-03
Which functions to export?


dbarneche:
I think we only need to export the main ones:
- bnec
- plot.bayesnecfit
- predict.bayesnecfit
- plot.bayesmanecfit

The users are only going to be interacting with these four functions. The remaining ones are accessories to make these four above work, so I think they don't need to be accessible by them, particularly if the main target audience is novice. The more advanced people will know to use ::: to explore not-exported functions.

What do you think?

beckyfisher:
Sounds good to me Diego

dbarneche:
2cbd741471a14f2dc61fdcfe794e0dabdc9c7191

PR #217, opened 2026-08-19
Draw the model-averaging index once and keep it on the object

### LINKED ISSUE #216: Model-averaged output is not reproducible: predict() on a bayesmanecfit differs between calls
## Summary

Any function that goes through the model-averaged posterior of a `bayesmanecfit`
returns a **different answer on every call**. `predict()` is the clearest case:

```r
library(bayesnec)
nd <- bnec_newdata(manec_example, resolution = 20)
p1 <- predict(manec_example, newdata = nd)
p2 <- predict(manec_example, newdata = nd)

p1[1, "Estimate"]   #> 2.183761
p2[1, "Estimate"]   #> 2.229868      (2% apart)
p1[1, "Q2.5"]       #> 1.154437
p2[1, "Q2.5"]       #> 1.130791
```

No seed is set by the user and none is set internally, so this is not
reproducible even within a session. It propagates to `plot()`, `autoplot()`,
`summary()`, `ecx()`, `nsec()` and anything else built on the averaged
posterior.

## Cause

Model averaging is done by resampling each component's draws in proportion to
its stacking weight, using an unseeded `sample()`:

```r
# R/helpers.R:102
w_post_pred_calc <- function(index, mod_fits, sample_size, mod_stats) {
  x <- seq_len(sample_size)
  size <- round(sample_size * mod_stats[index, "wi"])
  mod_fits[[index]]$pred_vals$posterior[sample(x, size), ]
}
```

There are several such sites, each drawing **independently**:

| | |
|---|---|
| `R/helpers.R:90` | `w_nec_calc()` — the averaged no-effect posterior |
| `R/helpers.R:105` | `w_post_pred_calc()` — the averaged predicted curve |
| `R/helpers.R:112` | `w_pred_list_calc()` |
| `R/ecx.R:294` | `sample_ecx()` |
| `R/nsec.R:233-234` | `sample_nsec()` — **two** separate `sample()` calls, one for the NSEC draws and one for `ecnsec` |

The method itself is fine — this is `weighted_samples`, the same approach
`ssdtools` uses and recommends. The problem is only that the randomness is drawn
afresh, independently, at every site and on every call.

## Two distinct consequences

**1. Results are not reproducible.** Six replicate calls, `resolution = 50`, no
seed:

| | Q50 | Q2.5 (lower) | Q97.5 |
|---|---|---|---|
| `nsec(manec_example)` | 1.4727–1.4782 | **0.7023–0.9543** | 1.5469–1.5544 |
| `ecx(manec_example)` | 1.4638–1.5288 | 0.9091–0.9742 | 1.5939–1.5939 |

Medians are stable to ~0.5%, but the **lower bound swings by 26%** between
identical calls. That is the end of the interval used to set a protective
concentration.

**2. Quantities that should be paired are not.** Because each site samples
independently, realisation *i* of the NSEC draws and realisation *i* of the
`ecnsec` draws come from different MCMC iterations, and neither corresponds to
row *i* of the averaged curve. Nothing currently reads across them, so no
present output is wrong because of this — but it rules out any joint use of two
per-draw quantities.

## Suggested fix

Draw the component index **once** and reuse it everywhere, rather than sampling
per site and per call:

```r
idx <- sample(seq_along(models), n, replace = TRUE, prob = weights)   # once
```

Realisation *i* then means "component model `m[i]`, iteration `j[i]`" for every
quantity, which fixes both problems at once.

Since `sample_size` and `mod_stats$wi` are already stored on the
`bayesmanecfit`, the cleanest option is probably to **store the realised index
on the object when it is built**, so every later call reuses it and no `seed`
argument is needed. A user-facing `seed` argument would also work but leaves the
default non-reproducible.

This is not a change of averaging method — `weighted_samples` stays. It only
changes where the randomness is drawn.

## Why now

Two reasons beyond the bug itself.

- It blocks regression testing by equality. The planned reversal of the
  `bayesnec` / `toxval` dependency (open-AIMS/toxval#39) is specified as
  "estimates must be identical before and after", which cannot be checked for a
  `bayesmanecfit` while repeated calls disagree with themselves.
- The shared index is a prerequisite for computing N(S)EC as a first-class
  estimate (open-AIMS/toxval#25, #18), which needs to know, per realisation,
  which component model it came from — and therefore whether that draw is a NEC
  or an NSEC.

Note that several of the sites above (`sample_ecx`, `sample_nsec`) are expected
to be **deleted** rather than fixed under open-AIMS/toxval#39, since `ecx` and
`nsec` will be computed from a single averaged curve matrix. The sites in
`helpers.R` that build that matrix are the ones that need this fix.

beckyfisher:
Fixes #216.

## What changed

Model averaging keeps `round(sample_size * wi)` of each component model's
draws. Which draws those were was settled by an unseeded `sample()` at every
call site, independently — so `posterior_epred()`, `posterior_predict()`,
`ecx()`, `nsec()` and everything built on them returned a different answer on
every call, and none of them agreed with the summaries already stored on the
object.

The draw is now realised **once**, in `expand_manec()`, and kept on the object
as a new `w_draw_index` slot, with the seed behind it in `w_draw_seed`. Every
later call reuses it, so realisation *i* means "component *m*[*i*], iteration
*j*[*i*]" for every quantity taken from one object.

- `R/helpers.R` — new `weighted_draw_index()` and `pull_draw_index()`.
  `w_nec_calc()`, `w_pred_list_calc()` and `w_grid_pred_calc()` now take the
  index instead of drawing their own.
- `R/expand_classes.R` — draws once, uses one index for the no-effect posterior
  and the grid posterior, stores both index and seed.
- `R/posterior_epred.R`, `R/posterior_predict.R`, `R/ecx.R`, `R/nsec.R` — read
  the index off the object rather than redrawing.
- `R/bayesmanecfit-class.R` / `man/` — the new slots, documented.

Also fixed, in the same function this moved the draw out of: `w_pred_list_calc()`
indexed with the default `drop = TRUE`, so a **single grid column** — a one-row
`newdata` — dropped to a vector and `rbind` stacked each model's draws as a
*row*. `fitted(manec_example, newdata = data.frame(x = 3))` returned 83 rows
instead of 1, with `Est.Error` and the quantiles computed across models rather
than over draws, so the point estimate looked plausible and its interval did
not. A `bayesnecfit` was unaffected, as was any `newdata` with more than one
row. Pre-existing on `dev`; the `resolution = 1` test added here walked into it.

The averaging method is unchanged. This is `weighted_samples` exactly as
before; only where the randomness is drawn has moved.

### Why the index is stored, not just the seed

A seed alone does not fix a draw. R 3.6.0 changed `sample()`'s algorithm, and
the same seed gives a different index either side of that change. A
`bayesmanecfit` archived now and reopened under a future R would silently
rebuild a different index and stop matching its own stored summaries — the same
bug, merely deferred. So:

- **`w_draw_index`** holds the realised indices and is what every call uses. It
  cannot drift. It costs one integer per draw (a few KB).
- **`w_draw_seed`** is the fallback, for the two cases the stored index cannot
  cover: a caller thinning to a different number of draws (`ndraws`,
  `draw_ids`), and objects saved before either field existed.
- `weighted_draw_index()` pins `sample.kind = "Rejection"` so that fallback is
  itself stable across R versions. "Rejection" is the post-3.6.0 default, so
  this changes nothing today. `RNGkind()` and `.Random.seed` are both restored
  afterwards, so averaging never disturbs a user's simulation stream.

`manec_example` predates all of this; it falls back to a fixed seed and was left
as-is rather than refitted.

### `ecx()` and `nsec()`

Included, overriding the original scope note. Those functions are headed for
`toxval`, but that migration is gated on #45 reaching CRAN, and until then
`bayesnec` ships irreproducible ECx and NSEC on model-averaged fits — which are
the numbers that go into reports. Six replicate calls before this change:

| | median | lower (Q2.5) |
|---|---|---|
| `nsec()` | 1.475–1.482 | 0.735–0.844 |
| `ecx()` | 1.491–1.496 | 0.875–0.963 |

The instability lands almost entirely on the lower bound, which is the end a
protective concentration is read off.

The same index also covers both halves of the `nsec()` result. Previously the
NSEC and its `ecnsec` came from two independent `sample()` calls, so a pair was
two unrelated draws of possibly two different models; element *i* of each is now
draw `idx[i]` of the same component.

This preserves current semantics — each NSEC is still read off its own model's
curve exactly as before — so it settles nothing about what those curves are
anchored to, and does not pre-empt #19.

## Verified

- `posterior_epred()`, `fitted()`, `ecx()` and `nsec()` on `manec_example` are
  identical across repeated calls.
- `nsec()`'s posterior and its `ecnsec_relativeP` reconstruct exactly from the
  per-model posteriors at the shared index — asserted, not inferred.
- `posterior_epred()` over the build grid now reproduces the stored
  `w_pred_vals` **exactly**. Before this change the two were independent redraws
  and agreed only in distribution — so the migration path promised in the
  `w_pred_vals` docs (added under #180) now actually returns the same draws.
- The caller's RNG stream and `RNGkind()` are both untouched.
- Under `RNGkind(sample.kind = "Rounding")` the index is unchanged, where an
  unpinned draw from the same seed would differ.
- Full suite: **1448 passed, 0 failed, 0 errors, 0 skipped**.

No existing test needed rewriting. The hazard note anticipated pinned
model-averaged constants, but nothing in the suite pins one — the existing
tests check the arithmetic of the weights, not stored values.

## Tests added

`tests/testthat/test-helpers.R`
- `weighted_draw_index()` is a pure function of its arguments: same seed → same
  index, different seed → different index, missing seed (legacy object) still
  self-consistent, sizes equal `round(n * wi)`, indices in range.
- It leaves the caller's RNG stream alone, and ignores the session's
  `sample.kind` — with the pre-3.6.0 draw computed inline to show it would
  otherwise differ.
- `pull_draw_index()` prefers the stored index, tolerates the integer/double
  mismatch between `nrow()` and the stored `sample_size`, falls back in range
  when a caller thins, and handles an object carrying neither field.
- Edge case: a model with `wi == 0` contributes no rows rather than erroring.
  `loo_model_weights()` returns exact zeros often enough for this to be a real
  path.

`tests/testthat/test-bayesmanec_methods.R`
- Repeated `posterior_epred()` / `fitted()` calls are identical.
- The single-grid-column case, asserted on **shape** as well as on agreement:
  two equally malformed results compare identical, so reproducibility alone did
  not catch the dropped dimension — it hid it. Covers both `resolution = 1` and
  the way a user actually meets it, a one-row `newdata`, checking the result is
  silent, is `sample_size x 1`, matches the shape of the single-model answer, and
  that the averaged estimate falls between the components it averages.
- `ecx()` and `nsec()` are reproducible, and the NSEC/`ecnsec` pairing is
  reconstructed from the component posteriors.

`tests/testthat/test-expand_classes.R`
- `posterior_epred()` on the build grid reproduces `w_pred_vals`.
- The existing `w_grid_pred_calc()` test updated for the new signature.

## One finding

**`predict()` / `posterior_predict()` are still not reproducible without
`set.seed()`, and that is not model averaging.** `predict()` on a single
`bayesnecfit` varies just as much — 2.100 vs 2.219 on `nec4param` alone —
because `brms::posterior_predict()` simulates new observations from the
likelihood. The issue's reproducer used `predict()`, which conflates the two
sources. The weighting half is now fixed; with a seed set, `predict()` on
`manec_example` reproduces exactly. I did **not** freeze the likelihood
simulation as well: that would make `bayesmanecfit` behave unlike both
`bayesnecfit` and plain `brms`, which is a user-visible behaviour change beyond
this issue.

Same applies to `ndraws` / `draw_ids`: `brms` picks that subset at random, so
`posterior_epred(x, ndraws = 40)` varies between calls for a single
`bayesnecfit` too. Our weighting index over that subset is deterministic.

## Not done

- `average_estimates.R:85`, `compare_estimates.R:75` and `compare_fitted.R:67`
  each hold a further unseeded `sample()`. Investigated and **filed separately as
  #218** rather than fixed here: those are full permutations, not weighted
  thinning, and their job is to pair draws across *independently fitted* models,
  so the fix in this PR does not transfer — there is no persistent object to hang
  an index on, and a session `set.seed()` already makes them reproducible, which
  it could not do here. #218 records the measurements, and the condition under
  which the permutation stops being correct (a factor covariate, #33).
- Vignettes not rebuilt (#190). Model-averaged numbers in them will shift when
  they are.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,29 @@
 # bayesnec 2.1.3.7
 
+- Model-averaged output from a `bayesmanecfit` is now reproducible. Averaging
+  keeps `round(sample_size * wi)` of each component model's draws, and which
+  draws those were used to be settled by an unseeded `sample()` at every call
+  site independently. `posterior_epred()`, `posterior_predict()` and everything
+  built on them therefore returned a different answer on every call, and none of
+  them agreed with the summaries stored on the object. The draw is now realised
+  once, when the object is built, and the seed that produced it is kept on the
+  object as the new `w_draw_seed` slot; every later call rebuilds the same
+  index. Realisation *i* now means "component *m*[*i*], iteration *j*[*i*]" for
+  every quantity taken from one object. The averaging method itself is
+  unchanged — only where the randomness is drawn. Two consequences worth
+  knowing: model-averaged numbers will differ from those produced by earlier
+  versions (the old ones were not reproducible, so there is no "before" to
+  match), and `posterior_epred()` over the build grid now returns exactly the
+  draws that `w_pred_vals` summarises rather than an independent redraw of them.
+  Objects saved before this version carry no `w_draw_seed` and fall back to a
+  fixed value, which is equally reproducible. `predict()` and
+  `posterior_predict()` still vary between calls unless a seed is set, because
+  they simulate new observations from the likelihood — the same behaviour
+  `brms` has for a single fit, and unrelated to model averaging. `ecx()` and
+  `nsec()` on a `bayesmanecfit` also still vary; their weighted resampling lives
+  in `R/ecx.R` and `R/nsec.R`, which are migrating to `toxval` and were left
+  untouched. See #216.
+
 - `brms (>= 2.23.0)` is now required. Earlier versions generate the
   `beta_binomial` likelihood by passing the whole `trials` array to
   `beta_binomial_lpmf` instead of `trials[n]`, so each response is evaluated
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,26 +3,29 @@
 - Model-averaged output from a `bayesmanecfit` is now reproducible. Averaging
   keeps `round(sample_size * wi)` of each component model's draws, and which
   draws those were used to be settled by an unseeded `sample()` at every call
-  site independently. `posterior_epred()`, `posterior_predict()` and everything
-  built on them therefore returned a different answer on every call, and none of
-  them agreed with the summaries stored on the object. The draw is now realised
-  once, when the object is built, and the seed that produced it is kept on the
-  object as the new `w_draw_seed` slot; every later call rebuilds the same
-  index. Realisation *i* now means "component *m*[*i*], iteration *j*[*i*]" for
-  every quantity taken from one object. The averaging method itself is
-  unchanged — only where the randomness is drawn. Two consequences worth
-  knowing: model-averaged numbers will differ from those produced by earlier
-  versions (the old ones were not reproducible, so there is no "before" to
-  match), and `posterior_epred()` over the build grid now returns exactly the
-  draws that `w_pred_vals` summarises rather than an independent redraw of them.
-  Objects saved before this version carry no `w_draw_seed` and fall back to a
-  fixed value, which is equally reproducible. `predict()` and
+  site independently. `posterior_epred()`, `posterior_predict()`, `ecx()`,
+  `nsec()` and everything built on them therefore returned a different answer on
+  every call, and none of them agreed with the summaries stored on the object.
+  For `nsec()` the instability landed almost entirely on the lower bound, which
+  is the end used to set a protective concentration: over six replicate calls it
+  spanned 0.735–0.844 while the median moved by less than 0.5%. The draw is now
+  realised once, when the object is built, and kept on the object as the new
+  `w_draw_index` slot, with the seed behind it in `w_draw_seed`; every later call
+  reuses it. Realisation *i* now means "component *m*[*i*], iteration *j*[*i*]"
+  for every quantity taken from one object. The averaging method is unchanged —
+  only where the randomness is drawn. Three consequences worth knowing.
+  Model-averaged numbers will differ from those produced by earlier versions;
+  the old ones were not reproducible, so there is no "before" to match.
+  `posterior_epred()` over the build grid now returns exactly the draws that
+  `w_pred_vals` summarises rather than an independent redraw of them. And an
+  `nsec()` and its `ecnsec` are now the same draw of the same component model,
+  where before they came from two independent `sample()` calls and a pair could
+  be two unrelated draws of two different models. `predict()` and
   `posterior_predict()` still vary between calls unless a seed is set, because
-  they simulate new observations from the likelihood — the same behaviour
-  `brms` has for a single fit, and unrelated to model averaging. `ecx()` and
-  `nsec()` on a `bayesmanecfit` also still vary; their weighted resampling lives
-  in `R/ecx.R` and `R/nsec.R`, which are migrating to `toxval` and were left
-  untouched. See #216.
+  they simulate new observations from the likelihood — the same behaviour `brms`
+  has for a single fit, and unrelated to model averaging. Objects saved before
+  this version carry neither field and fall back to a fixed seed, which is
+  equally reproducible. See #216.
 
 - `brms (>= 2.23.0)` is now required. Earlier versions generate the
   `beta_binomial` likelihood by passing the whole `trials` array to
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,15 @@
   this version carry neither field and fall back to a fixed seed, which is
   equally reproducible. See #216.
 
+- Model-averaged predictions from a `bayesmanecfit` no longer collapse when
+  `newdata` has a single row. Asking what the averaged curve predicts at one
+  concentration — `fitted(fit, newdata = data.frame(x = 3))` — returned one row
+  per retained draw instead of one row per prediction, because the weighted row
+  selection dropped to a vector and the per-model results were then stacked as
+  rows rather than columns. The point estimate looked plausible, but its
+  `Est.Error` and quantiles were computed across models rather than over draws.
+  A `bayesnecfit` was unaffected, as was any `newdata` with more than one row.
+
 - `brms (>= 2.23.0)` is now required. Earlier versions generate the
   `beta_binomial` likelihood by passing the whole `trials` array to
   `beta_binomial_lpmf` instead of `trials[n]`, so each response is evaluated
--- a/R/bayesmanecfit-class.R
+++ b/R/bayesmanecfit-class.R
@@ -24,6 +24,16 @@
 #' @slot mod_stats A \code{\link[base]{data.frame}} of model fit statistics.
 #' @slot sample_size The size of the posterior sample.
 #' Information on the priors used in the model.
+#' @slot w_draw_seed The seed that realised the model-averaging draw. Model
+#' averaging keeps a weighted subset of each component model's draws; this seed
+#' fixes which ones, so every call on the object uses the same subset.
+#' \code{fitted} and \code{posterior_epred} are therefore reproducible, and
+#' agree with \code{w_ne_posterior} and \code{w_pred_vals}. \code{predict}
+#' and \code{posterior_predict} still vary between calls unless a seed is set,
+#' because they simulate new observations from the likelihood, as they do for a
+#' single \code{\link[brms]{brmsfit}}. Objects saved before version 2.1.3.7 do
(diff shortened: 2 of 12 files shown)


beckyfisher:
## `w_pred_list_calc()` drops a dimension when the prediction grid has one column

Found while reviewing this PR. **Pre-existing on `dev`, not caused by this
change** — but it lands in the one function this PR rewrites, and the new
`resolution = 1` test walks straight into it, which is where the two testthat
warnings in the suite come from.

### What happens

`w_pred_list_calc()` indexes with R's default `drop = TRUE`:

```r
pred_list[[index]][draw_index[[index]], ]
```

`pred_list[[index]]` is a *draws x grid-points* matrix and the result is stacked
across models with `rbind`. When the matrix has exactly **one column** the
selection returns a *vector*, so `rbind` binds each model's draws as a **row**
instead of a column, recycling the shorter one.

The realistic trigger is not `resolution = 1` — it is a **single-row `newdata`**,
i.e. "what does the curve predict at concentration 3?":

```r
fitted(manec_example, newdata = data.frame(x = 3))
```

| | `bayesmanecfit` | `bayesnecfit` |
|---|---|---|
| `posterior_epred` | **2 x 83** | 100 x 1 |
| `fitted` | **83 x 4** | 1 x 4 |
| `predict` | **83 x 4** | 1 x 4 |

You ask for one prediction and get 83 rows, with only a recycling warning to say
so. It does not look broken either: `Estimate` lands near the right value
(-3.8 against a true -3.71), because the entries are still `epred` values — but
`Est.Error` and the quantiles are computed from **2 numbers**, so they are
meaningless and swing row to row (Est.Error 0.77, 0.62, 0.15 against a true
0.40). Multi-row `newdata` is unaffected, which is presumably why it has gone
unnoticed.

### Fix

```diff
-  pred_list[[index]][draw_index[[index]], ]
+  pred_list[[index]][draw_index[[index]], , drop = FALSE]
```

Verified on this branch (`d4b9578`), R 4.6.1, brms 2.23.0:

```
posterior_epred, 1-row newdata  ->  100 x 1
fitted, 1-row newdata           ->  1 x 4
predict, 1-row newdata          ->  1 x 4      (no warnings)
```

and the value is right: **-3.697**, sitting between the two components
(`nec4param` -3.599 at *wi* 0.827, `ecx4param` -4.307 at *wi* 0.173) and matching
the object's own stored `w_pred_vals` at x = 3 (-3.636) to within Monte Carlo
noise. Multi-row output is unchanged.

Tests on the three files this PR touches, with `NOT_CRAN=true`:

| file | before | with `drop = FALSE` |
|---|---|---|
| `test-bayesmanec_methods.R` | 33 pass, **2 warn** | 33 pass, 0 warn |
| `test-helpers.R` | 25 pass, 0 warn | 25 pass, 0 warn |
| `test-expand_classes.R` | 94 pass, 0 warn | 94 pass, 0 warn |

### Suggestion

Worth folding into this PR rather than filing separately: it is one word in a
function this PR already rewrites, it clears the two warnings the new test
introduces, and it makes that `resolution = 1` assertion actually test
something — at present both sides of the `expect_identical()` are equally
malformed, so it passes without checking the shape it was added to check.

An explicit `expect_equal(dim(...), c(100L, 1L))` alongside it would stop the
same thing recurring.

beckyfisher:
Applied in 6eb3a23d.

- `R/helpers.R` — `w_pred_list_calc()` now indexes with `drop = FALSE`, with a
  comment recording why it is load bearing.
- `tests/testthat/test-bayesmanec_methods.R` — the `resolution = 1` case now
  asserts the **shape**, not just agreement between calls, and adds the way a
  user actually meets it: a one-row `newdata`. That one checks the result is
  silent, is `sample_size x 1`, has the same shape as the single-model answer,
  and that the averaged estimate falls between the components it averages.
- `NEWS.md` — an entry, since this is a user-visible wrong answer rather than an
  internal tidy-up.

`test-bayesmanec_methods.R` with `NOT_CRAN=true`: **39 pass, 0 fail, 0 warn**
(was 33 pass, 2 warn). Full suite running.

Issue #215, opened 2026-08-18
Build and publish vignettes for both dev and master


beckyfisher:
## Problem

Vignettes are precompiled from `vignettes/*.Rmd.orig` into `vignettes/*.Rmd` and
the rendered articles are published from a single branch. There is no built
vignette output for `dev`, so:

- a vignette written or changed on `dev` cannot be read in its rendered form
  without a contributor precompiling and rendering it locally;
- reviewers cannot see figures, tables or resolved citations on a PR, which is
  where most vignette errors actually show up (a `\@ref()` that renders
  literally, a figure with missing axis labels, an unresolved `@citation`);
- drift between the code on `dev` and the vignette output published from
  `master` is invisible until release. Issue #190 is an instance of exactly
  this.

## Suggested change

Publish two sets of rendered articles:

- `master` -> the released documentation, as now;
- `dev` -> a development preview, at a distinct URL prefix
  (for example `articles/dev/`) so the two cannot be confused.

Ideally also build the vignettes on pull requests and attach or link the
rendered output, so a reviewer can read the article the PR produces.

## Notes

- `vignettes/precompile.R` re-knits **every** `.Rmd.orig` in the directory.
  For CI that is the right behaviour, but it means the job needs the full
  fitting stack (`cmdstanr` plus a working toolchain) and is not quick. Caching
  the compiled Stan models across runs is worth doing.
- Precompilation needs `NOT_CRAN=true`; without it every chunk is skipped and
  the build silently produces empty vignettes.
- Figures are written to the working directory and moved into `vignettes/` by
  `precompile.R`; any CI job needs that step too or the images will 404.

beckyfisher:
Merged as #220. `master` publishes to the site root and `dev` to a separate prefix, with a manually-dispatched workflow to rebuild the precompiled vignettes.

Follow-up on the mechanism: #230 records that `development: mode: auto` does not distinguish the two branches the way the config comment assumes, along with the missing workflow `permissions:` blocks and the no-op Stan cache.

Issue #84, opened 2021-09-01
Consider adding an argument to fix parameter values in models


darrenkoppel:
Many concentration-response curves I generate use data normalised to a control response. This gives response values in percent with controls defined as being 100%. As such it would be useful to fix a model's upper asymptote to 100% (or 1).  This would be equivalent to the 'fixed' argument in the drc package. 

![image](https://user-images.githubusercontent.com/29590117/131628399-e0c06e97-908b-4907-b990-ae8ae72be938.png)

Thanks!

beckyfisher:
Hi Darren, can you provide a reproducible example of doing this in drc?

darrenkoppel:
Hi Becky, 

Sure thing, here's an example using the earthworms dataset from drc package:
```{r}
library(drc)
data<-earthworms
str(data)

```
These data are from an avoidance bioassay - two containers are set next to each other, one containing contaminated soil and the other uncontaminated soil. If the earthworms don't like the contaminated soil they will move to the uncontaminated soil. 
The set-up is replicated with increasing doses of a contaminant to give a concentration - response gradient. The number of earthworms in the contaminated soil at the end of the bioassay (number) is compared to the number at the start (total) to give an earthworm proportion in the contaminated soil. At dose 0 the expected response should be 0.5 (no avoidance) between contaminated and uncontaminated soils.

We can fit a a 3-parameter log-logistic model: 

```{r}
model_1 <- drm(number/total~dose, weights = total, data = data,
                     fct = LL.3(), type = "binomial")
plot(model_1)
```
The upper limit/asymptote is a parameter estimated by the model, but we know _a priori_ that at dose 0 the number/total should be 0.5. So we can use the 'fixed' argument of drm to set that model parameter to 0.5:

```{r}
model_2 <- drm(number/total~dose, weights = total, data = data,
               fct = LL.3(fixed=c(NA,0.5,NA)), type = "binomial")
plot(model_2)
```

beckyfisher:
Thanks Darren

beckyfisher:
Hi Darren, apologies Diego and I have been too busy to implement this as yet. However, I have worked out some code to show you how to use bayesnec to fit the desired model, pull the brms formula and priors, and then refit using brms, using the constant. You can do this with:
`
library(drc)
data<-earthworms
str(data)

fit <- bnec(number | trials(total) ~ crf(dose, "ecxll3"),
               data, iter = 2e3, chains=1)
[...]

PR #221, opened 2026-08-21
Document how drc's NEC models map onto the bayesnec set (#139)

### LINKED ISSUE #139: Add NEC models from drc
The drc package includes an NEC function, that fits 2, 3 and 4 parameter NEC models. All 3 differ slightly from the original formulation of the exponential decay model in Fox 2010/Peres that is currently implemented.
The model formula is:
f (x) = c + (d − c) exp(−b(x − e)I(x − e)) + d2/(1 + exp(b2(log(x) − log(e2))))

With three variations:
NEC.2(upper = 1, fixed = c(NA, NA), names = c("b", "e"), ...)
NEC.3(fixed = c(NA, NA, NA), names = c("b", "d", "e"), ...)
NEC.4(fixed = c(NA, NA, NA, NA), names = c("b", "c", "d", "e"), ...)

The parameter e in NEC in "drc" corresponds to the parameter c’ in Pires et al (2002), the parameter b in NEC in "drc" corresponds to the parameter m’ in Pires et al (2002), the parameter d in NEC in "drc" corresponds to the parameter l’ in Pires et al (2002), and finally the parameter c in NEC in "drc" (the lower horizontal limit) is (implictly) fixed at 0 in Pires et al (2002).
More info at: https://cran.r-project.org/web/packages/drc/drc.pdf

beckyfisher:
**Release tier: 2.1.4.** Item 2 of the stack. **Stacked on #220** (`#215`) — base
is `issue-215-dev-vignette-ci`, so this diff shows only the #139 work. GitHub
retargets it to `dev` when #220 merges.

Closes #139, on RF's choice of **option A** (documentation only).

## What changed

`?models` and `vignette("example2b")` now document how `drc`'s `NEC.2()`,
`NEC.3()` and `NEC.4()` map onto the `bayesnec` model set:

| `drc` | fixes | `bayesnec` |
|---|---|---|
| `NEC.4()` | nothing | `nec4param` |
| `NEC.3()` | `c = 0` | `nec3param` |
| `NEC.2()` | `c = 0`, `d = upper` | none, by choice |

with `c = bot`, `d = top`, `e = nec`, `b = exp(beta)`; the one substantive
difference (our decay rate is positive by construction, so `drc`'s `b < 0`
region — a threshold followed by unbounded growth — is unreachable here, while a
threshold followed by an *increase* is available as `nec4param` with
`bot > top`); that the extra log-logistic term in `?drc::NEC` reproduces
Pires et al. (2002) but is **not** in the code `drc` fits; and why `NEC.2()` has
no equivalent, pointing at #84.

## Two corrections to the scoping comment, both found by testing its claims

**1. "Exactly 0" was true only at the one parameter set it tested.** Over a grid
varying every parameter I got a maximum absolute difference of **7.1e-15**. That
turned out to be the *test* round-tripping `b` through `exp(log(b))`, not any
difference between the models — `exp(log(b))` is not the identity in floating
point. Compared the way the reparameterisation actually runs, with
`b = exp(beta)`, the two are **bit-exact**.

Both the documentation and the tests now say the precise thing, and there is a
test pinning the round-trip error at ~1e-15, so a `drc` user who converts an
estimate via `log(b)` and sees a discrepancy knows what it is rather than
concluding the two packages disagree.

**2. `check_normalisation()` is not exported**, so the `\link{}` the first draft
used would have been a broken cross-reference under `R CMD check`. Replaced with
plain text. Flagging it because the rest of the stack will want the same care:
the internal check functions are not all exported.

## Tests

New `tests/testthat/test-pred_equations.R`, 13 assertions.

`drc` is not a dependency and adding one to verify a *documentation* claim would
be a poor trade, so its generator is transcribed inline from `drc:::NEC()$fct`.
That also makes the comparison explicit rather than hiding it behind a package
version.

- `nec4param` is exactly `NEC.4`, and `nec3param` exactly `NEC.3` (`tolerance = 0`).
- The equivalence holds across a grid of every parameter, including a threshold
  below the predictor range (step always fires) and one above it (never fires).
- The round-trip error is bounded and non-zero — the edge case worth pinning.
- `exp(beta)` cannot be negative, so the decline is monotone for any `beta`,
  which is the reason `drc`'s `b < 0` case is documented as a deliberate
  omission.

## Verified

|---|---|---|---|---|
| `dev` baseline (post-#217) | 0 | 11 | 0 | 1457 |
| this branch | 0 | 11 | 0 | **1470** |

+13, exactly the new test file; warnings unchanged.

## Left undone, deliberately

**Option C — the Pires et al. (2002) two-component model — is not implemented
and #139 should not be read as closing it.** It is genuinely absent from both
`bayesnec` and `drc`, and it separates a hard threshold from a smooth background
decline, which none of our equations does. Whether it is a real proposal or an
Rd copy-paste artefact cannot be settled from the package alone and needs
*Environmetrics* 13:15–27 read. Worth raising as its own issue.

**Option B (`nec2param`) is off the table**, on the #84 grounds rather than on
cost — it is a two-line equation. Recorded in the docs so the question does not
come back.

Vignette not rebuilt; `example2b.Rmd` stays stale until #190.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: bayesnec
 Title: A Bayesian No-Effect- Concentration (NEC) Algorithm
-Version: 2.1.3.8
+Version: 2.1.3.9
 Authors@R: c(person("Rebecca", "Fisher", email = "r.fisher@aims.gov.au", role = c("aut", "cre"),comment = c(ORCID = "0000-0001-5148-6731")), person("Diego R.","Barneche",role="aut",comment = c(ORCID = "0000-0002-4568-2362")), person("Gerard F.","Ricardo",role="aut",comment = c(ORCID = "0000-0002-2220-3971")), person("David R.","Fox",role="aut",comment = c(ORCID = "0000-0002-3178-7243")))
 Description: Implementation of No-Effect-Concentration estimation that uses 'brms' (see Burkner (2017)<doi:10.18637/jss.v080.i01>; Burkner (2018)<doi:10.32614/RJ-2018-017>; Carpenter 'et al.' (2017)<doi:10.18637/jss.v076.i01> to fit concentration(dose)-response data using Bayesian methods for the purpose of estimating 'ECx' values, but more particularly 'NEC' (see Fox (2010)<doi:10.1016/j.ecoenv.2009.09.012>), 'NSEC' (see Fisher and Fox (2023)<doi:10.1002/etc.5610>), and 'N(S)EC (see Fisher et al. 2023<doi:10.1002/ieam.4809>). A full description of this package can be found in Fisher 'et al.' (2024)<doi:10.18637/jss.v110.i05>. This package expands and supersedes an original version implemented in 'R2jags' (see Su and Yajima (2020)<https://CRAN.R-project.org/package=R2jags>; Fisher et al. (2020)<doi:10.5281/ZENODO.3966864>).
 Depends:
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,16 @@
 # bayesnec 2.1.4
 
+- `?models` and `vignette("example2b")` now document how the `drc` package's
+  `NEC.2()`, `NEC.3()` and `NEC.4()` map onto the `bayesnec` model set.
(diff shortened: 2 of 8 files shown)


beckyfisher (commented):
This all looks ok.

Issue #209, opened 2026-08-17
Add hurdle_poisson and hurdle_negbinomial as two-block families


beckyfisher:
## Summary

Add `hurdle_poisson` and `hurdle_negbinomial` as two-block families, the count
analogues of `hurdle_gamma`. These are the "two processes modelled separately"
case for count responses, and #104 does not provide it — the zero-inflated count
families added there deliberately take the single-block path with a constant
`zi`.

## Why this is the family that was actually missing

#104 added `zero_inflated_poisson` / `zero_inflated_negbinomial` and, correctly,
refused them a second bayesnec block: `zi` and `mu` are not separately
identified from the zeros, so a curve on `zi` is weakly informed exactly where
`mu` is small — the high-concentration end that determines the NEC. `zi` is also
a latent class, so a threshold on it describes something that was never
observed.

None of that applies to a hurdle on counts. Where the zeros are *observed* to be
structural — the individual died, the replicate failed — every zero belongs to
the survival process by construction, the likelihood factorises exactly, and
both blocks carry an interpretable concentration-response curve. That is the
same situation `hurdle_gamma` already serves, differing only in the family of
the non-zero part.

brms has both families, with `hu` as a predictable distributional parameter,
exactly parallel to `hurdle_gamma`:

```
hurdle_gamma       | dpars: mu, shape, hu
hurdle_poisson     | dpars: mu, hu
hurdle_negbinomial | dpars: mu, shape, hu
```

and the Stan density it generates factorises, with the positive part properly
zero-truncated (the `- log1m_exp(-lambda)` term is the truncation normaliser):

```stan
real hurdle_poisson_lpmf(int y, real lambda, real hu) {
    return bernoulli_lpmf(1 | hu);                             // y == 0
    return bernoulli_lpmf(0 | hu) +
           poisson_lpmf(y | lambda) - log1m_exp(-lambda);      // y > 0
}
```

## What this also fixes

`bnec_hurdle()` currently fits the growth component with an **untruncated**
count family on `data[y > 0, ]`. For `hurdle_gamma` that is exact, because the
Gamma has no mass at zero. For counts it is not: fitting `Poisson(mu)` to data
conditioned on `y > 0` estimates `mu / (1 - exp(-mu))` rather than `mu`. The
bias is negligible for large `mu` but grows as `mu` falls towards zero — again
the high-concentration end where the NEC and ECx are read off.

So today there are two count hurdle routes and neither is right:

[...]

PR #222, opened 2026-08-21
Guard define_prior against a response that is mostly zeros (#210)

### LINKED ISSUE #210: define_prior(): top and bot priors collapse when the response has many zeros
## Summary

`define_prior()` builds the `top` and `bot` priors from quantiles of the raw
response. When a large share of the response is exactly zero those quantiles are
zero, and the resulting gamma priors collapse onto the fudge term — pinning
`bot` at essentially zero, and in the extreme producing the literal prior string
`gamma(2, Inf)`.

## Reproduction

```r
devtools::load_all(".")
set.seed(1)
x  <- as.numeric(rep(1:10, each = 15))
mu <- 40 * exp(-0.35 * pmax(x - 2, 0)) + 5          # true bot = 5
y  <- as.numeric(rpois(length(x), mu) * rbinom(length(x), 1, 0.7))   # zi = 0.30

mean(y == 0)                                         # 0.327
define_prior("nec4param", validate_family("zero_inflated_poisson"), x, y)
#>                        prior nlpar lb ub
#>                 normal(0, 5)  beta
#>  gamma(2, 0.0833333333333333)  top  0
#>    gamma(2, 66.6666666666667)  bot  0     <- prior mean 0.03, true bot = 5
#>   gamma(5, 0.363636363636364)  nec  1 10
```

Three distinct failures:

**1. `bot` pinned at zero once `zi >= 0.25`.** `R/define_prior.R:64`

```r
u_b_g <- paste0("gamma(2, ",
                1 / ((quantile(response, probs = 0.25) +
                  min(response[response > 0]) / 100) / 2), ")")
```

With a quarter or more of the response zero, `quantile(response, 0.25)` is
exactly `0` and the rate is determined entirely by the
`min(positive) / 100` fudge term. Above, the true lower asymptote is 5 and the
prior mean is 0.03.

**2. `prior_type = "regularizing"` is worse, and unconditional.**
`R/define_prior.R:92` uses `quantile(response, probs = 0)` — the minimum — which
is `0` for *any* response containing a zero, whatever the proportion. `bot` is
pinned near 0.1 regardless of `zi`.

**3. `gamma(2, Inf)` past 75% zeros.** `R/define_prior.R:61` divides by
`quantile(response, 0.75) / 2`, which is `1/0` when three quarters of the
response is zero:

```r
y3 <- as.numeric(rpois(length(x), mu) * rbinom(length(x), 1, 0.2))
mean(y3 == 0)                                        # 0.773
define_prior("nec4param", validate_family("zero_inflated_poisson"), x, y3)
#>  gamma(2, Inf)   top
```

## Scope

This affects every model carrying `bot` that survives `check_models()` for a
count or Gamma family — `nec4param`, `ecx4param`, `nechorme4`, `ecxll4`,
`ecxll5`, the `ecxwb*` set, `ecxhormebc5`.

It is **not** new in #104: a plain `poisson` or `Gamma` response containing many
zeros has always hit this. But #104 makes it the normal case rather than an
accident, because the zero-inflated count families exist precisely for the
regime where a quarter or more of the response is zero.

The inline comment added at `R/define_prior.R:33-39` was corrected in the same
PR — it originally said the effect "places `top` a little low" and that the
priors were "weakly informative either way", which is wrong on both the
parameter and the magnitude.

## Suggested fix

Guard against a zero (or non-finite) quantile rather than reworking the tables.
Something along the lines of: where the quantile used to set a gamma rate is
zero, fall back to a scale derived from the positive part of the response — the
smallest positive value, or the corresponding quantile of `response[response > 0]`
— so the prior stays weakly informative instead of collapsing.

`define_hurdle_prior()` already sidesteps this by computing the mu-block priors
from the non-zero subset, which is exact for `hurdle_gamma` because the Gamma
has no mass at zero. That trick is **not** directly transferable to the
zero-inflated count families: conditioning on the non-zeros biases `mu` upward
there (the same truncation argument as in the companion `hurdle_poisson` issue),
so the guard is the safer route for the single-block path.

Whatever is chosen needs a test asserting the priors stay finite and sensibly
centred at, say, 30%, 50% and 80% zeros.

beckyfisher:
**Release tier: 2.1.4.** Item 3 of the stack. **Stacked on #221** (`#139`), which
is stacked on #220 (`#215`). GitHub retargets this to `dev` as those merge.

Closes #210.

## What changed

The `top` and `bot` gamma priors set their rate from a quantile of the response.
Those quantiles are zero once enough of the response is zero, and the rate then
either collapses onto the fudge term or divides by zero. New internal
`positive_scale()` falls back to the same quantile of the **positive part** of
the response wherever the raw quantile has collapsed.

All three failures in the issue, verified on the issue's own reprex:

| | before | after |
|---|---|---|
| `bot` rate at 33% zeros | 66.7 — prior mean **0.03**, true `bot` = 5 | 0.199 — prior mean 10.0 |
| `bot`, `prior_type = "regularizing"` | pinned near 0.1 regardless of the zero fraction | 1.52 — prior mean 3.3 |
| `top` at 77% zeros | **`gamma(2, Inf)`** | `gamma(2, 0.079)` |

A response with **no** zeros is untouched: rate 0.19960 against 0.19940 before,
and `positive_scale()` provably returns the raw quantile when every value is
positive (asserted).

## Why not `define_hurdle_prior()`'s approach

The obvious fix is the one `define_hurdle_prior()` already uses — compute the
prior from the non-zero subset. That is **exact** for `hurdle_gamma`, because a
Gamma has no mass at zero, so the non-zero subset *is* the mu process.

Under zero-inflation it is not. The base distribution emits zeros of its own, so
conditioning on the positives draws from a *truncated* count distribution and
biases the location upward — the same argument as the growth-component
likelihood in #209. So the positive part is used here only to recover a **scale**
when the raw quantile has collapsed, never as the estimate itself. That
distinction is documented on `positive_scale()`.

## New behaviour: an all-zero response errors

Previously this produced `gamma(2, Inf)`, which `brms` would reject much further
downstream with a far less useful message. It now raises:

> Cannot build priors for "top" and "bot": the response contains no positive
> values, so there is no scale to place them on.

## Out of scope, and done anyway — please read

The queue entry scoped this as "`R/define_prior.R` only". Two changes fall
outside that; both are flagged rather than folded in silently.

**1. `R/helpers.R` — `response_link_scale()` computed `min_z_val` eagerly.**
Surfaced by the all-zero test above: `min(response[response > 0])` on an
all-zero response is `Inf` and emits `no non-missing arguments to min`. The
value is never *used* for an identity-link count family — the branches that
consume it need a logit/log link or a bounded family — so an entirely legitimate
input warned about nothing. Made it lazy; the value is unchanged everywhere it
is actually used. Three lines, same defect class.

**2. `tests/testthat/test-failed_models.R` — de-flaked, in the #139 commit
below.** #221's CI went red on a test unrelated to either issue. It forces a
failure with `timeout = 1e-3` then asserts the recorded message matches
`"time limit"`; CI instead recorded *"Expecting a single value when fixing
parameter 'ec50'"* from the initial-value search. Where a 1 ms interrupt lands
is a race, so the message is not a stable thing to assert.

The arithmetic confirms it was not caused by #139: PR #220 ran 1449 pass / 0 fail
on the identical runner, #221 ran 1461 / 1, and 1449 + 13 new = 1462 = 1461 + 1.
Every new assertion passed; one pre-existing test failed.

Fixed at the base of the stack because otherwise **every PR above it inherits
the flake**. The test now asserts what it is named for — that the failure was
captured and reported with its priors and inits — rather than the incidental
wording.

## Tests

Six new blocks in `tests/testthat/test-define_prior.R` (41 → 75 assertions):
priors finite and positive at 30/50/80% zeros across both `prior_type` values;
`bot` no longer pinned once a quarter of the response is zero; the
`"regularizing"` path exercised with a **single** zero, since that path used the
minimum and so collapsed for any zero at all; a no-zeros response provably
unchanged; the all-zero error; and `response_link_scale()` silent on an all-zero
response.

## Verified

|---|---|---|---|---|
| #139 branch below this one | 0 | 11 | 0 | 1470 |
| this branch | 0 | 11 | 0 | **1504** |

+34, exactly the new assertions; warnings unchanged.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: bayesnec
 Title: A Bayesian No-Effect- Concentration (NEC) Algorithm
-Version: 2.1.3.9
+Version: 2.1.3.10
 Authors@R: c(person("Rebecca", "Fisher", email = "r.fisher@aims.gov.au", role = c("aut", "cre"),comment = c(ORCID = "0000-0001-5148-6731")), person("Diego R.","Barneche",role="aut",comment = c(ORCID = "0000-0002-4568-2362")), person("Gerard F.","Ricardo",role="aut",comment = c(ORCID = "0000-0002-2220-3971")), person("David R.","Fox",role="aut",comment = c(ORCID = "0000-0002-3178-7243")))
 Description: Implementation of No-Effect-Concentration estimation that uses 'brms' (see Burkner (2017)<doi:10.18637/jss.v080.i01>; Burkner (2018)<doi:10.32614/RJ-2018-017>; Carpenter 'et al.' (2017)<doi:10.18637/jss.v076.i01> to fit concentration(dose)-response data using Bayesian methods for the purpose of estimating 'ECx' values, but more particularly 'NEC' (see Fox (2010)<doi:10.1016/j.ecoenv.2009.09.012>), 'NSEC' (see Fisher and Fox (2023)<doi:10.1002/etc.5610>), and 'N(S)EC (see Fisher et al. 2023<doi:10.1002/ieam.4809>). A full description of this package can be found in Fisher 'et al.' (2024)<doi:10.18637/jss.v110.i05>. This package expands and supersedes an original version implemented in 'R2jags' (see Su and Yajima (2020)<https://CRAN.R-project.org/package=R2jags>; Fisher et al. (2020)<doi:10.5281/ZENODO.3966864>).
 Depends:
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,20 @@
 # bayesnec 2.1.4
 
+- `define_prior()` no longer collapses the `top` and `bot` priors when a large
+  share of the response is exactly zero. The gamma rates for those parameters
+  are set from quantiles of the response, and those quantiles are zero once
+  enough of the response is — so `bot` was pinned near zero whatever the real
+  lower asymptote was (from 25% zeros under the default `prior_type`, and from a
+  *single* zero under `"regularizing"`, which uses the minimum), and past 75%
+  zeros the `top` rate divided by zero and produced the literal prior string
+  `gamma(2, Inf)`. Where a quantile has collapsed, the scale now falls back to
+  the same quantile of the positive part of the response. This is not new to the
+  zero-inflated count families added in 2.1.4 — a `poisson` or `Gamma` response
+  with many zeros always hit it — but those families exist precisely for that
+  regime, which made it the normal case rather than an accident. A response with
+  no positive values at all now raises an informative error instead. Priors for
+  a response containing no zeros are unchanged. See #210.
+
 - `?models` and `vignette("example2b")` now document how the `drc` package's
   `NEC.2()`, `NEC.3()` and `NEC.4()` map onto the `bayesnec` model set.
   `NEC.4()` and `NEC.3()` are `nec4param` and `nec3param` — not approximations
--- a/R/define_prior.R
+++ b/R/define_prior.R
@@ -1,3 +1,57 @@
+#' Quantile of a response, guarded against exact zeros
+#'
+#' The gamma priors for "top" and "bot" set their rate from a quantile of the
+#' response. Where a large share of the response is exactly zero those
+#' quantiles are zero, and the rate either collapses onto a fudge term or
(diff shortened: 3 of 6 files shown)


beckyfisher (commented):
This all looks good.

PR #223, opened 2026-08-21
Accept a dispersion prior, and fill an incomplete prior set (#207)

### LINKED ISSUE #207: Accept a prior on the dispersion parameter through bnec(prior = ), and notice when a supplied prior is incomplete
## Summary

`bnec()` cannot be given a prior on the family's own dispersion parameter — `sigma`, `shape`, `phi` — and it cannot tell the user when a supplied prior is missing parameters the model needs. Two separate defects in the same place, both found while reviewing #202 (`get_priors()`, #141).

## 1. A dispersion prior is rejected

The dispersion parameter is left entirely to the `brms` default, and there is no route to change that. Supplying one fails in the initial-value search rather than being passed through:

```r
pr <- data.frame(prior = c("normal(1,1)", "normal(0,5)", "normal(0.5,1)",
                           "gamma(5,2)", "student_t(3,0,2.5)"),
                 class = c(rep("b", 4), "sigma"), coef = "", group = "",
                 resp = "", dpar = "", nlpar = c("top", "beta", "bot", "nec", ""),
                 lb = "", ub = "", stringsAsFactors = FALSE)
bayesnec:::make_inits("nec4param", c("b_top", "b_beta", "b_bot", "b_nec"),
                      priors = pr, chains = 2)
#> Error: In model nec4param, user-specific parameter prior names
#> (top, beta, bot, nec, sigma) do not match expectation (top, beta, bot, nec).
```

`make_inits()` (`R/inits_functions.R`) tests `identical(sort(par_names), sort(fct_args))` — exact set equality between the prior's parameter names and the *curve's* arguments. Any row naming something the mean curve does not have kills the call, whether or not an initial value is wanted for it.

**Consequences.** A user cannot regularise the dispersion parameter where the data are sparse and the default is too vague. `get_priors()` (#202) cannot report it either, so a fit's prior set is a complete record of everything *except* dispersion.

**Proposed fix**, which looks small:

- filter the supplied prior down to the model's own `class == "b"` rows *before* the name check, and check only those. `add_brm_defaults()` (`R/helpers.R`) already does exactly this for the parameters a `disp()` variance function adds, using `disp_pars()`, so the pattern is established;
- pass the whole prior — dispersion row included — to `brm()`;
- supply no initial value for it. Stan random-initialises any parameter absent from an init list, and `bayesnec` already never gives `sigma` an init, so nothing downstream needs teaching;
- drop the `class == "b"` filter in `usable_prior()` (`R/get_priors.R`) so the row round trips.

**Not in scope.** *Generating* a default dispersion prior is a separate decision: it would change every existing fit, and the current behaviour (a `brms` default) is at least well understood. This issue is only about accepting one the user supplies.

## 2. A prior missing the parameters the formula introduces is accepted silently

`validate_priors()` accepts any `brmsprior` wholesale, and `bnec()` then uses it as the complete set. Where a route B `disp()` term is in the formula, a prior set without the variance function's parameters therefore fits without complaint, on flat priors:

```r
base <- get_priors(y ~ crf(x, "nec4param") + disp("power"), data = nec_data,
                   family = gaussian())
partial <- base[!base$nlpar %in% c("c0", "c1"), ]     # user edits, drops rows
fit <- bnec(y ~ crf(x, "nec4param") + disp("power"), data = nec_data,
            family = gaussian(), prior = partial, iter = 400, warmup = 200,
            chains = 2)
pull_prior(fit)[[1]]
#>   prior nlpar  source
#>  (flat)    c0 default
#>  (flat)    c1 default
```

This is the state `define_disp_prior()` was written to prevent — its own documentation notes that `c0` and the slope are "near-perfectly confounded" and the induced prior on the dispersion parameter spans orders of magnitude without the centring prior. Nothing warns.

Editing a returned prior set and handing it back is exactly the workflow `get_priors()` invites, so this becomes easier to hit as of #202.

**Proposed fix.** Have `validate_priors()` compare the supplied prior against the model's full parameter set — curve parameters plus anything the formula introduces — and say what is missing. Whether that is an error or a warning-plus-fill-from-defaults is the open question; an error is more consistent with the existing "priors for **all** parameters" requirement.

## Verified on

`issue-141-get-priors` at 54986ed, R 4.6.1, brms 2.23.0.

beckyfisher:
**Release tier: 2.1.4** — the last item of that tier. Item 4 of the stack,
**stacked on #222** (`#210`) → #221 (`#139`) → #220 (`#215`).

Closes #207. Both defects in the issue, plus a correction to one of its
premises.

## Part 1 — a supplied dispersion prior is no longer rejected

`make_inits()` tested `identical(sort(par_names), sort(fct_args))` — exact set
equality between the prior's parameter names and the **curve's** arguments — so
any row naming `sigma`, `shape` or `phi` killed the call. There was no route to
regularise dispersion where the data are sparse and the `brms` default is too
vague, and `get_priors()` could not report it either, so a fit's prior set was a
complete record of everything except dispersion.

The prior is now filtered to the curve's own `class == "b"` rows before the
check. That is the same move `add_brm_defaults()` already makes for the
parameters a `disp()` variance function adds, so the pattern was established.
The prior still reaches `brm()` in full; only the initial-value search ignores
the dispersion row — which is correct, since Stan random-initialises any
parameter absent from an init list and `bayesnec` has never given `sigma` one.

Verified on the issue's exact reprex, and across all three dispersion classes.

## Part 2 — and a correction to the issue's premise

The issue was read as saying `disp()` loses `bayesnec`'s automatic prior
building. **It does not.** On `dev` at 9ef03b85:

```r
get_priors(y ~ crf(x, "nec4param") + disp("power"), data = nec_data,
           family = gaussian())
#>              normal(-1.515, 2)     b    c0
#>                   normal(0, 2)     b    c1
```

`define_disp_prior()` works, and every parameter the term introduces gets a
`bayesnec` prior. The priors are lost **only when the user supplies an
incomplete set**, which `validate_priors()` then accepted wholesale — so the
parameters they did not mention fell through to `brms`, which means a **flat**
prior, silently.

That is the opposite of what the package is for. Per RF's decision on the issue
— *bayesnec is proactive about weakly informative priors because flat priors are
rarely useful in non-linear modelling* — the gaps are now filled from
`bayesnec`'s own defaults, with a warning naming each one:

> The prior supplied for model nec4param has no entry for **c0, c1**. Using the
> bayesnec default for those parameters rather than leaving them on a flat
> prior; see ?get_priors to inspect the full set.

**Filling rather than erroring is the deliberate choice.** An error refuses the
user's partial set and produces no fit at all; filling gives them the model they
asked for with the priors `bayesnec` would have chosen anyway, and the warning
means it is never silent.

Verified end to end on the issue's second reprex. Before, `pull_prior(fit)` gave
`(flat) c0 default` and `(flat) c1 default`. Now:

```
normal(-1.515, 2)   b   c0   user
normal(0, 2)        b   c1   user
```

## ⚠ This changes results

A fit that previously ran on flat priors for parameters its supplied prior did
not mention will now run on `bayesnec` priors and give different numbers. That
is the intended fix, but it is not silent and it gets its own `NEWS.md`
paragraph rather than being folded into the Part 1 entry.

## Also: `usable_prior()` keeps dispersion rows

So that a supplied dispersion prior round trips and `get_priors()` becomes a
complete record of the fit. The `source == "user"` test is retained, so a
dispersion prior **brms** chose is still dropped — reporting that would suggest
`bayesnec` had made a choice it did not make.

`dispersion_classes()` is `c("sigma", "shape", "phi")`, covering every family
`bayesnec` supports. `zi` and `hu` are deliberately absent: those are modelled
as a second parameter block, so a prior on them carries class `"b"` with a
prefixed `nlpar` rather than a class of its own.

## Tests

`test-inits_functions.R` — a dispersion prior accepted across all three classes;
no init generated for it; a prior naming a parameter the curve does not have
still errors, so the name check keeps doing its job.

`test-helpers.R` — `fill_missing_priors()` fills and names the gaps; a complete
prior is returned unchanged and **silently**; a user value **overrides** the
default rather than being duplicated (matching is on class + nlpar + dpar); a
`sigma` row with an empty `nlpar` does not collide with a curve parameter; empty
defaults are a no-op.

## Verified

|---|---|---|---|---|
| #210 branch below this one | 0 | 11 | 0 | 1504 |
| this branch | 0 | 11 | 0 | **1526** |

+22, exactly the new assertions; warnings unchanged.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: bayesnec
 Title: A Bayesian No-Effect- Concentration (NEC) Algorithm
-Version: 2.1.3.10
+Version: 2.1.3.11
 Authors@R: c(person("Rebecca", "Fisher", email = "r.fisher@aims.gov.au", role = c("aut", "cre"),comment = c(ORCID = "0000-0001-5148-6731")), person("Diego R.","Barneche",role="aut",comment = c(ORCID = "0000-0002-4568-2362")), person("Gerard F.","Ricardo",role="aut",comment = c(ORCID = "0000-0002-2220-3971")), person("David R.","Fox",role="aut",comment = c(ORCID = "0000-0002-3178-7243")))
 Description: Implementation of No-Effect-Concentration estimation that uses 'brms' (see Burkner (2017)<doi:10.18637/jss.v080.i01>; Burkner (2018)<doi:10.32614/RJ-2018-017>; Carpenter 'et al.' (2017)<doi:10.18637/jss.v076.i01> to fit concentration(dose)-response data using Bayesian methods for the purpose of estimating 'ECx' values, but more particularly 'NEC' (see Fox (2010)<doi:10.1016/j.ecoenv.2009.09.012>), 'NSEC' (see Fisher and Fox (2023)<doi:10.1002/etc.5610>), and 'N(S)EC (see Fisher et al. 2023<doi:10.1002/ieam.4809>). A full description of this package can be found in Fisher 'et al.' (2024)<doi:10.18637/jss.v110.i05>. This package expands and supersedes an original version implemented in 'R2jags' (see Su and Yajima (2020)<https://CRAN.R-project.org/package=R2jags>; Fisher et al. (2020)<doi:10.5281/ZENODO.3966864>).
 Depends:
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,29 @@
 # bayesnec 2.1.4
 
+- `bnec()` now accepts a prior on the family's own dispersion parameter —
+  `sigma`, `shape` or `phi`. Supplying one previously failed in the
+  initial-value search, because it compared the prior's parameter names against
+  the *curve's* arguments as an exact set, so any row naming something the mean
+  curve does not have killed the call. There is now no way to regularise
+  dispersion where the data are sparse and the `brms` default is too vague, and
+  `get_priors()` could not report it either, so a fit's prior set was a complete
+  record of everything except dispersion. No initial value is generated for it:
+  Stan random-initialises any parameter absent from an init list. See #207.
+
+- A prior set that is missing parameters the model needs is now filled from the
+  `bayesnec` defaults, with a warning naming each one, instead of being accepted
+  as though it were complete. Previously the unmentioned parameters fell through
+  to `brms`, which means a **flat** prior — the opposite of what `bayesnec` is
+  for, since it generates weakly informative priors precisely because flat ones
+  are rarely useful in non-linear modelling. The case this bit hardest is the
+  one `define_disp_prior()` exists to prevent: drop the `c0` and slope rows a
+  route B `disp()` term adds and the fit ran on flat priors for parameters its
+  own documentation calls "near-perfectly confounded", with nothing to say so.
+  Editing a returned prior set and handing it back is exactly the workflow
+  `get_priors()` invites, so this was easy to hit. **This changes results**: a
+  fit that previously ran on flat priors for the parameters it did not mention
+  will now run on `bayesnec` priors and give different numbers. See #207.
+
 - `define_prior()` no longer collapses the `top` and `bot` priors when a large
   share of the response is exactly zero. The gamma rates for those parameters
   are set from quantiles of the response, and those quantiles are zero once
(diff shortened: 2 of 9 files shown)


beckyfisher (commented):
This all looks good.

PR #224, opened 2026-08-21
Support the rate() aterm, and fix dispersion()'s link (#136, #247)

### LINKED ISSUE #63: Allow dispersion estimates across all families
While technically on the families Poisson and Binomial can be over dispersed in bayesnec, the calculation of dispersion is possible for any family, and simply reflects the ability of the predicted model to reproduce the appropriate level of variability around the predicted values relative to the observed data. I think it is worth allowing this calculation to be done for all the currently implemented families.

### LINKED ISSUE #136: Support for `rate` aterm would be really useful for Poisson and Negbinom families
See https://rdrr.io/cran/brms/man/brmsformula.html

Alternatively, if `offset` is easier that would work too.

I have looked at taking a stab at it myself but I would honestly be in way over my head

Thanks for a great package!

### LINKED ISSUE #247: dispersion() applies the family's default link inverse, not the identity link bnec() fits with
`dispersion()` rebuilds the family with `get(fam)()`, which uses that family's
**default link** — `log` for Poisson, `logit` for Binomial. It then applies
`fam_fcts$linkinv()` to `posterior_linpred()`.

But `validate_family()` forces `link = "identity"` on every family, so
`posterior_linpred()` is **already on the response scale**. The `linkinv()` call
transforms a value that has not been transformed.

`R/dispersion.R:56-58` (dev @ 6e0cff6f, and unchanged on `master`):

```r
fam_fcts  <- get(fam)()                      # stats::poisson() -> log link
...
prd_mu    <- fam_fcts$linkinv(lpd_out[i, ])  # exp() of a response-scale mean
prd_var_y <- fam_fcts$variance(prd_mu)
```

For Poisson this exponentiates a mean of, say, 90 — giving a variance weight of
`2.65e39`. The weights do **not** cancel out of the final ratio, because
`rowSums(prd_sr)` and `rowSums(sim_sr)` are separate sums over observations with
wildly different weights. The statistic ends up dominated by the few
lowest-mean observations, and understates dispersion.

## Reprex

```r
library(bayesnec)
library(brms)

# The mechanism, no fitting required:
stats::poisson()$link    # "log"    <- what dispersion() rebuilds
stats::binomial()$link   # "logit"
# ...but bnec() always fits with link = "identity"

set.seed(333)
n  <- 100
x  <- runif(n, 0, 3.2)
nec4 <- function(x, top, bot, nec, beta)
  bot + (top - bot) * exp(-exp(beta) * (x - nec) * (x > nec))
mu <- nec4(x, top = 90, bot = 6, nec = 1.5, beta = 0.3)
y  <- rnbinom(n, mu = mu, size = 12)      # genuinely overdispersed counts
d  <- data.frame(x = x, y = as.integer(y))

fit <- bnec(y ~ crf(x, model = "nec4param"), data = d, family = "poisson",
            seed = 333, open_progress = FALSE, refresh = 0)

m   <- fit$fit
m$family$link                             # "identity"
range(posterior_linpred(m))               # 12.1 to 90.8  -- response scale
range(exp(posterior_linpred(m)))          # 1.85e+05 to 2.65e+39  <- feeds variance()

# reconstruct the statistic both ways
obs <- standata(m)$Y
epr <- posterior_epred(m)
lpd <- posterior_linpred(m)
set.seed(700); sim <- posterior_predict(m)
num <- sweep(-epr, 2, -obs)^2
den <- (sim - epr)^2
q   <- function(v) round(quantile(v, c(0.5, 0.025, 0.975)), 3)

q(rowSums(num / exp(lpd)) / rowSums(den / exp(lpd)))  # as shipped
q(rowSums(num / epr)      / rowSums(den / epr))       # identity linkinv
fit$dispersion
```

Output (bayesnec 2.1.3.17, brms 2.23.0, R 4.5.2):

```
                     Estimate    Q2.5   Q97.5
as shipped              1.656   0.663   4.813
fit$dispersion          1.660   0.671   4.763   # matches, up to the PPD draw
with identity linkinv   6.231   4.724   8.346
```

The data are negative-binomial with `size = 12`, so the true variance-to-mean
ratio is `1 + mu/12` — 8.5 at the top of the curve, 1.5 at the bottom. The
corrected value sits in that range; the shipped value does not.

**A 3.8x understatement, and it biases toward "this model is fine".**

## Binomial: same defect, much smaller effect

`binomial()$linkinv` is `plogis()`, applied to an already-response-scale
proportion. On a beta-binomial-generated test set the linear predictor spans
0.150–0.941 and `plogis()` compresses it to 0.537–0.719, where `p(1-p)` is
nearly flat (0.248–0.202). Nearly-constant weights largely cancel:

```
shipped                 6.033   4.498   8.286
with identity linkinv   5.940   4.392   8.110
```

~2% here. Still wrong in principle, and the magnitude is data-dependent, but
this is not the urgent half.

## Impact

- **`master` / CRAN is affected.** The line dates to 6ac481b6 (`close #63`) and
  is unchanged on `dev`.
- **The `Example 1` vignette prints a distorted number.** `exp_3$dispersion`
  reports `3.5337626 0.4859079 57.6675764` under text asserting "the dispersion
  estimate is much greater than one, indicating serious overdispersion" — a
  credible interval spanning two orders of magnitude that *includes 1*. Fixing
  this will move every dispersion number in that vignette, so #63's tests and
  the precompiled vignette output both need regenerating.
- Anyone using `dispersion()` to choose between `poisson` and `negbinomial`, or
  between `binomial` and `beta_binomial`, may have been steered wrong.

## Suggested fix

Take the link from the fitted model rather than the family default:

```r
fam_fcts <- get(fam)(link = model$family$link)
```

`bnec()` always sets `identity`, so `get(fam)("identity")` would also work, but
reading it off the fit is robust to a future path that does not force it.

Related: this is the same trap documented in the JSS paper — a `bnec()` fit and
a hand-rolled `brms` fit differ unless the link is passed explicitly. Here the
package has fallen into it internally.

Found while scoping a replacement count dataset for the `Example 1` vignette
(the current `count_data` is a rounded Gaussian, not a count — separate issue to
follow) and while testing #136 / #224.

beckyfisher:
> [!WARNING]
> **Do not delete `issue-136-rate-aterm` when merging.** #225
> (`issue-209-hurdle-counts`) is based on this branch, and deleting it closes
> #225 — which cannot be reopened until the branch is restored. Untick "Delete
> branch" in the merge UI, and do not pass `--delete-branch` to `gh pr merge`.

**Release tier: 2.1.4.** This PR originally opened a `# bayesnec 2.2.0` heading,
but #240's review renamed that tier back to `2.1.4` (`9bc45d2e`); this branch
simply predated the decision. Version `2.1.3.20`, one above `dev`.

**No longer stacked.** The stack below has landed, `issue-148-part-d` is gone
from the remote, and this PR now targets `dev` directly. `dev` has been merged
in and the two `NEWS.md` / `DESCRIPTION` conflicts resolved.

Note this PR is **not purely additive**: unrecognised aterms become an error
(its own `**Breaking:**` NEWS bullet, per RF on #136), and `dispersion()` values
change for every `poisson` and `binomial` fit (#247, below).

Closes #136. Both of RF's calls on the issue are implemented: the zero-inflated
count families are refused with their own message, and unrecognised aterms are
now an error.

## What `rate()` does here

`rate(denominator)` declares a count observed over an exposure. Because
`bnec()` forces `link = "identity"`, `brms` writes the denominator
**multiplicatively on the response scale** rather than as a log offset:

```stan
target += poisson_lpmf(Y | mu .* denom);
```

so `mu` *is* the rate and `top`, `bot` and `nec` stay directly interpretable as
counts per unit exposure — exactly the property the identity link exists to
preserve. `offset` is deliberately not offered: under an identity link it would
be additive on the mean, which is not what anyone wants.

## It was half-supported, and failed late

Before this, `rate()` emitted the generic unvalidated-aterm **message**, then
`brm()` fitted fine, and `bnec()` errored in `posterior_epred()` at
`expand_classes.R:46` — because the prediction grid carried no denominator
column. Tens of seconds after the message had scrolled past.

There was a quieter defect underneath: `get_priors()` derived `top` from **raw
counts** rather than `y / ex`, silently.

| | before | after | truth |
|---|---|---|---|
| `top` prior mean | ~61 | **19.8** | 20 |
| the reprex fit | errored in post-processing | NEC 3.893 [3.634, 4.125] | 4 |
| `ecx()` | unreachable | 4.071 [3.823, 4.289] | — |

## Where the denominator had to go

- **`prediction_grid()`** pins it at 1. With a varying exposure the "fitted
  curve" would be the curve times whatever denominator each grid row inherited,
  and `ecx()`/`nsec()` read off it would be meaningless. This is also what makes
  the grid resolvable at all. *The issue lists `bnec_newdata.R:57` and
  `expand_classes.R:44` as two sites; they are one function feeding both.*
- **`check_data()`** carries it as `mod_dat$denom`.
- **`get_priors()`** and **`amend()`** divide the response through before
  building priors.
- **`plot()` / `autoplot()`** put the observations on the rate scale to match.
- **`wrangle_model_formula()`** restricts it to `poisson` / `negbinomial`. The
  check needs the family, so it cannot live in `check_formula()`.
- **`check_hurdle_aterms()`** refuses it: only one component is a count, and the
  survival component is one Bernoulli trial per individual with no rate to take
  a denominator.

## `dispersion()` — the comment's worry turned out to be moot

The scoping comment flagged this as needing real thought, because for
negative binomial `brms` scales the shape by the denominator too, so the
count-scale variance is `mu_c + mu_c^2 / (shape * denom)` rather than a plain
multiple.

**`dispersion()` does not accept negbinomial at all** — `allowed_fams` is
`c("poisson", "binomial")` — so there is nothing to get wrong today. The Poisson
case *is* the exact `mu * denom` analogue of the binomial branch and is
implemented; a comment warns whoever widens `allowed_fams` to derive the
negbinomial form rather than copy that line.

I checked `brms:::posterior_epred_poisson` rather than assuming: it calls
`multiply_dpar_rate_denom`, so it returns expected **counts**, which is what
makes the count-scale variance the right correction. Note the scaling does *not*
cancel in the ratio, because the denominator varies by row.

## ⚠ Breaking: unrecognised aterms are now an error

Per RF on the issue. The validated set is `trials()`, `weights()`, `cens()` and
`rate()`; anything else stops the call:

> You have specified brms special aterms bayesnec does not support: se(sei). The
> supported aterms are trials, weights, cens and rate.

The silent pass-through is precisely what let `rate()` fit and then fail in
post-processing. Its own `NEWS.md` bullet.

**Two existing tests relied on the old behaviour and are updated**, which is the
hazard the queue entry named:

- `test-make_brmsformula.R` carried `se(sei)` inside two multi-aterm chains. It
  was incidental — those tests exercise parsing a *chain*, which
  `trials + weights + cens` still does. Added a test that an unvalidated term
  rejects the whole chain rather than being quietly ignored.
- `test-cens.R` asserted the message; it now asserts the error.

**Vignettes and the JSS article were swept for formulas carrying a
non-validated aterm: none.** Nothing there breaks.

## Two bugs of mine the tests caught

- **`[["rate_var"]]` on a named vector errors when the name is absent** — unlike
  a list, which gives `NULL`. That broke *every* existing fit through
  `prediction_grid()`. The existing `trials_var` lookups are safe only because a
  family branch guards them. All four sites now use single-bracket lookup.
- An assertion compared an integer response against `as.numeric()`.

## Tests

New `tests/testthat/test-rate.R`, 19 assertions, modelled on `test-cens.R`:
parsing and the model frame; **`retrieve_var()` by position with censoring
present**, which is the trap the issue names — term order in `short_form` and
name order in `pop_vars` must stay in lockstep; priors on the rate scale; the
grid denominator pinned at 1; the family restriction; the zero-inflated message;
`poisson`/`negbinomial` accepted; the hurdle refusal; and the aterm error.

## `dispersion()` was reading the wrong link — closes #247

`dispersion()` rebuilt the family with `get(fam)()`, which carries that
family's **default** link — `log` for Poisson, `logit` for Binomial — and
applied its `linkinv()` to `posterior_linpred()`. Since `bnec()` forces
`link = "identity"`, the linear predictor is already on the response scale, so
the value was transformed a second time. For a Poisson that is `exp()` of a
mean of ~90, giving variance weights near `1e39`.

The weights do **not** cancel out of the final ratio, because `rowSums(prd_sr)`
and `rowSums(sim_sr)` are separate sums over observations carrying wildly
different weights. The statistic ended up dominated by the lowest-mean
observations and understated dispersion:

| | shipped | corrected |
|---|---|---|
| Poisson fit to Poisson data | 0.70 [0.09, 23.17] | **0.96 [0.73, 1.29]** |
| Poisson fit to overdispersed data | 1.96 [0.24, 66.88] | **23.06 [17.71, 30.53]** |

It could not separate clean Poisson data from data 23× overdispersed. The fix
reads the link off the fit rather than hard-coding `"identity"`, so it stays
correct if a future path stops forcing it.

This is **not** a defect this PR introduced — it dates to 6ac481b6 (`close #63`)
and is live on `master`. It is fixed here because rewriting the Example 1 count
sections (below) is the only place in the current PR stack where the numbers
surface, so fixing it elsewhere would mean precompiling that vignette twice.

`dispersion()`'s Poisson `rate()` branch was already correct — verified that
`posterior_linpred()` returns the rate and `posterior_epred()` the counts, their
ratio being exactly the denominator, so `variance(rate) * denom` is the right
count-scale variance. It was simply being handed an exponentiated input.

Binomial is the same defect with a much smaller effect: `plogis()` compresses an
already-response-scale proportion into a band where `p(1-p)` is nearly flat, so
the weights largely cancel (6.03 → 5.94 on a beta-binomial test set).

## Example 1's count data was not a count

```r
count_data <- nec_data |>
  dplyr::mutate(y = (y * 100 + rnorm(nrow(nec_data), 0, 15)) |>
                round() |> as.integer() |> abs())
```

`nec_data$y` is a proportion, so this is a rounded continuous measurement with
additive Gaussian noise of constant SD. Variance ~225 regardless of the mean, so
the variance-to-mean ratio **rises as the mean falls** — 1.9 at the top of the
curve, 11.2 at the bottom. Poisson needs `var = μ`; negative binomial needs
`var = μ + μ²/φ`, monotonically *increasing* in μ. Neither can represent it.

The vignette therefore taught the dispersion diagnostic on data where the
prescribed remedy is the wrong one, half-admitted it ("it still might not be the
best model for these data"), and left the reader with two failed fits and no
well-specified count fit anywhere.

| | now | truth |
|---|---|---|
| `count_data`, genuine Poisson — `nec` | 1.490 [1.413, 1.556] | 1.5 |
| `count_data`, dispersion | 0.96 [0.73, 1.29] | ~1 |
| `count_data_od`, `rnbinom(size = 20)` — Poisson dispersion | 4.90 [3.73, 6.59] | ≫1 |
| `count_data_od` — negbinomial `nec`, 0 divergences | 1.510 [1.382, 1.612] | 1.5 |

`size = 3` and `size = 10` were both tried first: `size = 3` gave 33 divergent
transitions and a `bot` of 24.8 against a truth of 5, which is not a fit to
present as the remedy.

## A `rate()` section in Example 1

`rate()` had no vignette presence. The new section models offspring counts where
the number of fecund females per replicate varies between 8 and 24 — the raw
counts go to `bnec` unchanged, so a replicate with twenty females contributes
more information than one with eight, which is exactly what fitting `y/n` as a
continuous response would discard.

| | estimate | truth |
|---|---|---|
| `nec` | 1.528 [1.454, 1.609] | 1.5 |
| `top` | 6.007 [5.863, 6.151] | 6 |
| `bot` | 0.673 [0.150, 1.264] | 0.4 |
| `dispersion` | 0.883 [0.664, 1.187] | ~1 |

Parameters print in offspring-per-female, which is the point. The section closes
with the caveat that matters on real data: if the exposure is itself an effect of
treatment — `n_females` being the number *surviving* to the point of counting —
the endpoint silently becomes fecundity conditional on survival, which is a
different *NEC*.

`?dispersion`'s roxygen example was building its "counts" the same way Example 1
did (`round(nec_data$y * 100)`), and is fixed too.

## Vignette regeneration

`vignettes/example1.Rmd` and its figures are **not** regenerated in this PR —
that needs `precompile/example1` (the #215/#230 mechanism), including two new
figures, `exmp1-poisson-od` and `exmp1-rate`. Because #247 is fixed in this same
PR it only needs running once; precompiling before the fix would have shipped a
vignette whose prose contradicted the numbers printed directly beneath it.

## Verified

New regression tests in `tests/testthat/test-dispersion.R`, which previously
held a single gaussian test. The strongest reconstructs the statistic
independently and asserts exact equality — using `seed = 10`, the documented
default, so the posterior-predictive draws match to `1e-8`. Then a near-one
check on Poisson-simulated data, and an overdispersion-detected check on
negative-binomial data.

Full suite on the merged tree:

|---|---|---|---|---|
| after merging `dev` | **0** | 2 | 0 | **1717** |

`test-dispersion.R` goes from 2 assertions to 7. The new `rate()` vignette
chunks were also executed verbatim outside `knit()` to confirm they run clean,
since #242 fails precompile on a chunk that errored.


beckyfisher:

Reviewed independently of authoring. **Verdict: sound; one minor finding, no blocking issues.** Both hazards named in the `01_work_queue.md` entry for #136 are handled correctly, and I checked them empirically rather than by reading.

### Hazard 1 — `retrieve_var()` positional indexing: **verified correct**

The queue's warning was that `retrieve_var()` finds a variable by its **position** in `bnec_pop` and then indexes the model frame with it, so the term order in `short_form` and the name order in `pop_vars` must stay in lockstep. This PR inserts the rate slot between censoring and the grouping terms in `short_form`, and appends `rate_var` last in `pop_vars` — which is consistent, because group variables are not in `pop_vars` at all.

Checked across six formula shapes on the branch:

```
shape          mf cols                      bnec_pop names               y_var  x_var  rate_var
plain          y,x                          y_var,x_var                  ok     ok     absent
rate           y,x,ex                       y_var,x_var,rate_var         ok     ok     ok
trials         y,x,trials(tr)               y_var,x_var,trials_var       ok     ok     absent
rate+group     y,x,ex,grp                   y_var,x_var,rate_var         ok     ok     ok
plain+group    y,x,grp                      y_var,x_var                  ok     ok     absent
trials+cens    y,x,trials(tr),cen           y_var,x_var,trials_var,...   ok     ok     absent
```

`rate+group` is the case that would break if the ordering were wrong — the rate column sits immediately before the group column — and it resolves correctly. Worth noting for anyone reading this later: `bnec_pop` **drops** absent slots rather than holding an `NA` placeholder, which is what makes the positional scheme safe.

### Hazard 2 — `dispersion()` is not a copy-paste of the binomial branch: **handled, and documented**

The queue warned that the Poisson analogue is exact but the negative binomial one is not, because `brms` scales the shape by the denominator too. The branch added here is Poisson-only, and its comment states the negbinomial variance explicitly and tells whoever widens `allowed_fams` to derive it rather than copy the line. That is the right outcome.

Walking the arithmetic: `prd_mu` comes from `linkinv(posterior_linpred())`, which under the forced identity link is the **rate**; `variance(prd_mu)` is therefore `mu`; multiplying by the denominator gives `Var(count) = mu * denom`; and `prd_y` from `posterior_epred()` and `obs_y` from `standata()$Y` are both on the count scale. The Pearson residual is consistent. This rests on `posterior_epred()` returning the count-scale mean for a rate model, which parallels the binomial branch exactly and which I have not verified by fitting.

### The breaking change is handled properly

Its own `NEWS.md` bullet, marked **Breaking**, rather than buried in the `rate()` entry — as the queue asked. I scanned the vignettes and test suite for aterms that would now error: the only occurrences are `se()` in `test-cens.R`, `test-bnec_hurdle.R` and `test-rate.R`, and all three are deliberate assertions of the new error. **No vignette uses an unvalidated aterm.**

### Tests

`test-rate.R` 19, `test-cens.R` 36, `test-make_brmsformula.R` 12 — all passing on the branch.

### Finding — minor, documentation precision

`make_brmsformula()` is **exported** with `family = NULL` as its default, and the rate/family validation in `wrangle_model_formula()` is guarded by `!is.null(family)`. So the documented call

```r
make_brmsformula(y | rate(ex) ~ crf(x, "nec3param"), data)
```

skips the family check entirely and builds a `brmsformula` carrying a rate term for an unknown family.

This is **not** a fitting-path bug — `bnec()` always supplies a family via `fit_bayesnec()`, so the guard covers every path that actually fits. But `?bayesnecformula` now states flatly that for `rate()` "Any other family is an error, the zero-inflated count families included", and via this entry point it is not. Suggest either a clause noting that the check requires a family to check against, or moving the family-independent half of the message earlier. Low priority; a doc precision issue rather than a defect.

### Not raised as findings, but worth recording

- The comment explaining why `[` is used rather than `[[` on `bnec_pop` (subscript error vs `NULL` on a missing name) appears at the `expand_classes.R` site and is correct; the same idiom in `plot.R`, `autoplot.R` and `dispersion.R` relies on it without repeating the explanation, which is the right balance.
- Pinning the denominator at 1 on the prediction grid, and dividing the observations through in both plot paths so the curve and the points share an axis, is the part most likely to be got wrong in a reimplementation. Both are commented with the reasoning rather than the mechanics.

**Recommend merge** once the stack below it is reviewed, subject to the documentation nit above, which can be folded into a later commit rather than blocking.

beckyfisher:
Addressed the review finding on this PR.

`?bayesnecformula` said flatly that for `rate()`, *"Any other family is an error, the zero-inflated count families included."* That is true on every path that fits — `bnec()` always supplies a family — but not via a bare `make_brmsformula(formula, data)`, where there is no family to check against and the guard is skipped.

The sentence now says so explicitly rather than overstating, and `man/bayesnecformula.Rd` is regenerated.

No code change: the guard itself is in the right place. `wrangle_model_formula()` is where the family is available, and moving the check earlier would mean duplicating family resolution in `check_formula()`, which does not have it. The alternative — making `make_brmsformula()` refuse a `rate()` formula when no family is given — would break the documented example that builds a formula without one.

Everything else in the review stands: both named hazards verified, breaking change correctly flagged, tests green.

beckyfisher:

```
```

beckyfisher:
⚠️ **Do not delete `issue-136-rate-aterm` on merge** — #225 (`issue-209-hurdle-counts`) is based on it and would be closed with it, and cannot be reopened until the branch is restored. Untick "Delete branch" in the merge UI; do not pass `--delete-branch`.

PR #225, opened 2026-08-21
Add joint and factorised count hurdles (#209)

### LINKED ISSUE #209: Add hurdle_poisson and hurdle_negbinomial as two-block families

### LINKED ISSUE #249: Restore the factorised route for count hurdles via internal trunc(lb = 1) (blocked on brms)
Deferred out of #225 / #209. **Blocked on brms — do not start until the gate below is met.**

## What is missing

`bnec_hurdle()` refuses `poisson` and `negbinomial` growth families (added in
#225, and correctly so: it would fit the positive part *untruncated* to
`data[y > 0, ]`, estimating `mu / (1 - exp(-mu))` rather than `mu`, with the
bias growing exactly at the high-concentration end the *NEC* is read off).

The consequence is that `hurdle_poisson` and `hurdle_negbinomial` are strictly
less capable than `hurdle_gamma`:

|                              | `bnec_hurdle()` (factorised) | `bnec(family = "hurdle_*")` (joint) |
|------------------------------|------------------------------|-------------------------------------|
| model set, growth block      | full set                     | full set                            |
| model set, survival block    | full set                     | one fixed equation                  |
| crossed comparison           | `crossed_weights()`, `best_crossed()` | not available              |
| shared group-level effects   | no                           | yes                                 |

The documented workflow — select with `bnec_hurdle()` + `crossed_weights()`,
then refit the winner with `bnec_joint()` — is unreachable for counts, because
all three take a `bayesnechurdlefit`.

## The fix, once unblocked

Have `bnec_hurdle()` add `trunc(lb = 1)` to the growth formula internally. It
already knows it has subset to `data[y > 0, ]`, so the truncation is a fact
about the subset, not a user modelling choice, and nothing needs to be exposed
publicly.

Scope is small — unlike `rate()` in #136, `trunc(lb = 1)` carries no variable,
so there is no `ra_call`/`ra_var` slot, no `short_form` position, no `pop_vars`
name and no lockstep with `retrieve_var()`. Prototyped on a worktree with
**only** the aterm gate at `R/bayesnecformula.R:507` widened to `...|trunc\\(`:

```
formula parsed OK
model.frame OK: TRUE
  cols: y, x                      # trunc dropped; no data column needed
  bnec_pop: y_var=y x_var=x       # slots untouched
wrangle_model_formula OK: TRUE
  y | trunc(lb = 1) ~ top * exp(-exp(beta) * (x - nec) * step(x - nec))
get_priors OK: TRUE
```

The remaining work is a small `add_aterm()` helper, since the growth call
passes `formula` through unchanged and `swap_response()` strips aterms.

The likelihood would be exact, not an approximation:

```
trunc(lb=1) + poisson :  poisson_lpmf(Y[n] | mu[n]) - poisson_lccdf(lb[n] - 1 | mu[n])
hurdle_poisson        :  poisson_lpmf(y | lambda)   - log1m_exp(-lambda)
```

`poisson_lccdf(0 | lambda)` **is** `log1m_exp(-lambda)`. Bernoulli plus
zero-truncated-positive sums exactly to the joint hurdle lpmf, so the elpd
additivity `crossed_weights()` relies on is exact.

## Why it is blocked: brms treats `trunc(lb = )` inconsistently on counts

The generated Stan uses `lb - 1`, i.e. **inclusive** of `lb`, so the model that
is *fitted* is right. `posterior_epred_trunc_discrete()` sums over
`(min_lb + 1):ub` and divides by `cdf(ub) - cdf(lb)` — **exclusive**, so it
describes `Y >= lb + 1`. It fails silently, with no warning.

Verified on a real fit, brms 2.23.0, `y | trunc(lb = 1) ~ x`, poisson:

| quantity | brms 2.23.0 | inclusive (what Stan fitted) |
|---|---|---|
| `posterior_epred[,1]` | 2.5139 (= E[Y \| Y>=2]) | 1.7546 (= E[Y \| Y>=1]) |
| `min(posterior_predict[,1])` | 1 | 1 (agrees) |

So `ecx()`, `nsec()`, `fitted()`, `conditional_effects()` and every plot would
be silently wrong.

`trunc(lb = 0)` is not a workaround: Stan then fits *untruncated* (correct
under the inclusive convention — coefficients match a plain fit to 3 s.f.)
while the R side zero-truncates. Different model, no warning.

Secondary: `posterior_epred_trunc_poisson()` / `_negbinomial()` cap `ub` at
`3 * max(mu)`, so the result is an approximation even where the bound is right.

## The gate

Both must hold before this is picked up:

1. **`log_lik`** — paul-buerkner/brms#1903, fixed by paul-buerkner/brms#1922
   (`lb <- ceiling(lb) - 1` under `use_int()`), merged 2026-08-24.
   ✅ on master, ❌ **not yet in a CRAN release** (2.23.0 is current).
2. **`posterior_epred`** — paul-buerkner/brms#1923, filed 2026-08-21,
   **still open**. #1922 touched `R/log_lik.R` only;
   `posterior_epred_trunc_discrete()` on master is byte-identical to 2.23.0.

Both were reported upstream by @deeb01, not by us — **do not file a duplicate.**
Watch paul-buerkner/brms#1923.

Raising our minimum brms is not a route out: there is no version, released or
unreleased, where this works end to end.

## One design question to settle before any of it is built

Even with correct brms the two routes would not agree for counts.
`posterior_epred` on a truncated growth fit gives **E[Y | Y > 0]**; the joint
`hurdle_poisson` `mu` block gives **lambda**. Those give different ECx for the
same data. `hurdle_gamma` has no such split, because a Gamma's epred *is* `mu`.

E[Y | Y > 0] is arguably better — it makes the combined endpoint exactly
`E[Y|Y>0] * P(Y>0) = E[Y]` — but if we take it, `ecx(dpar = "mu")` on the joint
fit must be defined to match, or the two routes silently disagree.

beckyfisher:
## What

Adds `hurdle_poisson` and `hurdle_negbinomial` as joint two-block families and
restores the factorised count-hurdle workflow in `bnec_hurdle()`.

The factorised growth block is now fitted with an internal `trunc(lb = 1)`.
That route requires brms 2.23.2 or later, which contains the inclusive-bound
fixes from paul-buerkner/brms#1903 and paul-buerkner/brms#1923. Earlier brms
versions stop before fitting and direct the caller to the joint family, whose
likelihood is already correct.

Closes #209.
Closes #249.

## Rationale

For structural count zeros, the positive block is conditional on `Y > 0`.
Fitting an ordinary Poisson or negative-binomial likelihood to that subset is
biased where the underlying mean approaches zero. brms 2.23.2 makes
`log_lik()` agree with the inclusive bound used by Stan. `bayesnec` then uses
closed-form conditional-positive means instead of brms's finite-grid expected-
prediction approximation, so its factorised outputs remain exact for heavy
negative-binomial tails.

The joint and factorised routes now report the same growth quantity:
`E[Y | Y > 0]`. For joint count hurdles, `ecx(dpar = "mu")` and
`nsec(dpar = "mu")` convert brms's underlying count mean to that conditional
positive mean. Stored `top` and `bot` parameters remain on the underlying mean
scale.

## Evidence

- Focused count-hurdle, hurdle-family, formula and zero-inflated regression
  tests pass.
- The mean-support completeness suite covers both new registry entries.
- A real factorised Poisson fit succeeds against an isolated brms 2.23.2 build;
  its growth data contain only positive counts, its Stan data contain the
  inclusive lower bound, and its corrected expected predictions have zero
  numerical difference from the analytical conditional mean.
- `R CMD check --no-manual --no-build-vignettes` completed at the final reviewed
  head with no errors. Its two warnings and two notes are the expected omitted
  vignette outputs, source-check `.git` directory and existing `modifyList`
  import note. GitHub Actions provides the platform matrix.

<details>

- Registers both joint families in `mod_fams`, `hurdle_fams` and
  `hurdle_mu_fams`, and regenerates `R/sysdata.rda`.
- Adds count-family routing to `hurdle_mu_family()`.
- Appends `trunc(lb = 1)` only inside `bnec_hurdle()`. User-supplied `trunc()`
  remains unsupported. `cens()` remains available for continuous growth, but
  is refused for count growth because brms does not condition its censored
  count likelihood on the truncation bound.
- Computes factorised Poisson and negative-binomial expected responses in
  closed form throughout fitting summaries, stored predictions,
  `posterior_epred()`, `fitted()`, diagnostics, ECx and NSEC. Negative-binomial
  mean and shape draws remain paired, and Poisson dispersion uses the exact
  conditional-positive variance.
- Keeps relative ECx and ECNSEC asymptotes on the same conditional-positive
  scale as predictions, including model-averaged draws and brms-specific mean
  links.
- Keeps zero-inflated Poisson and negative-binomial models on their existing
  mixture path; they do not factorise because an observed zero can arise from
  either component.
- Documents the family choice, brms version gate and growth-scale semantics in
  the help pages, NEWS and the hurdle vignette.
- Rebases the formerly parked branch onto current `dev` and resolves the stale
  conflict.

</details>

<details>
<summary>Verification details</summary>

Focused local results:

```text
test-hurdle_counts.R          PASS 64, SKIP 1
test-zero-inflated-counts.R   PASS 42, SKIP 1
test-bnec_hurdle.R            PASS 51, SKIP 1
test-hurdle_family.R          PASS 124, SKIP 1
test-mu_support.R             PASS 372
```

The isolated upstream integration fit used brms 2.23.2, fitted both blocks,
verified `standata(growth_fit)$lb` is present and compared bayesnec's expected
predictions with the analytical conditional-positive mean. The ordinary
installed brms 2.23.0 remains untouched and is covered by the early-error test.

</details>

### DIFF
--- a/R/bnec_hurdle.R
+++ b/R/bnec_hurdle.R
@@ -47,11 +47,14 @@
 #' two-block family (\code{hurdle_gamma}, \code{zero_inflated_beta}) is refused,
 #' because \code{bnec_hurdle} is itself the two-part model, and so are the
 #' zero-inflated count families, which are mixtures rather than hurdles -- see
-#' \code{\link{bnec}}. Note that a count family is accepted here but fitted
+#' \code{\link{bnec}}. The plain count families (\code{poisson},
+#' \code{negbinomial}) are refused too: \code{bnec_hurdle} would fit them
 #' \emph{untruncated} to the non-zero subset, whereas the positive part of a
 #' hurdle on counts is zero-truncated. That overestimates the mean where the
-#' mean is small, which is the upper end of the concentration range; a
-#' zero-truncated count family is not yet available.
+#' mean is small, which is the upper end of the concentration range. Use
+#' \code{bnec(family = "hurdle_poisson")} or
+#' \code{bnec(family = "hurdle_negbinomial")}, where \pkg{brms} writes the
+#' zero-truncated positive part itself.
 #' @param ... Further arguments passed to both \code{\link{bnec}} calls.
 #'
 #' @details
@@ -282,11 +285,22 @@ check_hurdle_growth_family <- function(family) {
   # is exact, because a Gamma has no mass at zero, which is why this was never
   # a problem before counts were in scope.
   #
-  # The separate-fits path cannot express the truncation without a trunc()
-  # aterm, and the joint families added in #209 do not need one: brms writes
-  # the zero-truncated positive part itself, via the - log1m_exp(-lambda)
-  # normaliser. So the fix is to send the user to the path that is correct by
-  # construction rather than to approximate it here. See #209.
+  # The obvious repair -- add trunc(lb = 1) to the growth formula, which is a
+  # fact about the y > 0 subset rather than a user modelling choice -- is
+  # blocked upstream, and that is the durable reason this stays refused rather
+  # than the scope of #209. brms handles trunc(lb =) inconsistently on discrete
+  # families: the generated Stan normalises at lb - 1 (inclusive, so the fit is
+  # right), but posterior_epred_trunc_discrete() sums over (min_lb + 1):ub and
+  # divides by cdf(ub) - cdf(lb) (exclusive), so ecx(), nsec(), fitted() and
+  # every plot would be silently wrong with no warning. log_lik had the same
+  # off-by-one; that half is fixed on brms master (paul-buerkner/brms#1922) but
+  # unreleased, and the epred half (paul-buerkner/brms#1923) is still open, so
+  # no brms version makes the route work end to end and pinning is not a way
+  # out. trunc(lb = 0) is not a workaround either -- Stan then fits untruncated
+  # while the R side zero-truncates. Tracked in #249; until it closes, the
+  # joint families are the only correct route, and brms writes their
+  # zero-truncated positive part itself via the - log1m_exp(-lambda)
+  # normaliser. See #209 and #249.
   if (fam_tag %in% c("poisson", "negbinomial")) {
     stop("bnec_hurdle cannot use ", fam_tag, " as the growth family. Its",
          " positive part would be fitted with an untruncated ", fam_tag,
@@ -296,7 +310,10 @@ check_hurdle_growth_family <- function(family) {
          " bnec(family = \"hurdle_", fam_tag, "\"), where brms writes the",
          " zero-truncated positive part itself. That fit also gives both",
          " blocks a concentration-response curve, which is what a count hurdle",
-         " is for. See ?bnec.",
+         " is for. Restoring the separate-fits route needs a zero-truncated",
+         " count likelihood, which is deferred on an upstream brms bug in the",
+         " post-processing of trunc() for discrete families",
+         " (open-AIMS/bayesnec#249). See ?bnec.",
          call. = FALSE)
   }
   if (is_hurdle_family(fam_tag)) {
--- a/man/bnec_hurdle.Rd
+++ b/man/bnec_hurdle.Rd
@@ -32,11 +32,14 @@ positive continuous response, \code{Beta} for one bounded on (0, 1). A
 two-block family (\code{hurdle_gamma}, \code{zero_inflated_beta}) is refused,
 because \code{bnec_hurdle} is itself the two-part model, and so are the
 zero-inflated count families, which are mixtures rather than hurdles -- see
-\code{\link{bnec}}. Note that a count family is accepted here but fitted
+\code{\link{bnec}}. The plain count families (\code{poisson},
+\code{negbinomial}) are refused too: \code{bnec_hurdle} would fit them
 \emph{untruncated} to the non-zero subset, whereas the positive part of a
 hurdle on counts is zero-truncated. That overestimates the mean where the
-mean is small, which is the upper end of the concentration range; a
-zero-truncated count family is not yet available.}
+mean is small, which is the upper end of the concentration range. Use
+\code{bnec(family = "hurdle_poisson")} or
+\code{bnec(family = "hurdle_negbinomial")}, where \pkg{brms} writes the
+zero-truncated positive part itself.}
 
 \item{...}{Further arguments passed to both \code{\link{bnec}} calls.}
 }


beckyfisher:

Reviewed independently of authoring. **Verdict: the design call here is the right one, and better than what the queue asked for. One real gap — a documentation contact point the queue specifically flagged, and which was missed.**

### The correctness fix: refusal rather than approximation

`01_work_queue.md` said of the untruncated growth fit: *"Fix it and say so — it is a wrong answer today, not a missing feature."*

This PR does not fix it by implementing a truncated count fit in the separate-fits path. It **refuses** that path for counts and directs the user to `bnec(family = "hurdle_poisson")`, where `brms` writes the zero-truncation itself via the `- log1m_exp(-lambda)` normaliser.

**That is the better answer**, and the reasoning in the code says why: the separate-fits path cannot express the truncation without a `trunc()` aterm, so implementing it there would be approximating something the joint family gets right by construction. Refusing and redirecting removes the wrong answer without adding a second way to get it nearly right.

The check is applied on the **auto-selected** family as well as a supplied one, which matters — an integer response picks `poisson` at `bnec_hurdle.R:196` without the user ever naming a family, and that is the path someone reaches silently. Good catch by the author; it would have been easy to guard only the explicit branch.

### The D4 hazard is handled

The queue warned: *"Do not reuse D4's reasoning in reverse... Say which of the two you are building and why."*

`?bnec` now draws the contrast explicitly — hurdle where the zeros are **observed** to be structural, single-block zero-inflated where they are **latent** and the zero-probability and mean are weakly separated exactly where the mean is small. That is the right distinction, stated in the right place, and it does not overclaim in either direction.

### Dispatch verified

The queue's claim that `ecx()` and `nsec()` pick the new families up with no code change, because `is_hurdle_family()` and `hurdle_dpar()` are table-driven, checks out on the branch:

```
hurdle_gamma             is_hurdle=TRUE  dpar=hu   mu_family=Gamma
zero_inflated_beta       is_hurdle=TRUE  dpar=zi   mu_family=beta
hurdle_poisson           is_hurdle=TRUE  dpar=hu   mu_family=poisson
hurdle_negbinomial       is_hurdle=TRUE  dpar=hu   mu_family=negbinomial
zero_inflated_poisson    is_hurdle=FALSE dpar=ERR  mu_family=ERR
```

The last row is the important one: the zero-inflated count families correctly stay **off** the two-block path, so D4 is preserved in code and not only in prose.

### ⚠️ Finding — the `?ecx` / `?nsec` doc touch was not done

The queue named this as the single toxval contact point:

> The only contact is the roxygen family list at `R/ecx.R:25-26` and `R/nsec.R:18-19` — a one-line doc touch in a migrating file. Acceptable; **note it in the PR so the migration carries it.**

Neither file is in this diff, and both still say:

```
#' @param dpar For a joint two-block fit only (`family = "hurdle_gamma"`
#' or `"zero_inflated_beta"`), the parameter block to report...
```

So a user who fits `hurdle_poisson` and reads `?ecx` is told `dpar` does not apply to their fit. **It does** — the dispatch above shows `ecx(fit, dpar = "hu")` will work correctly. This is documentation-only, but it is wrong rather than merely incomplete, and it is exactly the line the migration was supposed to carry across.

Two ways to close it, and the choice is really about the migration rather than about this PR:

1. Make the one-line edit here, accepting a touch in a migrating file — which is what the queue sanctioned.
2. Leave it and record it against the migration explicitly, so it is not lost.

I would take (1): it is two lines, and the alternative depends on a handoff that has not been written down anywhere yet.

### Also worth noting

- `R/sysdata.rda` is rebuilt and `data-raw/sysdata.R` updated together, so the binary is reproducible from source rather than hand-edited.
- The behaviour change is flagged in `NEWS.md` under its own **Behaviour change:** heading, with the `mu / (1 - exp(-mu))` bias stated explicitly and the reason it never bit `hurdle_gamma`. That is the standard the 2.1.4 tier set and it is met here.

### Tests

`test-hurdle_counts.R` (97 lines, new) plus updates to `test-zero-inflated-counts.R`. A run of the hurdle and zero-inflated test files is in flight at the time of writing; I will follow up on this PR if anything fails. The suite ran green in CI on this branch.

**Recommend merge** once the `?ecx`/`?nsec` line is decided — that is the only thing I would want changed before this lands.

beckyfisher:
Test run from the review above, now complete — all green on the branch:

```
test-hurdle_counts.R           pass=23  fail=0 err=0 skip=0
test-zero-inflated-counts.R    pass=43  fail=0 err=0 skip=1
test-hurdle_family.R           pass=107 fail=0 err=0 skip=1
test-bnec_hurdle.R             pass=49  fail=0 err=0 skip=1
```

Nothing to add to the review. The `?ecx`/`?nsec` `dpar` documentation remains the only thing I would want changed before merge.

beckyfisher:
Addressed the review finding on this PR.

`?ecx` and `?nsec` now list all four two-block families for `dpar`:

```
#' @param dpar For a joint two-block fit only (`family = "hurdle_gamma"`,
#' `"zero_inflated_beta"`, `"hurdle_poisson"` or
#' `"hurdle_negbinomial"`), the parameter block to report:
```

`man/ecx.Rd` and `man/nsec.Rd` regenerated. No code change — the dispatch already handled the new families, which is what made this documentation-only rather than a defect.

This is the one touch in a migrating file that `01_work_queue.md` sanctioned for #209:

> The only contact is the roxygen family list at `R/ecx.R:25-26` and `R/nsec.R:18-19` — a one-line doc touch in a migrating file. Acceptable; note it in the PR so the migration carries it.

**Noting it for the migration**, as asked: the `toxval` copies of `ecx()` and `nsec()` need the same four-family list, not the two-family one.

Also merged the doc fix from PR #224 below, to keep the stack conflict-free.

Everything else in the review stands. Tests were green (23/43/107/49) and the refusal-rather-than-approximation design is the right call.

beckyfisher:

```
```

beckyfisher:
## This closes #209 only halfway — recording why, so the remainder is explicit

The refusal added here is correct and should stay. But it leaves the count
hurdles materially less capable than `hurdle_gamma`, and the PR text frames
that as a redirect ("use `bnec(family = "hurdle_poisson")`") when it is a
permanent loss of a route. I went and checked whether the missing route can be
restored. It can, cheaply, on the bayesnec side — but **brms currently blocks
it**, and not in a version we can pin our way out of.

### 1. What counts lose relative to `hurdle_gamma`

|                                   | `bnec_hurdle()` (factorised) | `bnec(family = "hurdle_*")` (joint) |
|-----------------------------------|------------------------------|--------------------------------------|
| model set on growth block         | full set                     | full set                             |
| model set on survival block       | full set                     | **one fixed equation** (`model_survival`) |
| crossed comparison                | `crossed_weights()`, `best_crossed()` | not available                 |
| shared group-level effects        | no (breaks factorisation)    | yes                                  |

The documented workflow is *select with `bnec_hurdle()` + `crossed_weights()`,
then refit the winner with `bnec_joint()`*. For `hurdle_poisson` /
`hurdle_negbinomial` the front half of that is now unreachable —
`crossed_weights()`, `best_crossed()` and `bnec_joint()` all take a
`bayesnechurdlefit`, which counts can no longer produce.

### 2. The bayesnec-side cost of restoring it is ~20 lines

The fix is for `bnec_hurdle()` to add `trunc(lb = 1)` to the growth formula
itself — it already knows it has subset to `data[y > 0, ]`, so the truncation
is a fact about the subset, not a user modelling choice. Nothing needs to be
exposed publicly.

`rate()` was expensive in #136 because it carries a *variable*: five
touchpoints (the `validated` regex, a `ra_call`/`ra_var` slot in
`split_calls()`, a position in `short_form`, a name in `pop_vars`, and lockstep
with `retrieve_var()`'s positional lookup). `trunc(lb = 1)` carries no
variable, so none of that applies.

Tested on a worktree at this branch with **only** the gate at
`R/bayesnecformula.R:507` widened to `...|trunc\\(`:

```
formula parsed OK
model.frame OK: TRUE
  cols: y, x                      # trunc dropped; no data column needed
  bnec_pop: y_var=y x_var=x       # slots untouched
wrangle_model_formula OK: TRUE
  y | trunc(lb = 1) ~ top * exp(-exp(beta) * (x - nec) * step(x - nec))
get_priors OK: TRUE
```

`split_calls()` ignores aterms it does not recognise, so `trunc()` never enters
`short_form` or `pop_vars`; `wrangle_model_formula()` assigns
`brms_bf[[1]][[2]] <- lhs(formula)`, so it reaches brms verbatim. The only
other work is a small `add_aterm()` helper for `bnec_hurdle()`, since the
growth call passes `formula` through unchanged and `swap_response()` strips
aterms.

### 3. The likelihood would be exactly right, not an approximation

```
trunc(lb=1) + poisson :  poisson_lpmf(Y[n] | mu[n]) - poisson_lccdf(lb[n] - 1 | mu[n])
hurdle_poisson        :  poisson_lpmf(y | lambda)   - log1m_exp(-lambda)
```

`poisson_lccdf(0 | lambda)` *is* `log1m_exp(-lambda)`. Identical. And
Bernoulli + zero-truncated-positive sums exactly to the joint hurdle lpmf, so
the elpd additivity `crossed_weights()` relies on would be exact.

### 4. brms blocks it, on CRAN and on dev

Stan and the R post-processing disagree about whether `lb` is inclusive.
`log_lik_truncate()` uses `cdf(lb)` and `posterior_epred_trunc_discrete()` sums
`(min_lb + 1):max(vec_ub)` — both **exclusive** of `lb`. The generated Stan
uses `lb - 1`, i.e. **inclusive**.

Fitted `y | trunc(lb = 1) ~ x`, poisson, first observation, lambda = 0.6225, y = 1:

| quantity | value |
|---|---|
| `log_lik[1,1]` | **0.9487** |
| `dpois - log(1 - ppois(1, lambda))` → P(Y>=2) | **0.9487** (match) |
| `dpois - log1m_exp(-lambda)` → P(Y>=1), what Stan used | -0.3273 |
| `posterior_epred[1,1]` | **2.2299** = E[Y \| Y>=2] |
| E[Y \| Y>=1] = lambda/(1-e^-lambda) | 1.3433 |

So the model that gets *fitted* is right, and `loo()`, `log_lik()`,
`posterior_epred()`, `fitted()`, `ecx()`, `nsec()` and `crossed_weights()` are
all silently wrong. `loo()` returns without a warning.

Note this is a bug, not a semantics choice: `posterior_predict()` on the same
fit returns a minimum of 1, i.e. it is Stan-consistent. One fit, three
semantics.

**`trunc(lb = 0)` is not a workaround — it is worse.** Stan then computes
`poisson_lccdf(-1 | mu)` = 0, i.e. no truncation at all, while the R side
normalises correctly:

```
lb=0 coefs : 1.5991  -0.3770
untruncated: 1.5993  -0.3753     # the same fit
lb=0 log_lik == zero-truncated?  max diff: 4.4e-16
```

An untruncated fit carrying a correct-looking `loo` for a model nobody fitted.

(Separately: `posterior_epred_trunc_poisson()` caps `ub` at `3 * max(mu)`, so
it is an approximation even where the bound is right.)

### 5. Upstream status — and the version question

- **CRAN brms 2.23.0** (2025-09-08, current): both bugs present.
- **brms master**: `log_lik` **fixed** — `lb <- ceiling(lb) - 1` for integer
  families, NEWS: *"Normalize truncated `log_lik` by `P(lb <= Y <= ub)` for
  integer responses, matching the generated Stan code, so that `loo` and `waic`
  agree with the model that was fitted"* (paul-buerkner/brms#1903). Unreleased.
- **`posterior_epred_trunc_discrete()` / `_poisson` / `_negbinomial` on
  master**: byte-identical to 2.23.0. **Not fixed**, and no NEWS entry.

There is therefore no brms version, released or unreleased, where this works
end to end. Bumping our minimum brms would buy correct `crossed_weights()` once
the `log_lik` fix ships, but `ecx()`, `nsec()` and every plot would still be
wrong. Pinning is not a route out of this one.

### 6. What this means for the PR

The refusal should stay, and merge as is. What I would change:

1. **Reword the justification.** The PR gives scope (#136 having just made
   unvalidated aterms an error) as the reason. The stronger and more durable
   reason is that the route does not currently work in brms *and fails
   quietly*. That is what should be written into
   `check_hurdle_growth_family()` and the code comment, so nobody re-opens it
   on scope grounds alone.
2. **`R/bnec_hurdle.R:51-54`** still says a zero-truncated count family "is not
   yet available" — this PR makes that false. Update it to say the factorised
   route on counts is blocked on brms post-processing, with the issue number.
3. **Open a follow-up issue** for internal `trunc(lb = 1)` in `bnec_hurdle()`,
   gated on (a) a CRAN brms carrying #1903 and (b) an epred fix, and reference
   it from the error message so it reads as deferred rather than permanent.
4. **Report the epred bug upstream.** It is the same two-line shape as #1903 —
   sum from `min_lb`, divide by `cdf(ub) - cdf(lb - 1)` for integer families.
5. **The hurdle vignette should state plainly** that the crossed workflow is
   Gamma/Beta only for now. Otherwise a reader follows it with count data and
   hits the error halfway through.

### 7. One design question that survives the brms fixes

Even with correct brms, the two routes would not agree for counts.
`posterior_epred` on a truncated growth fit gives **E[Y | Y > 0]**; the joint
`hurdle_poisson` `mu` block gives **lambda**. Those yield different ECx for the
same data. `hurdle_gamma` has no such split, because a Gamma's epred *is* `mu`.

E[Y | Y > 0] is arguably the better choice — it makes the combined endpoint
exactly `E[Y|Y>0] * P(Y>0) = E[Y]` — but if we take it, `ecx(dpar = "mu")` on
the joint fit needs to be defined to match, or the two routes silently
disagree. Worth settling in the follow-up issue before any of it is built.

beckyfisher:
## Re-verified the brms claim before acting on it — it holds, with two corrections, and it is already reported upstream

I went back over [the comment above](https://github.com/open-AIMS/bayesnec/pull/225#issuecomment-5393333688) rather than take it on trust, and checked it three ways: reading the brms sources, reading the generated Stan, and running a real fit. The substance stands. Two details in it were wrong, and one of the five recommendations turns out to be already done by someone else.

### It reproduces

`make_stancode(y | trunc(lb = 1) ~ x, family = poisson())` emits `poisson_lpmf(Y[n] | mu[n]) - poisson_lccdf(lb[n] - 1 | mu[n])` — support `Y >= lb`, **inclusive**. `extract_valid_sample()` keeps `x >= lb & x <= ub`, so `posterior_predict()` is inclusive too. And decisively, `make_standata()` **accepts `y == lb` and errors on `y < lb`**: brms's own data validation treats `lb` as inclusive, so this is not a convention disagreement.

Real fit, cmdstanr, zero-truncated Poisson, `y | trunc(lb = 1) ~ x`:

| quantity | brms 2.23.0 | inclusive (what Stan fitted) |
|---|---|---|
| `log_lik[,1]` | −0.4625 | −1.2136 |
| **implied pmf summed over `y >= 1`** | **2.1864** | 1.0000 |
| `posterior_epred[,1]` | 2.5139 (= E[Y \| Y>=2]) | 1.7546 (= E[Y \| Y>=1]) |
| `min(posterior_predict[,1])` | 1 | 1 ✓ |

A truncated pmf that sums to 2.19 settles it. `trunc(lb = 0)` also reproduced: coefficients match a plain untruncated fit (0.905/0.399 vs 0.899/0.407) while `log_lik` equals the zero-truncated value to 4.4e−16.

### Two corrections to that comment

1. **"One fit, three semantics" is two, not three.** `log_lik` and `posterior_epred` are both exclusive and agree with each other; Stan and `posterior_predict()` are both inclusive. Worth having right before anyone quotes it.
2. **The `lb = 0` attribution was inverted.** It said Stan is wrong there and "the R side normalises correctly". It is the other way round: under the inclusive convention `lb = 0` means no truncation, which is exactly what Stan does; the R side wrongly zero-truncates. PR [paul-buerkner/brms#1922](https://github.com/paul-buerkner/brms/pull/1922) uses `ceiling(lb) - 1`, giving −1 at `lb = 0`, which confirms Stan is the reference. The practical advice — don't reach for `lb = 0` — is unchanged.

### Recommendation 4 is already done, by someone else

Both halves were reported upstream by @deeb01, **not by us**. We should not file a duplicate:

- **[paul-buerkner/brms#1903](https://github.com/paul-buerkner/brms/issues/1903)** (`log_lik` off by one) — closed 2026-08-24 by [#1922](https://github.com/paul-buerkner/brms/pull/1922), `lb <- ceiling(lb) - 1` guarded by `use_int()`. On master, **unreleased**.
- **[paul-buerkner/brms#1923](https://github.com/paul-buerkner/brms/issues/1923)** (`posterior_epred()` drops the lower bound for truncated discrete models) — filed 2026-08-21, **still open**. Names binomial, poisson, negbinomial, negbinomial2 and geometric, plus the knock-on to `fitted()` and `conditional_effects()`.

#1922 touched `NEWS.md`, `R/log_lik.R` and its tests only — `posterior_epred_trunc_discrete()` on master is byte-identical to CRAN 2.23.0. So the conclusion is unchanged: **there is no brms version, released or unreleased, where the factorised count route works end to end.**

### What has been done here

- **#249 opened** — the deferred internal `trunc(lb = 1)` work, with the prototype evidence, the exact-likelihood argument, the two-condition gate (a CRAN brms carrying #1922, *and* #1923 fixed), and the `E[Y | Y > 0]` vs `lambda` design question that survives both fixes. Recommendations 1–3 of the comment above.
- **2ca6374 pushed** — `family_growth`'s docs still said a count family is accepted and that no zero-truncated count family exists; this PR makes both false. The code comment now leads with the brms blocker rather than with #136's scope, since scope is the weaker reason and invites reopening, and the refusal message says the route is deferred on an upstream bug rather than reading as permanent. Tests green (`test-hurdle_counts.R` 23, `test-bnec_hurdle.R`, `test-zero-inflated-counts.R`).

Watching #1923 is the trigger to pick #249 back up.

beckyfisher:
## Retargeted from `issue-136-rate-aterm` onto `dev`

`issue-136-rate-aterm` is now an ancestor of `dev`, so the base branch this was
stacked on no longer represents anything distinct and the diff shown here was
the whole stack rather than this change. Retargeted onto `dev`; it now shows the
18 files that belong to #209.

**Still a draft, and still parked.** Nothing about the gate has changed:
brms#1923 is unresolved and #249 is open, so the count hurdles wait for the
factorised route. Reverse with `gh pr ready 225` when that gate lifts.

The note in the body that "#226 → #227 → #228 and #238 are stacked above this
branch and are parked with it" is out of date. #226 merged on 2026-08-24 and
#238 now targets `dev` directly; #238 was still carrying this PR's files as a
remnant of that stack and they have been removed from it, so this PR is again
the only place they exist.

beckyfisher:
[claude]

## What

A documentation pass over the package (local branch off `dev`, not yet pushed)
cut about 55 lines from `bnec.Rd` that argued why `zi` is held constant for
`zero_inflated_poisson` and `zero_inflated_negbinomial`. The operative
behaviour stayed in the help file; the argument did not, because a help file
describes behaviour rather than defending a design decision.

That argument is the long form of this PR's section *Why these are two-block
where the zero-inflated counts are not*. It is reproduced in full below so that
it can be placed when this branch lands.

Nothing is asked of this branch before it merges. The audit branch does not
touch any file this PR changes.

## Why it matters

The cut text has no home at present. `vignette("example6")` is titled *Hurdle
and zero-inflated concentration-response models* and covers `hurdle_gamma` and
`zero_inflated_beta`, and does not mention the count families at all. Once this
PR lands, that vignette gains two more hurdle families and the contrast with the
zero-inflated counts becomes a question a reader of it will have.

The PR body notes that a test asserts the zero-inflated count families are still
not two-block, so that #104's decision cannot be undone by pattern-matching on
the name. A test records the decision but not the reasoning, and the reasoning
is now only in `ignore/details.md` on a machine and in this comment.

## Evidence

The text as it stood in `bnec.Rd`, verbatim.

> That looks inconsistent with `"zero_inflated_beta"`, which is a two-block
> family here, and the reason is worth stating because the family names suggest
> otherwise. Zero-inflation differs from a hurdle only when the base
> distribution can itself produce a zero. Neither Gamma nor Beta can, so for
> those the two coincide: every zero must have come from the inflation
> component, the likelihood separates exactly into a Bernoulli term over all
> observations and a positive-response term over the rest, and the Stan density
> `brms` generates for `zero_inflated_beta` is the hurdle form with no mixture
> at zero. Two blocks are then an accurate description of the model.
>
> Poisson and negative binomial *can* produce a zero, so the equivalence fails.
> An observed zero is evidence about both components at once: the likelihood
> carries a `log_sum_exp` over them and does not factorise into a term for each
> block. That rules out `bnec_hurdle()`, which is precisely that factorisation
> performed as two separate fits, and it refuses these families accordingly.
>
> It does not, on its own, rule out a *joint* fit carrying a second curve.
> `brms` treats `zi` as a distributional parameter like any other and will model
> it on a predictor; the `log_sum_exp` is internal to the density and is
> indifferent to what drives `zi`. bayesnec nonetheless holds `zi` constant, for
> two reasons that outlast the factorisation argument:
>
> - `zi` and `mu` trade off against one another -- a low count rate is either a
>   small `mu` or a large `zi` -- and the data separate them only through the
>   shape of the positive counts. In a concentration-response design `mu` runs
>   from large at the control to near zero at the top concentration, so a `zi`
>   curve would be well informed at the low concentrations and almost
>   unidentified at the high ones that set the threshold. A hurdle has no such
>   problem: every zero belongs to the second block by construction, so that
>   block is informed at every concentration.
> - `zi` is a latent class rather than an observable. Nothing in the data records
>   which zeros are structural, so a threshold estimated from a `zi` curve would
>   describe something that was never measured. In the survival case the deaths
>   were observed, and that is what makes the second block interpretable.

The `log_sum_exp` point is the one this PR's compressed version leaves out, and
it is the part that rules out `bnec_hurdle()` specifically rather than a joint
`zi` curve.

<details>

### The behaviour `bnec.Rd` retains

The behaviour a user has to know, in place of the argument above:

- the count families take the ordinary single-block path, with the curve fitted
  to `mu` and a single constant `zi` estimated alongside it, and no
  `model_survival`;
- `bnec_hurdle()` refuses them, because a count distribution can itself produce
  a zero and the likelihood does not separate into a term per block;
- `disp()` is refused for both, because the dispersion parameter describes the
  count component while the response is the mixture;
- `posterior_epred()` returns `mu * (1 - zi)`, so `predict`, `fitted`,
  `autoplot` and the stored `pred_vals` sit a factor `1 - zi` below the `top`
  and `bot` that `summary()` reports on the `mu` scale, while `ecx()` and the
  no-effect threshold are unaffected;
- the rule for choosing: where the structural zeros can be identified the model
  is a hurdle, and zero-inflation is for the case where they cannot.

It also still says that a genuine hurdle on a count response needs a
zero-truncated count family, pointing at `family_growth` in `bnec_hurdle`. That
sentence will need revisiting when this lands, since it is the thing this PR
supplies on the joint route.

### A possible home for the text

`vignette("example6")` is the natural place, as a short section beside the
existing treatment of `hurdle_gamma` and `zero_inflated_beta`. It needs no new
model fits, but example6 is precompiled, so adding it means a re-render.

The alternative is to leave it here. The argument is on this PR, on #104 and in
the audit's `ignore/details.md`, and a reader who goes looking will find it. The
case for the vignette is that a reader choosing a family will not go looking.

### Its present location

`ignore/details.md` entry 1, on the machine the audit ran on. That file is
git-ignored and is for review rather than for keeping, so this comment is the
durable copy.

</details>

beckyfisher:
Independent review completed after five cycles. The findings that identified
real problems, all fixed in this PR, were:

- Truncated-count expected predictions could omit a large negative-binomial
  tail; several summaries and diagnostics still used that approximation, and
  Poisson dispersion used the untruncated variance. These paths now use exact
  conditional-positive means and variance.
- Separate negative-binomial mean and shape predictions could select different
  posterior draws. They now share one draw index.
- Combining count truncation with `cens()` produced an invalid censored
  likelihood. Count growth now rejects that combination before fitting.
- Relative ECx and ECNSEC asymptotes were sometimes left on the underlying
  count-mean or link scale, including model-averaged results. They now match
  the conditional-positive prediction scale and preserve draw pairing.
- Poisson family routing did not consistently support brms-specific
  `softplus` and `squareplus` links. It now uses the fitted brms link functions
  and family constructor throughout.

Regression tests cover each case. The reviewer verified the final responses
and agreed that the PR is sound.

Issue #56, opened 2021-06-02
Add model validation plotting methods


beckyfisher:
It would be good to have some model validation plot methods, along the lines of:
    https://cran.r-project.org/web/packages/DHARMa/vignettes/DHARMa.html#binomial-data
to allow more thorough model validation.

beckyfisher:
Closing this into #148, where model validation is now scoped as a whole rather than as plots alone.

The substance is carried over, with one change of means. `DHARMa`'s core output is a uniform quantile residual; the Bayesian equivalent is already reachable inside the current dependency set as LOO-PIT, via `pp_check(fit, type = "loo_pit_overlay")`. `bnec()` already attaches the `loo` criterion at `R/add_criteria.R`, and `loo` is in `Imports`, so this needs no new dependency. `DHARMa`'s tests are also frequentist in framing — a p-value against a null of correct specification — which sits awkwardly beside everything else the package reports.

So #148 delivers `pp_check()` methods for `bayesnecfit`, `bayesmanecfit` and `bayesnechurdlefit`, documents the LOO-PIT recipe, and adds the local variance and control lack-of-fit checks that a plot alone does not give. See the scoping there.

Reopen or raise fresh if `DHARMa` is wanted by name later.

Issue #219, opened 2026-08-21
Add a dedicated vignette showing a complete analysis workflow, including model checking and model exclusion


beckyfisher:
## The gap

There is no vignette that shows what a **complete analysis** looks like from data to reportable toxicity estimate, including the model-checking and model-exclusion steps that any real application requires.

`example2` comes closest — it checks chains, priors, and Rhat — but it is also the *first* introduction to multi-model averaging, and it should not be made longer or more detailed. Its job is to explain what a `bayesmanecfit` is; it is not the place for a full diagnostic protocol.

What is missing is the middle of the workflow. `example2` demonstrates the diagnostics one at a time and then says, in effect, "there is no harm in leaving in poor models with low weight". That is true for a badly *shaped* model with low weight, and it is not true for a model that has not converged — the vignette says so, but never shows the consequence, and never shows a model being removed. In practice the exclusion step is not optional and it is not incidental: it changes the model set, and therefore changes the weights and the model-averaged N(S)EC that gets reported.

## Proposal

A dedicated vignette — working title *A complete analysis workflow* — that runs end to end:

1. **Data preparation and family choice.** Including the pre-fit screen actually used in practice: fit one cheap single model, look at `dispersion()`, and switch binomial → beta-binomial where it exceeds 1. This is a decision point that currently appears nowhere in the docs.
2. **Fitting the candidate set**, with a word on `iter`/`warmup`/`adapt_delta`/`max_treedepth` when convergence is hard, and on `amend(add = )` for models that fail on the first pass.
3. **Sampler diagnostics.** Rhat, divergent transitions, chain mixing. Per candidate model, as a table, not one at a time.
4. **Fit diagnostics.** `pp_check()` and `check_fit()` — does the fitted model reproduce the data's variability and its level at the control.
5. **Exclusion.** Screening on the diagnostics, then re-inspecting the weights. This is the step with no coverage anywhere at present.
6. **Reporting.** Weights table, model-averaged estimate, and what to state in a methods section about which models were retained and why.

### What it needs to contain

- **More than one family.** At minimum a continuous case and a proportional/binomial case, since the family choice in step 1 and the fit checks in step 4 both behave differently between them.
[...]

PR #226, opened 2026-08-21
Add check_fit() and pp_check() methods (#148, #56)

### LINKED ISSUE #148: Add more diagnostic methods and tests for bayesnecfit objects

beckyfisher:
**Base retargeted to `dev`.** This was item 7 of the 2.2.0 stack, cut from
`issue-209-hurdle-counts` (#225) behind `issue-136-rate-aterm` (#224). #225 is
stalled on a `brms` bug and both are `DIRTY`, so this branch has been rebased
straight onto `dev` rather than waiting on them.

Nothing here depended on either. `bnec_hurdle()`, `is_hurdle_family()`,
`hurdle_component_preds()` and `bayesnechurdlefit` are all already on `dev`, and
the hurdle test fixture is a continuous gamma hurdle, not one of the count
families #225 adds. Rebasing rather than merging was necessary — a merge would
have dragged all of #224 and #225 into `dev`.

Closes #148, together with #240 which carried Part D. #56, already closed, asked
for the model-validation plotting this adds.

## What it delivers

Parts A, B and C of #148, plus decisions **(b)** and **(d)**. Part D — the
sampler diagnostics, `check_sampling()` and `screen_models()` — is not here; it
shipped separately as #240 and is on `dev`.

### `check_fit()` — the substance

Reports, per group of the predictor, the observed **location and scale** of the
response against what the fitted model simulates, with a posterior predictive
p-value for each and the control group flagged. Named to sit alongside
`check_chains()` (the sampler), `check_priors()` (the priors) and
`check_sampling()` (convergence): this checks the fit against the data.

**It reproduces the finding the issue rests on.** On `manec_example`:

| | value |
|---|---|
| global `dispersion()` | **1.011 [0.71, 1.44]** — a clean bill of health |
| `check_fit()` control group `sd_ratio` | **0.796** — the model simulates ≈26% *more* variability than the data show |
| steep tail group `sd_ratio` | **2.212**, `ppp_sd` 0.00 — failing the opposite way |

The scoping comment measured ~27% by hand; the function gets ≈26%. A single
global number averages both of those away, which is the argument for the check
being local. That comparison is pinned as a test — it is the one assertion that
would catch an implementation that looks right but is not actually local.

Why it matters: `nsec()` sets its reference from the `sig_val` quantile of the
posterior of the control mean. Overstate control variability and the reference
falls, the curve crosses later, and *NSEC* moves right — a less protective
number. Nothing previously reported this.

### Design points that are not obvious

- **Residuals, not raw values.** Within a group the raw SD mixes residual
  variability with the slope of the curve across that group, which would make
  every steep region look overdispersed.
- **Residuals taken against the same fitted mean** for observed and simulated
  response. Using each draw's own mean would subtract out exactly the
  discrepancy being measured.
- **Per-model rows with weights** for a `bayesmanecfit`. Stacking weights come
  from a global `elpd`, so a candidate can hold high weight while fitting the
  control badly. Here `nec4param` (*wi* 0.827) and `ecx4param` (*wi* 0.173) both
  show the control at 0.73–0.79, which is precisely that case.
- **Zero-fraction columns** for the mixture families: `mean(y == 0)` against
  `mean(yrep == 0)`. Whether the zero fraction is right is the whole question
  those families exist to answer, and nothing else reports it.
- **Grouping prefers replication, falls back to binning with a warning.**
  Requiring replication would refuse `nec_data`, the package's own example.
- **`ndraws` is clamped to what the fit holds.** `brms` errors rather than
  truncating when it is exceeded, and `manec_example` carries 100 draws, so a
  default of 1000 would have made `check_fit()` fail on the package's own example
  object — exactly what someone runs a diagnostic on first. Found by running it,
  not by reading it. Tested.

`plot.checkfit()` shows the same table graphically: per group, the observed
statistic against the 95% span of what the fit simulates, in separate location
and scale panels with the control drawn apart. The table stores the spans
(`sim_mean_lo/hi`, `sim_sd_lo/hi`) and `print.checkfit()` drops any column
matching `_(lo|hi)$`, so the console output is unchanged.

### Decision (d) — the combined hurdle check

`check_fit()` on a `bayesnechurdlefit` returns a third element, `combined`,
alongside `growth` and `survival`. It draws alive/dead from the survival fit and
a growth value for each survivor, reconstructing the response **as the user
handed it in, zeros included** — which neither per-component table asks about.
It rests on the same fact `hurdle_component_preds()` does: the hurdle likelihood
factorises, so the two posteriors are independent.

The observed response does not exist as a vector anywhere — `bnec_hurdle()`
splits the data into a survival frame (every exposed row) and a growth frame
(survivors only). `hurdle_observed_response()` reconstructs it, resting on
`bnec_hurdle()` subsetting rather than reordering. That assumption is **asserted
in a test rather than assumed**, because if it ever stops holding the check
silently compares against the wrong vector.

### Decision (b) — surfacing it automatically

A message at the end of `bnec()` **and** a line in `summary()`, both thresholding
on the **ratio** of observed to simulated, not on the `ppp`.

That is measured, not stylistic: on two independently fitted parameterisations of
the same simulated data the simulated control mean came out at 5.5–5.6 against an
observed 4.50 and a true 4.77 — a ~19% overshoot reproducing across fits — while
**both `ppp` values sat at about 0.82 and neither came near flagging**. A `ppp`
threshold would stay silent on exactly the case the check exists to catch.

`fit_ratio_cutoff = 1.15`, **confirmed by RF**. The spec settles that the
threshold is on the ratio and not the `ppp`, but does not name a value; 1.15 is
tight enough to catch that documented 19% case with margin. Exposed as an
argument to `summary()`, and `check_fit = FALSE` switches the axis off.

The summary block covers both axes in one, per D5:

```
Rhat > 1.01 (no convergence):
  -  nec4param
  -  ecx4param

control observed/simulated ratio beyond 1.15 (see ?check_fit):
  -  nec4param
  -  ecx4param

Convergence failures should be dropped (see ?amend or ?screen_models).
A control mis-fit is a result, not a fault: inspect it with check_fit().
```

The advice differs by axis deliberately: a model that did not converge is
unusable, while one that converged but reproduces the control badly is a
*modelling result*, and dropping it silently would hide what the user most needs
to see.

Recomputed on every call rather than cached, because #180 / PR #205 removed the
stored prediction matrices deliberately and stashing one back here would undo
that. Thinned to 200 draws to keep the recompute cheap.

### `pp_check()` methods, and #56

Dispatch for `bayesnecfit`, `bayesmanecfit` and `bayesnechurdlefit`.
`pp_check(x, type = "loo_pit_overlay")` gives a LOO-PIT check — the Bayesian
counterpart of a uniform quantile residual — on the `loo` criterion `bnec()`
already adds. **No `bayesplot` and no `DHARMa` dependency**, as decided.

For `bayesmanecfit` it runs per candidate model rather than on the averaged
posterior: the averaged posterior predictive is a weighted mixture and `brms` has
no object representing it. `check_fit()` is what reports on the average. For
`bayesnechurdlefit` it returns one result per component, following
`dispersion()`'s precedent — the two components carry different response vectors.

## What the restack onto `dev` needed

Two conflicts, both against #240, which touched the same warning block from the
other side. In both the resolution keeps this branch's structure and grafts
#240's hardening onto it:

1. **`R/print.R`** — kept the single two-axis block (per D5, the sampler question
   and the fit question feed the same drop-or-keep decision), with #240's two
   guards added. `which()` rather than logical indexing on *both* axes, so a
   `manecsummary` stored by an older version carrying an `NA` cannot print
   `- NA` as though a model called `NA` had failed; and the `rhat_cutoff`
   fallback stays **1.05**, not 1.01, because an object stored before that field
   existed was assessed against the old 1.05 grep and reporting 1.01 would
   attribute the wrong threshold to it. `fit_ratio_cutoff` gets the same
   treatment for the same reason.
2. **`R/summary.R`** — `chk_number`, not `chk_numeric`, for both cutoffs: the
   documented type is a vector of length 1 and `chk_numeric` admits any length.

**NEWS placement changed.** The entries go under `# bayesnec 2.1.4` rather than
opening a `# bayesnec 2.2.0` heading as this branch previously did. Part D of the
same issue already shipped into 2.1.4 via #240, and `dev` is still at 2.1.3.x
working towards 2.1.4 — opening a 2.2.0 section for the other half of an issue
whose first half sits in 2.1.4 would split one issue across two releases. A
one-line move if the 2.2.0 tier is restored before release.

Version 2.1.3.19 → 2.1.3.20.

## Note on the comment thread below

The earlier comments are kept deliberately, including the retracted review and
its correction. They record a real validation failure — a review checked against
`01_work_queue.md` instead of the authoritative spec at
`notes/tasks/148-model-fit-diagnostics.md`, and inherited four errors from it —
and that record is worth more intact than tidied. **This description supersedes
them where they disagree**; in particular, the original description said decision
(d) was left unimplemented, and it is implemented here.

## Tests

`tests/testthat/test-check_fit.R`: shape and completeness (every observation in
exactly one group); exactly one control group and it is the lowest predictor
value, matching what `nsec()` reads; p-values are probabilities; `ndraws`
clamping; binning warns and replication does not; explicit `group` honoured;
`ppp_value()` tail behaviour including degenerate input returning `NA` rather
than a silent 0 or 1; per-model rows carrying the object's own weights rather
than recomputed ones; the plot's interval columns; the hurdle response
reconstructed exactly, zeros included; `combined = FALSE` skipping it; and the
local-versus-global comparison above.


beckyfisher:

Reviewed independently of authoring. **Verdict: the substance is right and the two findings that shaped the design are correctly implemented. One settled requirement is missing and is not flagged.**

### The two findings from the scoping comment are honoured

These were the whole justification for the design, so they are what I checked first.

**The check is local.** `dispersion()` reports one global statistic, and for any family with a free dispersion parameter that parameter absorbs exactly the discrepancy the global statistic measures. `check_fit()` reports per predictor group with the control flagged, which is the only way the control-region discrepancy becomes visible. The NEWS entry states the `manec_example` numbers — global ratio 1.011 [0.71, 1.44] against ~26% excess simulated variability in the control region — so the justification travels with the code rather than living in an issue comment.

**The scale statistic is residual-based.** `check_fit.R:170-183` takes both observed and simulated SDs against the fitted mean. This is the one that would have been easy to get wrong: within a group the raw SD mixes residual variability with the slope of the curve across the group, which makes every steep region look overdispersed.

### The rest of the settled scope

- **Grouping** prefers genuine replication and falls back to binning with a warning (`check_fit_groups()`, `:108-113`) — as decided.
- **Per-candidate rows for a `bayesmanecfit`**, carrying `wi` alongside, so the table can answer which model is doing the damage. Correct, and the reason matters: stacking weights come from a global `elpd`, so a candidate can hold high weight while fitting the control badly.
- **Zero fraction** reported for the mixture families (`is_mixture`, `:198-205`) — observed against simulated. Nothing else in the package reported it, and it is the question those families exist to answer.
- **No `bayesplot`, no `DHARMa`.** `importFrom(brms, pp_check)` adds nothing, since `brms` is in `Depends`.
- **LOO-PIT** via `pp_check(x, type = "loo_pit_overlay")`, using the `loo` criterion `bnec()` already adds — so #56 is genuinely closed rather than deferred.
- **Decision (d)**, the combined hurdle check, is correctly left out **and flagged in the PR body**, which is exactly what the queue asked for.

### ⚠️ Finding — the plot half of part B is missing, and is not flagged

`01_work_queue.md` item 7 lists among the settled decisions:

> deliver **both** a numeric test and a plot

This PR delivers the numeric test. There is no plot: `NAMESPACE` gains `print.checkfit` and nothing else, and `R/check_fit.R` contains no `ggplot`, `geom_` or `autoplot` anywhere.

Unlike decision (d) — which is deliberately omitted and says so — the plot is not mentioned in the PR body at all, so this reads as complete when it is not. That is the part worth correcting regardless of what is decided about the plot itself: a reviewer or a future session cannot tell a considered omission from an oversight.

On the substance, I would not block the merge for it. The numeric table is the load-bearing half, and the natural home for the plot is the model-checking section of a vignette, where it can be shown against a fit that actually fails. But it is a settled requirement, so it needs either implementing or an explicit decision to drop it, and the same argument that made the check local applies to the plot: a per-group table with a dozen rows is harder to read than a caterpillar of observed-vs-simulated intervals, and the control row is exactly the one a reader would want to see rather than scan for.

### One thing to raise now rather than at migration

This PR is the anchor for **#219** (the complete-analysis-workflow vignette), which gates its steps 3 and 5 on a *"Part D of #148"* — a per-model sampler diagnostic table, a screening helper, and an `amend(drop = )` guard. **`01_work_queue.md` scopes #148 as parts A, B and C only; there is no Part D**, and this PR correctly does not implement one.

That is not a criticism of this PR — it implements what was scoped. But it means #219's stated dependency does not exist, and someone should decide whether a Part D is wanted as package API. It is not needed for the vignette: `rhat()`, `check_chains()` and `amend(drop = )` already cover both steps, which is what PR #238 does. Recorded on #219; noting it here because this is the PR a reader would look at to check.

### Tests

`test-check_fit.R` is new (114 lines). A run is in flight; I will follow up on this PR with the result. CI is green on the branch.

**Recommend merge** once the plot is either implemented or explicitly dropped in the PR body.

beckyfisher:
## Test follow-up, and a correction to something I nearly reported

**All 38 assertions in `test-check_fit.R` pass**, run locally with `NOT_CRAN=true`:

```
pass=38 fail=0 err=0 skip=0
    3.7s  check_fit returns a row per group with both statistics
    4.1s  exactly one group is flagged as the control, and it is the lowest
    3.9s  posterior predictive p-values are probabilities
    3.2s  ndraws is reduced to what the fit holds rather than erroring
    3.4s  an unreplicated predictor is binned, with a warning
    0.0s  a replicated predictor is grouped by its distinct values
    0.0s  an explicit group vector is used as given
    0.1s  ppp_value is the tail probability, and handles degenerate input
    7.3s  a bayesmanecfit reports per-model rows carrying their weights
    3.3s  check_fit reproduces the local finding a global statistic misses
```

That last one is the test that validates the entire design rationale, and it is good that it exists as an executable assertion rather than a claim in a comment.

### The correction

My first local run showed `pass=11 skip=7`, and I went looking for whether the fitting tests run in CI at all — `R-CMD-check.yaml` sets no `NOT_CRAN` in its `env:` block and passes `--as-cran`, which looked like it would skip them everywhere.

**That was wrong, and I checked before reporting it.** The CI logs for this branch show:

```
issue-148-check-fit    [ FAIL 0 | WARN 10 | SKIP 3 | PASS 1600 ]
issue-136-rate-aterm   [ FAIL 0 | WARN 10 | SKIP 3 | PASS 1538 ]
```

Three skips, not seventy-odd, and 62 more passing assertions on this branch than the one below it — so `NOT_CRAN` is set somewhere in the `r-lib/actions` chain and these tests do run in CI. No issue here. Recording the negative result so nobody else spends the same twenty minutes on it.

### One small thing that is real

`test-check_fit.R` gates its fitting tests with

```r
skip_if(Sys.getenv("NOT_CRAN") == "")
```

where the other **17** test files in the suite use `skip_on_cran()`. The two are not quite equivalent: `skip_on_cran()` runs the test only when `NOT_CRAN` is truthy, whereas this runs it whenever the variable is non-empty — including `NOT_CRAN=false`, which would run the fitting tests when someone had explicitly asked not to.

In practice it makes no difference on CRAN (the variable is unset there, so both skip) or in this CI. But it is a one-file deviation from a convention the rest of the suite follows, and the deviation is in the permissive direction. Suggest `skip_on_cran()`.

### Unrelated observation for the release gate

`WARN 10` appears on every platform and on the branch below this one too, so it is pre-existing rather than anything this PR introduced. Ten testthat warnings is worth a look before submission — noting it here for phase 4 of `06_review_run.md` rather than as a finding against this PR.

beckyfisher:
Addressed both review findings on this PR. `test-check_fit.R` now passes **48/48**.

## The plot — implemented rather than dropped

`01_work_queue.md` listed *"deliver **both** a numeric test and a plot"* among #148's settled decisions, so I built it rather than asking. Two things made it cheap:

**The table only stored medians** of the simulated statistics, so there was nothing for a plot to draw against. It now also stores the 95% spans — `sim_mean_lo/hi`, `sim_sd_lo/hi` — and `print.checkfit()` drops any column matching `_(lo|hi)$`, so **the console output is unchanged**. The intervals are on the object for `plot()`; they are not in the printed table, which would otherwise run past a readable line width.

**`plot.checkfit()` is built directly in `ggplot2`**, already in `Depends`, so the no-`bayesplot` decision is untouched. Per group: the observed statistic as a point against the simulated 95% span as a bar. A point outside its bar is a group the model does not reproduce.

Two design choices worth flagging for override:

- **Two panels, not one.** Location and scale fail independently and for different reasons — a model can get every group's mean right while simulating far too much spread, which is precisely the case `check_fit()` was built to catch and which a combined panel would hide.
- **The control is drawn differently** (colour and shape), because it is not just another group: `nsec()` reads its reference from it, so a discrepancy there moves a reported no-effect concentration in a way one at the top of the curve does not.

For a `bayesmanecfit` the facets are model × statistic, so the per-model rows stay visible — they are the point of that method.

Four tests added, including one asserting the intervals are **on the object but not in the printed output**, which is the thing most likely to regress if someone later tidies the print method.

**One test of mine was wrong and the code was right**, worth recording: I asserted one control row per statistic panel, but `manec_example` is a `bayesmanecfit` with two candidate models, so it is 2 × n_models. Corrected to compute it from the table rather than hard-coding 2.

## The skip idiom

All 11 gates changed from `skip_if(Sys.getenv("NOT_CRAN") == "")` to `skip_on_cran()`, matching the other 17 test files. The old form ran the fitting tests whenever the variable was non-empty — including `NOT_CRAN=false`, i.e. when someone had explicitly asked not to.

## Unchanged

Decision (d) remains correctly out of scope and flagged. Everything else in the review stands: the check is local, the scale statistic is residual-based, per-model rows carry their weights, and the mixture zero fraction is reported.

Also merged the doc fixes from #224 and #225 below, to keep the stack conflict-free.

beckyfisher:
## Retracting my review above — it validated against the wrong source

RF asked me to re-read the #148 thread against `01_work_queue.md`. **The queue entry was wrong in four ways, my review inherited all of them, and this PR is materially less complete than I said.**

I checked this PR against the queue entry. I never opened the issue, and I never opened **`notes/tasks/148-model-fit-diagnostics.md`** — the authoritative 23 KB spec, written at scoping time and referenced from the thread, which covers Parts A–D, the settled decisions, the hazards and a definition of done. That is exactly the failure I spent the afternoon flagging in these PRs: trusting a claim *about* the source instead of the source.

### What the thread actually says

**1. There are four parts, not three.** The thread has a section headed *"Part D — sampler diagnostics and screening, folded in"*: D1 a per-model sampler diagnostic table, D2 the ESS decision (settled: report the absolute, threshold at 400), D3 the screening helper `check_sampling()` plus a wrapper that messages what it dropped, D4 an `amend(drop = )` guard. Folded in deliberately — both halves answer *which candidate models belong in the averaged set*, feed the same `amend(drop=)` step and the same `summary()` block, so splitting them means designing that step twice.

**2. Decision (d) was answered "in", and prototyped.** My review said it was *"correctly left out and flagged, exactly what the queue asked for."* Wrong on both halves. The Final decisions table reads `(d) | combined hurdle simulation in`, and a later comment carries a working six-line implementation with a 400 x 90 result matrix.

**3. Decision (b) is missing from this PR entirely** — a message at the end of `bnec()` **and** a line in `summary()`, thresholding on the **ratio** with the `ppp` alongside. That threshold was settled explicitly, because a `ppp` threshold stayed silent on both test fits and *"silence reads as a pass"*.

**4. A doc fix rides on (b).** `vignettes/example2.Rmd.orig:102` claims the summary warns about divergent transitions. It does not — `print.manecsummary` warns on Rhat only (`R/print.R:136`), hard-coded at 1.05 while the same vignette recommends 1.01.

### Where that leaves this PR

Checked on the branch: it touches only `R/check_fit.R` and `R/pp_check.R`. No `summary()` line, no `bnec()` message, no `check_sampling()`, nothing of Part D.

So against #148 **as actually scoped**, this delivers **A, B (numeric, plus the plot added under review), and C**. Outstanding: **(b)**, **(d)**, and **all of Part D**.

Nothing here is wrong — what is here is good, and the review's technical findings stand (the check is local, the scale statistic is residual-based, per-model rows carry weights, the mixture zero fraction is reported). It is **incomplete**, not incorrect.

### Plan, per RF: all of it pre-CRAN

- **(b) and (d) into this PR** — both are small and belong beside the code they surface; (d) is essentially already written on the issue.
- **Part D as its own PR**, taking the spec's own advice: it needs no `loo`, does not depend on PR #217, and "can be built and merged first, on its own".

`01_work_queue.md` item 7 has been corrected and now points at the spec file as authoritative.

beckyfisher:
## This PR has grown: it now carries all of #148 except Part D

Recording the change of scope, since the description above describes only parts A, B and C.

Following the correction in #239 — `01_work_queue.md` item 7 had recorded #148 as three parts with decision (d) unanswered, when the issue thread settled four parts and answered (d) *in* — this branch now also carries:

### Decision (d) — the combined hurdle check

`check_fit()` on a `bayesnechurdlefit` returns a third element, `combined`, alongside `growth` and `survival`. It draws alive/dead from the survival fit and a growth value for every exposed row; the product reconstructs the observed response including its zeros, which is the only check of the fit *as a model of the data the user handed in*.

Verified on a freshly fitted hurdle model: survival on 90 rows, growth on 44, and the reconstructed response matches the original data exactly. Two things the prototype on the issue does not cover, which only appear when it is built inside the package:

- **The observed response does not exist as a vector anywhere.** `bnec_hurdle()` splits the data into a survival frame (every exposed row) and a growth frame (survivors), so neither component's data is the response as measured. `hurdle_observed_response()` reconstructs it, resting on `bnec_hurdle()` subsetting rather than reordering — which is asserted in a test rather than assumed, because if it ever stops holding the check silently compares against the wrong vector.
- **`mu` for the residuals** is `E[growth] x P(alive)`, needing `posterior_epred` from both components on the full exposed set.

`check_fit_table()` is split so the statistics are computed once and reused with a `yrep` the combined path builds itself.

### Decision (b) — the automatic surfacing

A line in `summary()` **and** a message at the end of `bnec()`. The summary block covers both axes in one, per D5:

```
Rhat > 1.01 (no convergence):
  -  nec4param
  -  ecx4param

control observed/simulated ratio beyond 1.15 (see ?check_fit):
  -  nec4param
  -  ecx4param

Convergence failures should be dropped (see ?amend or ?screen_models).
A control mis-fit is a result, not a fault: inspect it with check_fit().
```

The advice differs by axis deliberately (D3b): a model that did not converge is unusable, while one that converged but reproduces the control badly is a *modelling result*, and dropping it silently would hide what the user most needs to see.

Recomputed on every call rather than cached, because #180 / PR #205 removed the stored prediction matrices deliberately and stashing one back here would undo that. Thinned to 200 draws to keep the recompute cheap, and `check_fit = FALSE` switches the axis off.

## ⚠️ One number is mine, not the spec's — please set it

`notes/tasks/148-model-fit-diagnostics.md` settles Rhat **1.01**, ESS **400** and divergences **10** explicitly. For this one it settles only that the threshold is **on the ratio, not the `ppp`** — and the evidence for that is worth repeating, because it is the reason the check exists:

> measured twice on independently fitted parameterisations of the same simulated data, the simulated control mean came out at 5.5–5.6 against an observed 4.50 and a true 4.77 — a ~19% overshoot reproducing across fits. **Both `ppp` values were about 0.82 and neither came near flagging.**

It does not name the ratio's value. I set **`fit_ratio_cutoff = 1.15`**, tight enough to catch that documented 19% case with margin, exposed as an argument to both `summary()` and the internal message.

**That is a judgement I made and you have not seen.** It decides what users get warned about, and it will be quoted in methods sections. Treat 1.15 as a placeholder until you have set it deliberately — the same standing the spec gives the divergence default of 10.

## Merge note, unchanged

This branch and #240 both open the `# bayesnec 2.2.0` heading; whichever merges second needs a two-line dedup. #240 is now the base of this stack, so in the intended order it merges first.

beckyfisher:

```
```

beckyfisher:
## Restacked onto `dev` — base retargeted, description rewritten

`#225` is stalled on a `brms` bug and `#224` sits under it; both are `DIRTY`.
This branch has been rebased off them and onto `dev` so it can merge on its own.

**It was separable.** Nothing here depended on either branch: `bnec_hurdle()`,
`is_hurdle_family()`, `hurdle_component_preds()` and `bayesnechurdlefit` are all
already on `dev`, and the hurdle test fixture is a continuous gamma hurdle, not
one of the count families `#225` adds. A *merge* would have dragged all of `#224`
and `#225` into `dev`, so it had to be a rebase — hence the force-push. No
inline review comments existed on the branch, so nothing was lost.

**Two conflicts, both against `#240`**, which touched the same warning block from
the other side. Both resolved by keeping this branch's structure and grafting
`#240`'s hardening onto it:

- `R/print.R` — kept the single two-axis block (per D5), added `#240`'s two
  guards: `which()` rather than logical indexing on *both* axes, so a stored
  older `manecsummary` carrying an `NA` cannot print `- NA` as a failed model;
  and the `rhat_cutoff` fallback stays **1.05**, not 1.01, because an object
  stored before that field existed was assessed against the old 1.05 grep.
  `fit_ratio_cutoff` gets the same treatment.
- `R/summary.R` — `chk_number`, not `chk_numeric`, for both cutoffs.

**`fit_ratio_cutoff = 1.15` is confirmed** — flagged above as a judgement the
spec does not settle, now RF's decision and the documented default.

**NEWS entries moved to `# bayesnec 2.1.4`**, not a new 2.2.0 heading. Part D of
this same issue already shipped into 2.1.4 via `#240`, and `dev` is at 2.1.3.x
working towards 2.1.4; opening a 2.2.0 section for the other half of an issue
whose first half is in 2.1.4 would split one issue across two releases. A
one-line move if the tier is restored before release. Version 2.1.3.19 →
2.1.3.20.

**The comments above are kept deliberately**, retracted review included — they
record a review that validated against `01_work_queue.md` instead of the
authoritative spec and inherited four errors from it, and that is worth keeping.
The rewritten description supersedes them where they disagree; most importantly,
the original description said decision (d) was left unimplemented, and it is
implemented here.

Issue #193, opened 2026-08-11
Rewrite `example7` as a growth-data case study using the `alga` dataset


beckyfisher:
## Summary

`vignette("example7")` — *Censored responses* — was removed in `4ce11e33` ("remove
example7 pending rewrite") while #189 was being prepared. Its mechanics moved into a short
*Censoring* section of `vignette("example1")` (`ee76e815`), and the argument it was really
making was re-scoped as a **growth-data case study** built on the `alga` dataset.

The plan for that rewrite currently lives only as a comment on the merged PR #189 —
[#189 (comment)](https://github.com/open-AIMS/bayesnec/pull/189#issuecomment-5235587110) —
so nothing open tracks it. This issue is that tracker, with the plan reproduced so it does
not have to be excavated from a closed PR.

**Blocked on:** #191 / #192. Section 5 below demonstrates the dispersion sub-model, and the
vignette should demonstrate the feature rather than describe it, so the interface has to
exist first.

**Status of the dataset:** `alga` already ships (`data/alga.rda`), is documented
(`man/alga.Rd`) and is built by `data-raw/alga.R`. Nothing in the package uses it yet — it
exists for this vignette. Full characterisation in `notes/alga_dataset.md`.

## Why the original `example7` was withdrawn rather than edited

Three problems, from the #189 review:

1. **It generalised from the case that motivated it.** It asserted that censored
   observations cluster at the high-concentration end, so the bias lands where the response
   changes fastest. That is true only for a declining response censored at a lower detection
   limit. A saturating instrument censors the *controls*; an increasing response such as
   mortality inverts it entirely. The vignette had no basis for assuming where a reader's
   censored values sit.

2. **Case A did not motivate the feature.** The `herbicide` zeros are recording artefacts —
   `fvfm` rounded to two decimal places, so a recorded `0.00` means the truth fell in
   `[0, 0.005)`. Formally that is interval censoring and treating it as such is not *wrong*,
   but it is 23 of 580 values with an interval of width 0.005 on a response spanning 0–0.73.
   It changes nothing. An example where the careful treatment and the crude one agree is an
   argument for the crude one. `check_data()`'s existing nudge to `min(y[y > 0]) / 10` is the
   right answer there, and it should be presented as a *when not to bother*.

3. **One passage was right but read as wrong.** "Left-censoring at the bound then asserts
   that the truth lies in `(0, bound]`" is correct — in left-censoring the bound is a ceiling
[...]

Issue #184, opened 2026-08-05
Implement future_apply so the model set can be fitted in parallel


beckyfisher:
## Summary

`bnec()` fits its model set one model at a time. A 23-model run is 23 sequential
`brm()` calls, each of which typically leaves most cores idle. Fitting the models
in parallel is the single largest available speed-up for multi-model runs, which
is the normal way the package is used.

This issue proposes using the `future` ecosystem (`future.apply::future_lapply`)
so the model loop can run in parallel under a plan the *user* chooses, with the
sequential behaviour preserved as the default.

It is written to be worked through in order: **investigate first, resolve the
design questions, then implement**. Several of the hazards below are correctness
issues rather than performance ones, and at least one of them (RNG) will silently
produce wrong-but-plausible output if it is got wrong.

---

## Where the sequential loops are

Three call sites, two of which are the same pattern:

| file | what it does |
|---|---|
| `R/bnec.R` (~line 368, `for (m in seq_along(model))`) | the main multi-model loop; calls `fit_bayesnec()` inside `try()`, storing `NA` on failure |
| `R/amend.R` (~line 122, `for (m in seq_along(model_set))`) | same pattern, but builds `brm_args` and priors per model inside the loop |
| `R/bnec_hurdle.R` (~lines 149 and 153) | two independent `bnec()` calls, growth and survival — a second, nested level of parallelism |

`R/bnec.R` and `R/amend.R` are the targets. `bnec_hurdle()` should be considered
but is probably better left sequential at first (see nested parallelism below).

## Design questions to resolve before writing code

**1. How does the user ask for it?**

The idiomatic `future` answer is that they do not ask `bayesnec` at all — they
call `future::plan(multisession, workers = 4)` and the package simply uses
`future_lapply()`. Recommended, because it composes with whatever the user is
already doing and adds no argument to an already wide `bnec()` signature.

The alternative — a `cores` or `parallel` argument — is more discoverable but
duplicates state the plan already holds, and gets ambiguous the moment the user
has also set a plan. Worth a look at how `brms` itself handles the equivalent
choice before deciding.

**2. Nested parallelism — the central problem.**

`brm()` *already* parallelises, across chains, via its own `cores` argument.
Parallelising across models on top of that gives `workers x chains` processes,
which on most machines will oversubscribe and run *slower* than sequential.

Note that `bayesnec` currently passes no `cores` argument at all — `grep -rn
[...]

PR #228, opened 2026-08-21
Add a vignette on grouping and factor covariates (#6, #33)

### LINKED ISSUE #6: Add the capacity to deal with random structure
This capacity is now implemented in the experimental branch random_effects. This is achieved through two additional arguments. The first, random, is a named list containing model formula for the random effects that you want to include. This can include all/any of the non-liner parameters, and bayesnec will automatically use the ones relevant to a given model. The second is a character vector specifying the columns in data that contain the random effects. Below is an example of the code for implementing the random effects, where "genotype" is added as a random effect on the "nec" parameter and "chamber" is added as an offset term (ost). Note that for random offsets, the syntax requires the random list includes this as element "ost", and that "ost" is used in the model formula. Note also that this example is likely to be a sensible model, and that use of the random effects structures needs to be carefully considered, explored and validated by the user.

random_term <- list( nec = as.formula(nec ~ (1|genotype)),
ost = as.formula(ost ~ (1|chamber)))

temp5<- bnec(data = surv_24h_data,
x_var = "log.x",
y_var = "y",
trials_var = "trials",
model = "ecxll4",
iter=2e3,
random = random_term, random_vars = c("genotype", "chamber")

beckyfisher:
**Closes #6. Documents #33's `bnec_group()`.**

## What

Adds `vignette("example8")` on group-level terms and factor covariates, and the
three datasets it is built on: `lum31`, `coral_colour` and `coral_pam`.

The vignette works through what `ogl()`, `pgl()` and `(par | group)` each do to
the equation, when a grouping admits only a displacement and when it admits a
difference in shape, and when levels are fixed and of interest in themselves and
so belong to `bnec_group()` instead.

The document is an introduction to the three syntaxes followed by three case
studies, one per dataset. Every figure is on a log concentration axis, and each
dataset opens with a raw-data figure, a link to the publication behind it, and
what that study measured and estimated.

## Why it matters

No vignette or manual page states what `brms` formula each of the three
group-level syntaxes produces, so a user choosing between them has to read the
source. Nothing states which one a given design admits, and the commonest error
--- placing a random structure across levels that are fixed and of interest ---
is not addressed anywhere.

The case study gives a reason to use a group-level term that does not depend on
the reader already wanting one. Dividing a response by its control mean is
standard practice in ecotoxicology, and Ritz et al. (2026) show it biases effect
concentrations downwards and gives intervals that are too narrow. A group-level
term on the control parameter targets the same quantity with the uncertainty
propagated, so the grouping machinery is the answer to a problem users already
have.

## Evidence

Three findings from the exploratory fits decide what the sections claim. All are
recorded with their settings in `prompts/grouping-vignette-dataset.md`, and the
scripts that produced them are in `notes/scripts/example8/`.

**Normalisation is unavailable on this assay rather than merely inadvisable.**
The plate reader sets its gain per read, so dividing by the largest reading in
the dataset leaves plate controls between 0.171 and 0.898 of scale. What makes a
normalised analysis work is every curve's control landing near the top of a
bounded scale, and that needs a per-curve divisor --- which is what Ritz et al.
object to.

**A low ECx tracks model-set composition; EC50 does not.** Among the equations
holding weight, EC50 spans 17 per cent and EC10 spans a factor of four. Each
equation's own EC10 is stable across plate structures to within 10--30 per cent,
while the model average moves from 0.662 to 0.287 mg/L.

**The coral control has to be given a value before it can be fitted on a log
predictor.** `coral_colour` and `coral_pam` record the untreated control as an
exact zero, and `bnec()` refuses a predictor that reaches its data check as
`-Inf`. The treatments are a half-decade dilution --- 0.29, 0.96, 2.9, 9.6, 29
µg/L, each about 3.16 times the last --- so one further step below the lowest
treatment is 0.29/3.16 = 0.092, and the vignette substitutes 0.1 for the control
in a visible chunk. Half the lowest treatment, 0.145, would place it a third of
a step away, nearer the lowest dose than any two doses are to each other.
`data-raw/coral_diuron.R` still ships the recorded zero, so the substitution
stays a decision of the analysis rather than a property of the data.

**Estimates were reported from model averages containing equations whose chains
never converged.** `example8` did not call `check_sampling()`. Measured on the
store at `bnec()`'s defaults, the weight sitting on equations that fail the
sampler screen was 0.170 for `fit_plain`, 0.122 for `fit_ogl`, 0.617 for
`fit_pooled`, 0.000 for `fit_plate` and 0.030 for `fit_pam`. The five
single-object fits are now screened with `screen_models()` before anything is
read from them. `ecxexp` in `fit_ogl` held 0.122 of the weight with 159
divergent transitions, and its unbounded low-concentration tail put that fit's
EC10 lower bound at 0.015 mg/L, below the lowest tested concentration, and drew
a credible band reaching four million relative light units against data topping
out near six hundred thousand. Screened, that bound is 0.158 and the band is
inside the data. The two `bnec_group()` fits cannot be screened, because neither
`check_sampling()` nor `screen_models()` accepts a `bayesnecgroupfit` (#376).

**A constant-dispersion Gamma overstates the variance at the control about
eight-fold.** Observed within-cell CV runs 0.051 at the control to 0.362 at the
bottom; the fitted constant is 0.43. `disp("power")` brings the implied control
CV to 0.079, and both it and an unconstrained smooth agree there.

<details>

### The datasets

`lum31` --- acute copper and zinc tests from the Lum-31 bioluminescent bacterial
assay of Luter et al. (2025). 2904 rows, 33 plates, five batches. Replication at
three scales: `conc_group` (four wells at one concentration on one plate),
`plate` (a complete series), and toxicant by exposure time. `rlu` is reported as
recorded, including one blank-corrected negative the source did not floor, which
is the evidence that the zeros are floored negatives rather than absences of
light; `censoring` and `rlu_cens` give the left-censored form.

`coral_colour` and `coral_pam` --- two endpoints of one diuron by climate
scenario exposure, Flores et al. (2021), the study the JSS article cites for
`compare_posterior()`. Five concentrations in the colour endpoint is the
conventional minimum, which is what section 6 needs; the yield endpoint gives the
well-conditioned contrast at 54 chambers.

Source workbooks and CSVs for all three are git-ignored under `data-raw/`.
Redistribution is recorded in `notes/dataset_provenance.md`.

### Settings

The vignette fits 180 models: `model = "all"` for the `lum31` and `coral_pam`
arms, `"decline"` for the herbicide arm so it reproduces the JSS article's final
figure, and deliberate two-equation subsets for section 6's parameter-count
contrast. Sampling is at `bnec()`'s own defaults: no call names `iter`, `warmup`
or `control`, so a diagnostic reported in the vignette is one a reader gets by
writing the same call. `model = "all"` resolves to fewer equations on a logged
predictor than
on a recorded one: a concentration below 1 has a negative logarithm, which rules
out `necsigm` and `ecxsigm`, whose curves raise the predictor to a fractional
power, and `ecxhormebc5`, which cannot be reliably initialised where an
identity-linked mean must stay positive (#344). `fit_pam` fits 15 equations and
each `lum31` set 14.

Chains are 4 at `iter = 5000, warmup = 4000` --- half `bnec()`'s
default iterations and half its retained draws, keeping the 80 per cent warmup
ratio. At 16000 draws the lowest effective sample size on any contributing
equation was 2856, so 4000 draws has measured headroom over the conventional
floor of 400.

`adapt_delta` is 0.99 throughout. At the default, `ecxll5` gave 880 divergent
transitions of 16000 while holding a fifth of the weight; fitted alone it gives
0 at 0.99, so the cause is the step size rather than the design.

Precompiled on the HPC under #308, reading every fit from the store that #357
put on `dev`. The 180 units are fitted as a cluster array by
[open-AIMS/grouping-structures](https://github.com/open-AIMS/grouping-structures)
and assembled into the nine objects the vignette asks for, so the render samples
nothing and takes 33 minutes. `hpc/run.precompile` keeps the walltime raised
from 24 hours to three days for the route that does fit them. Every vignette
inherits that; say so if it should be per-run.

A fit is filed under a hash of its call and a digest of its data, so a vignette
edit that changes a fit call changes its key and the render stops rather than
loading the previous draft's fit. `assemble_store.R` refuses a set whose units
were fitted by a different `bayesnec` for the same reason.

### Known limitations

The model set is written out as `"all"` in each call rather than held in a
variable. That began as a workaround for #319, which is now closed and on `dev`,
so a variable would work; changing it would change every fit call and so every
key in the store, which is 180 units of sampling for a change of spelling.

`check_formula()` prints the formula it returns, and a `bayesnecformula`'s print
method shows its environment, so the rendered vignette includes a line reading
`<environment: 0x…>` whose address differs on every render. It predates this
branch and belongs to the print method rather than to the vignette.

</details>


beckyfisher:

Reviewed independently of authoring. **Verdict: complete against the queue's content list, and protocol-compliant. One collision to resolve, and it is mine, not this PR's.**

### Content, against `01_work_queue.md` item 9

The queue listed six things this vignette had to contain. All six are present, each with its own section:

| required | where |
|---|---|
| all three group-level term types, and when each is right | `## ogl()`, `## pgl()`, `## (par \| group)` |
| what the terms do to the curve and to the estimates | within each of those, against the ungrouped fit |
| priors for the group-level SDs, and `run_par_checks` | `## Priors for the group-level standard deviations` |
| the diagnostics and the **failure modes** | `## Failure modes, and being honest about them` |
| the #33 route alongside | `# Fitting each level separately`, `## Crossed model weights` |
| a dataset with genuine grouping structure | `nassarius` |

The dataset choice is better than it first looks. `tank` is **within** concentration — each tank has one dose — while `contaminant` is **across** concentration, spanning the full dose range. Those are the two structurally different cases, and having both in one dataset is what lets `# Which kind of grouping is it?` be a real decision rather than a definition. Restricting to survivors' growth to get a continuous response is a sensible simplification and is stated as one.

The JSS caution about hierarchical effects in a non-linear setting is carried into the closing section rather than asserted and dropped, as the queue asked.

### Protocol compliance

- **`vignettes/example7*` untouched** — #193 is another session's.
- **Only `.Rmd.orig` is committed**; `precompile.R` was not run and the rendered `.Rmd` is correctly absent. It renders under #190.
- All nine chunks parse (`knitr::purl()` → `parse()`).
- `library(dplyr)` and `library(ggplot2)` are both already declared — `dplyr` in `Imports`, `ggplot2` in `Depends` — and `example1`/`example6` already use `dplyr` the same way. I checked this specifically because a vignette dependency would not be caught by CI: `R-CMD-check.yaml` passes `--ignore-vignettes`. No issue, recording the negative result.

### ⚠️ Filename collision — my error, and I am fixing it

This PR creates `vignettes/example8.Rmd.orig`. **So does PR #238**, the #219 complete-analysis-workflow draft I opened during phase 1, branched from `issue-148-check-fit`.

Since `issue-6-33-grouping-vignette` sits above `issue-148-check-fit` in the stack, both would land at the same path and conflict hard — two entirely unrelated vignettes.

**This PR has the prior claim** — it is older, it is part of the planned stack, and `example8` follows naturally from the existing `example1`–`example7`. I picked the same name for the #219 draft without checking what the branches above had already taken, which was careless. I am renaming mine to `example9` rather than asking this PR to move.

Nothing to do here; flagged so that whoever merges knows the conflict is being resolved on the other side and does not resolve it by dropping one of the two vignettes.

### Smaller notes

- `# Choosing between the two routes` is the section that justifies the queue's decision to cover #6 and #33 together rather than separately, and it earns that decision. A reader who arrives wanting "a factor covariate" and finds two mechanisms needs exactly this.
- The one thing I cannot check is whether the prose about what the terms *do* to the estimates matches what the fits actually produce — that is only verifiable once #190 renders it. Worth a read at that point rather than assuming the narrative survived.

**Recommend merge**, subject to the stack below it and to #238 renaming (which is on me).

beckyfisher:
Restacked over the phase 3 fixes to #224, #225, #226 and #227 — no conflicts. The vignette itself is unchanged; this only brings the branch up to date so the stack merges cleanly.

The `example8` collision noted in the review is resolved on the other side: PR #238's vignette is now `example9`.

beckyfisher:
## Finding — the `nassarius` subset conditions on survival, and the deaths are dose-dependent

Raised by RF on review. Checked against the data, and it is a real problem rather than a stylistic one.

The vignette does:

```r
surv <- nassarius |> filter(alive == 1, record == "measured")
```

That discards **332 of 1208 rows — 27% — non-randomly**, and the discard rate rises with dose:

```
dose 0.3   47% dead        dose 6    47% dead
dose 1     30% dead        dose 10   33% dead
dose 2.5   83% dead        dose 15   54% dead
dose 4     47% dead        dose 20  100% dead
```

Consequences, none of which are about grouping:

1. **It conditions on survival** — precisely what `hurdle_gamma` and `bnec_hurdle()` exist to avoid, and what `vignette("example6")` warns against. A reader who read that vignette and then this one gets contradictory guidance from the same package.
2. **The top of the curve disappears.** At dose 20 every animal died, so that concentration is absent from the fit entirely — the strongest part of the concentration-response signal.
3. **High-dose growth is survivorship-selected.** At 2.5 only 3 of 18 survive, and those three are not a random sample of what was exposed. This is the same truncation argument that drove #209's refusal to fit untruncated counts to a conditioned subset.

`nassarius` is the motivating dataset for `hurdle_gamma` for exactly this reason. Using it here means the reader has to step over an unrelated modelling trap to reach the lesson the vignette is about.

## The alternative, and its cost

**`herbicide`** fits well: 7 herbicides x 6-9 concentrations x 5-12 replicates, genuine replication at every design point (so `check_fit()` groups on real replicates rather than binning), and **no mortality**, so nothing to condition on. The grouping question is also the substantive one — does the concentration-response *form* differ between herbicides — which is #33's premise.

**But it is flat.** Herbicide x concentration, nothing nested. It cannot demonstrate the **within-concentration vs across-concentration** distinction that `# Which kind of grouping is it?` is built on, and that contrast (`tank` has one dose; `contaminant` spans all doses) is a genuinely good piece of teaching.

| | nested contrast | avoids conditioning |
|---|---|---|
| `nassarius` (current) | yes | **no** |
| `herbicide` | **no** | yes |

## Two ways to have both — RF's call

1. **Switch to `herbicide`, keep the contrast as prose.** Explain within- vs across-concentration grouping using `nassarius` as a *described* example, one or two sentences and no fit, then demonstrate everything on `herbicide`. The teaching survives; the fitting is clean.
2. **Keep `nassarius` and model it honestly** — `bnec_hurdle()` or `family = "hurdle_gamma"`, so the deaths are in the model rather than filtered out. Keeps the nested structure and the real dataset, but makes a grouping vignette also a hurdle vignette.

My inclination is (1) — the vignette's job is grouping, and (2) asks the reader to learn two hard things at once. But there is a fair case that a grouping vignette which quietly sidesteps the hurdle question is doing the same thing as the filter it would be replacing.

Either way, note `herbicide` has **23 observations at or beyond the 0/1 boundary**, so it needs `Beta` with the boundary handling `check_data()` applies — worth stating in the vignette rather than leaving to be discovered.

`alga` is not an option: #193 owns `example7`, which uses it.

**Not changed pending RF's decision** — this is a judgement about what the vignette should teach, not an implementation detail.

beckyfisher:

```
```

beckyfisher:
## Handover note — state as of 2026-08-25, and the open question

Pausing here. Everything below is on the branch (`f7f3d751`) or in
`notes/tasks/245-and-example8-overnight.md` on `notes-245-example8`.

### The open question, which blocks the family choice

RF asked why the vignette fits `Beta` to a response with mass at both
boundaries. Checking the source paper settled part of it and opened something
larger.

**What the paper did.** Brinkman et al. (2023) analysed this dataset with
bayesnec 2.0.1: the proportional decline in tissue survivorship modelled with
**Bernoulli or beta-binomial** (for 1-MN, Bernoulli at T1 only and
**beta-binomial elsewhere**, so day 4 is beta-binomial), on **square-root
concentration**, with the **`decline`** model set and an identity link.

**Why `Beta` is hard to defend here.** Day 4 is 49 zeros, 60 ones, 41 interior
— **72.7% of observations sit on a boundary** and are nudged by 0.001. A hurdle
or zero-inflated Beta cannot rescue it either: the mass is at *both* boundaries,
which is the trap already documented when the Nordborg coral dataset was
rejected for #175, and bayesnec has no `zero_one_inflated_beta`.
`beta_binomial` avoids it outright, since 0 and 20-of-20 are legal values.

**But the switch does not work, for a reason nobody has explained yet.**

| configuration | max R-hat | divergences |
|---|---|---|
| `beta_binomial` / `binomial`, ungrouped, raw **or** sqrt conc | ~1.00 | **0** |
| `beta_binomial` + `ogl(chamber)` | 2.23 | **2000/2000** |
| `binomial` + `ogl(chamber)`, `adapt_delta` 0.95 and 0.99 | 2.7–2.9 | **2000/2000** |
| `binomial` + `ogl`, sd prior tightened to `student_t(3, 0, 0.02)` | 2.66 | **2000/2000** |
| `binomial` + `ogl`, `top` capped at 0.95 | 2.84 | **2000/2000** |
| `Beta` + `ogl(chamber)` | 1.00 | 0 (at `adapt_delta` 0.99) |

Curve estimates stay stable and plausible throughout (NEC 900–955); the chains
disagree on the chamber block. The generated brms formula is well-formed —
`trials(tot)` preserved, `ogl` added correctly — so it is not a malformed model.

Three explanations were proposed and **all three refuted by test**: competing
dispersion between `phi` and the chamber intercept (refuted — `binomial` has no
`phi` and behaves identically); the offset scale (refuted — tightening the sd
prior does not help); the mean bumping into 1 (refuted — capping `top` at 0.95
does not help). Recorded so nobody spends the same hours twice.

One observation that may or may not matter: the binomial families estimate
`sd(chamber)` at ~0.19, where `Beta` estimates 0.005. Nudging turns 62 ceiling
observations into 0.999 with almost no variance, while 20-of-20 under a
binomial is a genuine draw, so between-chamber differences become real signal.

### What this means for the vignette

On this dataset it appears to be **either a working grouped demonstration
(`Beta`, 73% nudged) or the published family (`binomial`/`beta_binomial`, clean
ungrouped, no working grouped fit) — not both.** That decision is open.

**It is also material to #250**, which claims group-level terms work for bounded
families generally. On this evidence that holds for `Beta` and not for the
binomial families.

### Branch state

`f7f3d751` carries **both** a verified change and an unverified one:

- **Verified:** a plot for every fit — `autoplot()` for the ogl, pgl,
  `(nec | colony)`, ungrouped and hurdle fits, plus `bnec_group()`'s own
  `plot()` method for the per-level and composed fits. That method was
  documented in the vignette but never called, which was a real gap in the #33
  half. All eight figures executed.
- **Unverified:** the switch to `beta_binomial` with the counts convention
  (`live = round(surv * 20)`, `trials` 20). The grouped chunks would currently
  render badly. Committed rather than reverted so the work is not lost; revert
  that half if the decision goes back to `Beta`.

A rendered review copy of the **pre-switch** (Beta) vignette, with all eight
plots and every chunk executed, is at
`ignore/example8-review/example8-review.html` — self-contained, and not in git.

### Two side findings worth their own issues

1. **`Beta` with a transformed predictor fails**: `crf(sqrt(conc), ...)` gives
   "Family 'beta' requires response greater than 0". `check_data()` computes the
   boundary nudge but `fit_bayesnec()` only writes it back when
   `find_transformations()` is empty, so the un-nudged response reaches `brm()`.
2. **The default `nec` prior is `gamma(5, 2/median)`**, so its mean is always
   2.5 × median — outside the data range whenever `max/median < 2.5`, which
   `sqrt` and `log` transforms routinely produce (here 84.8 against a max of
   61). It fitted cleanly regardless, so this is an observation and **not** the
   cause of anything above.

### Still open from earlier, unchanged

- **AIMS permission** to redistribute `coral` is unconfirmed and blocks merge.
- **`example_fi.csv` was not shipped**, against what RF asked for — `coral`'s
  `light` covers the factor.
- **Part 3 reports a negative result**, not the fitted grouped hurdle the plan
  wanted; those data cannot support one.
- **Merge order**: #250 → `dev` → this branch, and this branch also needs #246.
  Expect an `R/amend.R` conflict where both sides are wanted.

beckyfisher:
## #257 was the untested explanation, and it accounts for most of the divergences — but not all of them

The handover note records three explanations proposed and all three refuted by
test: competing dispersion between `phi` and the chamber intercept, the offset
scale, and the mean bumping into 1. A fourth was never tested — that `brms`
declares the group-level offset unconstrained while the identity link leaves the
mean constrained, so every leapfrog step carrying `mu` out of `(0, 1)` is
rejected. That is #257, now implemented in PR #284.

**Run on this vignette's own call and its own data**, so the comparison is
against what the vignette does rather than against a simulation:
`live | trials(tot) ~ crf(conc, "nec4param") + ogl(chamber)`, `d4` as the
vignette builds it — 150 rows, 30 chambers — 2 chains, `iter = 4000`,
`warmup = 3000`, seed 228. The two arms differ in the parameterisation and in
nothing else; the additive arm keeps `adapt_delta = 0.99`.

| family | parameterisation | divergent | max R-hat | min bulk ESS | *NEC* |
|---|---|---|---|---|---|
| `beta_binomial` | **transformed** | **30 / 2000 (1.5%)** | 1.126 | 11 | 959 |
| `beta_binomial` | additive | 1900 / 2000 (95.0%) | 1.564 | 4 | 941 |
| `binomial` | **transformed** | 661 / 2000 (33.0%) | **1.050** | 33 | 1537 |
| `binomial` | additive | 1690 / 2000 (84.5%) | 2.959 | 2 | 1769 |

## What this settles

**The mechanism is confirmed.** Divergences fall 63-fold for `beta_binomial` and
by more than half for `binomial`, and `binomial`'s R-hat falls from 2.96 to 1.05.
That is the same signature the handover note records — 2000/2000 and R-hat 2.7 —
responding to a change that touches only how the deviation is applied.

## What it does not settle, and this is the part that matters for the vignette

**No arm converges.** The package screens on `rhat_cutoff = 1.01`; the best of
the four is 1.050 and the `beta_binomial` arm is 1.126 with a bulk ESS of 11.
**#284 does not give this vignette a clean grouped binomial fit**, so the choice
the handover note describes — the published family with no working grouped fit,
or `Beta` with 73% of observations nudged — is narrowed but not removed.

**The two families disagree about the *NEC*.** 959 against 1537, on the same
data and the same equation. Neither has converged, so neither number should be
read, but the disagreement is a second reason not to present a grouped fit on
this dataset as a worked example yet.

**The design is thin for what it is being asked to do.** 150 rows across 30
chambers is five observations per chamber, and `chamber` is nested within
concentration, so no chamber spans any part of the curve. A group-level term on
that structure is weakly identified whatever the parameterisation, which is
consistent with the ESS of 11 to 33 surviving the fix.

## Suggested reading

The family question is no longer the blocker; the design is. Two options that do
not depend on any further package change:

1. **Demonstrate `ogl()` on a dataset whose grouping spans the curve**, and keep
   this dataset for the ungrouped published-family fit, which the handover note
   records as clean at R-hat ~1.00 with 0 divergences.
2. **Keep this dataset and show the diagnostics honestly** — fit it, report that
   it does not converge, and use it as the worked example of a group-level term
   that the design cannot support. That is a defensible vignette section and
   arguably a more useful one, but it is a different section from the one
   currently written.

Both are RF's call. Recorded here rather than acted on.

Measured on `batch-4-group-scale` at `95466292`, R 4.6.1, brms 2.23.0.

beckyfisher:
**Branch rebuilt and body updated.** A review posted against the earlier text was
written against a 46-file diff that no longer exists, so this notes what moved.

The branch was cut fresh from `dev` rather than merged. The previous tip
`f7f3d751` was 158 commits behind and every `R/`, `tests/` and `man/` file on it
had already reached `dev` through the batch stack, so merging would have resolved
conflicts in files whose branch side was dead. The pre-rebuild tip is preserved
on the remote as `issue-6-33-grouping-vignette-pre-rebuild`.

The diff is now `vignettes/example8.Rmd.orig` and 10 lines of
`vignettes/bayesnec.bib`. Anything anchored to a file or line outside those two
no longer has a target. GitHub reports the PR as mergeable, where it previously
reported a conflict.

Two changes of substance since the last review:

- A new section, "Group-level term syntax", is added before any model is fitted.
  It generates the `brms` formula for `ogl()`, `(par | group)` and `pgl()` with
  `make_brmsformula()`, gives a `gaussian()` contrast, and maps design structure
  to term choice.
- The `coral` dataset is no longer on the branch. Redistribution permission is
  unconfirmed and the dataset allocation for the case studies is still open, so
  it was not carried across. The case-study sections are unchanged and still
  refer to it; they are rewritten once that decision is made.

Every stated dependency in the old body is now on `dev`, so none of it remains
outstanding. The vignette is not precompiled; that is a separate step under #190
and #220.

https://claude.ai/code/session_01JnAiNVJnKrdpa1PwQP2kCm

beckyfisher:
**Dataset decision, 2026-09-09.**

The vignette is rebuilt on the acute copper and zinc data from **Luter et al.
(2025)**, *A bioluminescent bacterial toxicity assay for tropical marine
environments*, Environmental Toxicology, doi 10.1002/tox.70003. **RF has
confirmed that shipping this data in the CRAN package is agreed.** The analysis
code is public at `open-AIMS/Lum31-tox-assays` and RF is an author.

It is the only dataset screened that supplies all three structures the vignette
teaches: four replicate wells per concentration per plate for `ogl()`, 16 copper
and 17 zinc plates each spanning the whole series for `pgl()` and
`(par | plate)`, and toxicant crossed with exposure time for `bnec_group()`.

## Why this dataset rather than a structurally adequate one

Its published analysis divides each plate by the median of four control wells and
then by the plate maximum. **Ritz, Gerhard and Streibig (2026)** show that
dividing by an estimated control mean induces correlation between all
observations sharing the denominator, and that ignoring it biases effect
concentrations downwards and gives intervals that are too narrow. Their
simulation gives 2.6-6.8% ED10 bias and 0.88-0.91 coverage for that approach
against 0.7-2.1% and 0.94-0.96 for modelling the raw response. They identify
per-curve normalisation on few control measurements as the worst configuration,
which is what a four-well control median is.

Under `Gamma`, `(top | plate)` generates `bnectop ~ top * exp(topgl)`, a
per-plate multiplicative factor. The plate reader uses auto-scale gain, which is
also a per-plate multiplicative factor. So the group-level term targets what the
division targets, while estimating it with uncertainty rather than conditioning
on a point estimate. The grouping machinery is the remedy for a problem users
already have.

## Other case studies retained

- **`herbicide` with `bnec_group()`** reproduces `fig:fullbayesmanecplot` of the
  JSS paper in a single call, where the paper assembled seven. All seven levels
  converge: 0 divergent transitions of 800, R-hat 1.005-1.018, minimum effective
  sample size 317-406.
- **`CoralColour` and `pamdat`** (Flores et al. 2021) for design against model
  complexity. `pamdat` with `ogl(chamber)` is a clean 54-chamber fit;
  `CoralColour` has five concentrations, where four-parameter equations give 291
  divergent transitions of 800 and three-parameter equations give none. This is
  the content of #301.

## Dropped

`coral` is not on the branch and is no longer a blocker. `nassarius`, the
Antarctic `example_dat.csv` (origin cannot be established), `sea_urchin` (only
one of four tests is published) and the 2023 Lum-31 extract are all out.

Provenance and redistribution status for every dataset screened is recorded in
`notes/dataset_provenance.md`. The plan is in
`notes/tasks/228-grouping-vignette-{human,claude}.md`.

Two questions are outstanding with Heidi Luter and block nothing: whether the
per-plate blank readings survive, which would give a principled censoring bound
in place of the smallest resolved reading, and whether the plate reader gain was
logged, which would allow the plate variance to be split into instrumental and
biological parts.

https://claude.ai/code/session_01JnAiNVJnKrdpa1PwQP2kCm

beckyfisher:
## Status, 2026-09-11

The branch now has the dataset, a drafted vignette and the exploratory fits that
establish what each section can claim. **None of the fitted numbers ship**:
`precompile.R` re-runs every chunk, and the fits behind this draft predate three
changes that are now on `dev` or pending. What is settled is the structure and
the claims; what is not is every estimate.

## What is on the branch

`lum31`, the acute copper and zinc Lum-31 assays of Luter et al. (2025), built
from the AIMS repository workbooks: 2904 rows, 33 plates, five batches, with
`conc_group` (the four wells at one concentration on one plate), `well`,
`censoring` and `rlu_cens`. Manual page, `NEWS.md` entry, and `data-raw/lum31.R`
asserting the design invariants.

`vignettes/example8.Rmd.orig` with sections 2 and 3 written in full, sections 4,
5 and 7 structured with working chunks, and four `DRAFT NOTE` comments marking
what must not be written yet and why. The coral-based sections are gone.

`notes/scripts/example8/` with the ten scripts that produced the measurements,
and `notes/tasks/lum31-normalisation-paper-{human,claude}.md`, a plan for the
methods paper this work has turned out to support.

## What the fits established

Three results the vignette is written around, all measured on the zinc
15-minute arm, 748 rows over 17 plates, `bnec` at `dev` `eebccdb3` + #304,
rstan, `crf(log(conc), "all")` retaining 15 of 23 equations, 4 chains,
`iter = 8000`, `adapt_delta = 0.99`, seed 228.

**Normalisation is unavailable here, not merely inadvisable.** Dividing by the
largest reading in the dataset leaves plate controls between 0.171 and 0.898 of
scale, because gain is set per read. The property that makes a normalised
analysis work needs a per-curve divisor, which is what Ritz et al. (2026) object
to.

**A low ECx tracks model-set composition; EC50 does not.** Among the equations
holding weight, EC50 spans 17 per cent and EC10 spans a factor of four. Each
equation's own EC10 is stable across plate structures to within 10--30 per cent,
while the model average moves from 0.662 to 0.287 mg/L.

**The constant-dispersion Gamma overstates the variance at the control about
eight-fold.** Observed within-cell CV runs 0.051 at the control to 0.362 at the
bottom; the fitted constant is 0.43. `disp("power")` brings the implied control
CV to 0.079 and narrows the EC10 interval from 0.273 to 0.065.

## What it is blocked by

**Four decisions, none of which I can take.**

1. **Section 6's datasets.** The plan specifies `CoralColour` and `pamdat`, but
   both are in `ignore/` and are not package data, so a shipped vignette cannot
   load them. Ship one, substitute a demonstration on shipped data, or transcribe
   with a provenance note.
2. **Whether the case study adopts `disp("power")`.** The evidence says yes and
   the objection I had to it is gone --- `ogl(conc_group)` strengthens rather than
   collapses under it. It changes section 3 from "Gamma" to "Gamma with a
   dispersion sub-model" and means every arm is refitted.
3. **Section 7's `lum31`-by-toxicant structure.** It is the last use of the case
   study and should use whatever sections 3 to 5 settle. Fitted on pooled plates
   it would reproduce the bias section 5 measures.
4. **Where the methods paper lives**, if it proceeds: a compendium repository,
   named and with authorship set.

**And three changes that make the current numbers provisional.** #304 changed the
predictor prior, though not on the logged branch these fits use. #308 moved
precompilation to the HPC with the `cmdstanr` backend, which changes the sampler.
A dispersion sub-model would change the third. The re-run should be one
deliberate set of runs with all three in place, not an accumulation.

<details>
<summary>What a fresh session needs</summary>

**Read in this order.** `notes/tasks/228-grouping-vignette-claude.md` §2 for the
dataset and §4 for the task specifications; §5 for what was rejected and why, so
it is not re-proposed. Then `prompts/grouping-vignette-dataset.md`, which holds
every measurement with the settings that produced it --- it is git-ignored, so it
exists only in the checkout that ran the fits. Then the four `DRAFT NOTE`
comments in `vignettes/example8.Rmd.orig`.

**Fits are at `cache/section-fits/`** (git-ignored, about 5 GB): the four section
5 plate structures, both section 4 arms, and the herbicide `bnec_group()`.

**Settings that are not obvious.** `adapt_delta = 0.99` is needed: at the default,
`ecxll5` gave 880 divergent transitions of 16000 while holding a fifth of the
weight, and fitted alone it gives 0 at 0.99. `ecxhormebc5` fails to initialise
under the default prior on the logged branch and samples cleanly under a width
matched to the tested range --- fitted evidence bearing on #302, which was settled
on priors alone. The vignette's own chunks use a reduced equation set and 2 chains
at 2000 iterations; the reason is memory, not wall clock, and it is recorded on
#190.

**Two things measured here that belong to other issues**, already commented
there: the sequential model loop costs 206 minutes for fourteen equations where
fourteen workers would take about 25 (#184), and `bnec_group()` over seven levels
at ordinary draws extrapolates to 34 GB against the HPC job's 32 (#190).

**Tasks 6, 7 and 8 are untouched**: the style pass over the older sections, the
HPC precompile, and closing #6 and #33.

</details>

beckyfisher:
Body updated to describe the current change rather than the state at branch rebuild. What moved: the diff is no longer two files --- it now adds three datasets (`lum31`, `coral_colour`, `coral_pam`) with their manual pages, `data-raw/` scripts and `NEWS.md` entries, the exploratory fit scripts under `notes/scripts/example8/`, and a raised precompile walltime in `hpc/run.precompile`. The Evidence section is new and states the three findings the sections are written around.

Precompile job 896331 is running on the HPC with the full retained set at `iter = 5000, warmup = 4000`. The section conclusions and the provenance section are written from its output when it lands.

beckyfisher:
[claude]

The 189 fits in this vignette no longer have to be rendered in sequence. Do not
launch another full `example8` precompile expecting it to finish; the one
currently running has been going for eight hours.

**What now exists.** A companion compendium,
[open-AIMS/grouping-structures](https://github.com/open-AIMS/grouping-structures),
runs every one of those 189 fits as its own SLURM array task and writes the nine
assembled objects to a store. bayesnec PR #357 adds
`vignettes/fit_store.R` and a hook in `vignettes/precompile.R` that answers
`bnec()` and `bnec_group()` from that store. With `BAYESNEC_FIT_STORE` unset,
nothing changes, so neither CI nor a local run is affected.

**What this means for editing the vignette.** Prose, headings, figures, tables
and anything that reads a fit can be changed freely and re-rendered in minutes.
The fit calls themselves are the exception: a fit is keyed by a hash of the call
and a digest of its data, so changing `iter`, `seed`, `adapt_delta`, the family,
the formula or the data subset gives that call a new key and its units have to be
refitted. The render stops at that chunk and names the key it wanted rather than
loading the previous draft's fit or silently sampling for a day.

**Refitting after a change.** The array is idempotent, so only the units whose
keys changed are refitted:

```sh
cd grouping-structures
# edit hpc/local.conf: BAYESNEC_REF=<the branch>
./hpc/deploy.sh
```

`hpc/run.install` rebuilds the manifest against the deployed vignette and runs
`analysis/check_keys.R` as a gate before anything samples, so a mismatch is
caught in seconds.

**One thing to know about the numbers.** An assembled model set is not
bit-identical to one from a single `bnec()` call. Each equation is --- the
`seed` reaches `brm()` either way --- but `expand_manec()` draws from the
session's RNG stream to build the averaged prediction grid, which is the same
caveat `bnec()` already documents for its own parallel model loop. Any figure
quoted in the vignette should be read off the store's fits rather than from an
earlier sequential render.

The first array run was submitted against `issue-6-33-grouping-vignette-v2` at
`5f98d8d6`. If the fit calls have moved since, re-run `deploy.sh` and only the
changed ones will be fitted again.

beckyfisher:
[claude]

The store is built and the vignette renders from it. Numbers, then what to do
with them.

**Measured on the AIMS HPC overnight**, against `issue-6-33-grouping-vignette-v2`
at `5f98d8d6`:

| | |
|---|---|
| fitting the 189 units holds | 22.2 hours |
| the array took | 52 minutes (40 tasks resident, 4 cores each) |
| median unit | 1.3 min; slowest 36.2 min |
| rendering `example8` from the store | 38.5 minutes, `Precompiled without error`, nine figures |
| store | 484 MB on scratch |

The 38.5 minutes samples nothing. It is `ecx()` and `nsec()` extraction across the
14- and 18-equation sets, which is the cost #306 measured on `example7` and a
property of the vignette rather than of this arrangement. If it wants to come
down further, that is the thing to look at, not the fitting.

**Where things are.**

- store: `/export/scratch/rfisher/grouping-structures/store` on the HPC
- rendered output from that store: `C:/Rworking/renders/example8-store-render/`
- compendium: [open-AIMS/grouping-structures](https://github.com/open-AIMS/grouping-structures)
- the `bayesnec` half: #357, all five checks passing

**Do not commit the render in that folder to this branch as it stands.** It was
produced from a tree merging this branch with #357, so its R code is ahead of
this branch. Once #357 is in `dev` and `dev` is merged here, re-render and commit
that.

**Three defects were found and fixed while building this**, and each would have
produced quietly wrong output rather than an error. They are worth knowing about
because they are properties of `bayesnec`, not of the compendium.

1. `c.bnecfit()` refuses a set that mixes a hormesis equation with a non-hormesis
   one, although the data are identical --- `brms` orders `fit$data` columns by
   the model formula and `check_data_equality()` compares with `identical()`.
   Filed as #361. The compendium combines through `expand_manec()` instead, which
   is what `bnec()` itself does, verified to give the same weights and EC10 to
   every digit on a set `c()` could handle.
2. A bare combination carries neither `attach_failed_models()` nor
   `attach_bnec_record()`, so an assembled set whose call said `"all"` reported
   `requested = "ecxll3"`. Both are re-attached.
3. `ecxhormebc5` fails to fit on all four `lum31` sets and on both `fits_tox`
   levels --- six of the 189 units, and the initial-value problem of #344. Those
   sets are averaged over the equations that did fit and record the failure,
   which is what `bnec()` does. `failed_models()` on each of them names it.

**One thing to check before quoting any figure.** An assembled set is not
bit-identical to one from a single `bnec()` call. Each equation is, but
`expand_manec()` draws from the session RNG for the averaged prediction grid.
Every model-averaged number in the vignette should be read off this render rather
than carried over from an earlier sequential one.

beckyfisher:
## State, 2026-09-16

The prose is complete: every section written, no draft notes left. What remains
is verification against a render, and review.

**Render 911009 is running against the #357 fit store**
(`/export/scratch/rfisher/grouping-structures/store`, 484 MB, 9 keys). It samples
nothing, so it should finish in well under an hour rather than the 22 hours the
last production run took. If a fit call has drifted from its key the render stops
in under a minute naming the key, which is the designed behaviour.

`issue-228-fit-store` is merged into this branch. Two conflicts were resolved
rather than taken from a side: the version is 2.1.3.38, past both, and the
`sbatch` line now carries one `--export` list holding `BAYESNEC_FIT_STORE` and
`BAYESNEC_VIGNETTE_WORKERS` while keeping the `--cpus-per-task` and `--mem`
overrides `--workers` sets.

<details>
<summary>What is left, in order</summary>

1. **Verify every number in the prose.** A block at the top of
   `vignettes/example8.Rmd.orig` lists each one and which run it came from.
   Sections 5 and 7 quote job 902840, which predates the 221-commit merge of
   `dev`; sections 4 and 6 quote the shape-check render 910761 at 2 chains and
   400 iterations. None of them has been checked against a production run made
   since the merge.
2. **Run the `prose-check` skill** on the finished document, checking the added
   sections in isolation as well as the whole file.
3. **Grep the render for developer details** before committing it --- cluster
   paths, a home directory, compiler output.
4. **Commit the rendered `.Rmd` and its nine figures**, then confirm CI.
5. **Review**, which is the one thing outstanding that cannot be done from here.

Four issues came out of this work and none blocks the merge: #319 `crf()`
resolving its model argument in the global environment only; #329, whose residual
#355 identifies as the family's closures retaining the knitr environment; #338
the sequential level loop; #344 `ecxhormebc5` failing to initialise.

</details>

beckyfisher:
`vignettes/example4.Rmd.orig:40` states that `bnec` does not allow an
interaction with a fixed factor and recommends independent fits per factor level
as the route to take instead. That is accurate on `dev` and becomes wrong when
this pull request merges. #364 holds the detail and the proposed wording; the
edit belongs in this pull request rather than after it, so that the text and the
capability change together.

beckyfisher:
[claude]

PR #366 adds an `articles:` block to `_pkgdown.yml` and this branch needs one
line in it, or the pkgdown job on this pull request fails once #366 merges.

Branch: `issue-6-33-grouping-vignette`. File: `_pkgdown.yml`. Add `example8`
under `Designs and response types`, after `example7`:

```yaml
  - title: Designs and response types
    navbar: Designs and response types
    contents:
      - example6
      - example7
      - example8
```

`pkgdown:::data_articles_index()` aborts on a vignette that is present in the
build but absent from the block, with

```
In _pkgdown.yml, 1 vignette missing from index: "example8".
```

Do not remove the `navbar:` keys. Without them pkgdown replaces the Articles
dropdown with a plain link to the index, and `pkgdown::check_pkgdown()` reports
no problem, so the loss reaches the published site unremarked. #366 records the
measurement.

Definition of done: `pkgdown::check_pkgdown()` reports "No problems found" on
this branch merged with `dev`, and the pkgdown job passes on this pull request.

`notes/vignette_numbering.md` rule 5 on #366 records the same obligation for
every vignette after this one.

beckyfisher:
The body has been updated: the vignette is now on a log concentration axis
throughout, the two coral endpoints are fitted on `crf(log(diuron_adj), ...)`,
and each dataset opens with a raw-data figure. The sections that changed are
*What*, *Evidence* (a new paragraph on the coral control) and, inside the fold,
*Settings* and *Known limitations*. The model count there is 180 rather than 189.

Rendered at `02ce807f` against a store rebuilt at `bayesnec` 2.1.3.38: 180 units
fitted as a cluster array, assembled, and read by the render, which samples
nothing and takes 33 minutes.

Two things a reviewer arriving from an earlier read should know.

Sections 4, 5 and 7 are numerically unchanged. The printed output of this render
was diffed against the one at `54c6e7da`, and every difference falls in section
6 --- estimates, model weights and `prob_diff` are identical to every digit
elsewhere. The whole set was refitted only because `assemble_store.R` compares
the version stamp and refuses a set whose units were fitted by a different
`bayesnec`.

Section 6's argument changed with the scale, and the passage follows it. On the
recorded predictor both four-parameter equations recorded divergent transitions
at an R-hat under 1.018, so the section said R-hat alone would not separate the
two sets. On `log(diuron_adj)`, `nec4param` records 212 divergences at an R-hat
of 1.3537 and fails both checks; the point is now made on `ecx4param`, which
records 326 at 1.0088.

Open questions for review are the 0.1 µg/L substituted for the coral control,
which reverses what `data-raw/coral_diuron.R` argues for and is therefore made
in the vignette rather than in the data, and the reading of `fit_pam`'s
estimates, whose N(S)EC and EC10 upper bounds sit above the EC50 median. That is
attributed to `ecxhormebc4` holding 0.64 of the weight and placing its threshold
where the curve turns upward, which is an interpretation rather than something
the output states.

beckyfisher:
Correcting an explanation in the comment above, and recording what replaced it.

That comment said `fit_pam`'s model-averaged N(S)EC has an upper bound above the
EC50 median because `ecxhormebc4` holds 0.64 of the weight and places its
threshold where the curve turns upward. Pulling each of the fifteen equations
out of the assembled fit and estimating it separately shows the opposite.

| equation | type | weight | estimate, µg/L |
|---|---|---|---|
| `ecxhormebc4` | NSEC | 0.639 | 0.226 (0.124--0.292) |
| `ecxwb1` | NSEC | 0.233 | 0.179 (0.115--0.230) |
| `nec4param` | NEC | 0.039 | 2.107 (1.815--2.372) |
| `ecxwb2` | NSEC | 0.026 | 1.615 (1.195--1.943) |
| `nec3param` | NEC | 0.007 | 1.657 (1.284--2.013) |

`ecxhormebc4` has no upper bound above 0.3 and is what holds the averaged
*median* at 0.217. The averaged upper bound of 2.112 comes from `nec4param`,
`ecxwb2` and `nec3param`, which hold seven per cent of the weight between them.
That is enough to reach the 97.5th percentile of a mixture, and the `nec4param`
estimate of 2.107 and the averaged bound of 2.112 agree to three decimal places.
The same draws make the step in the credible band at about 2 µg/L, where its
upper edge falls from roughly 0.58 to 0.31: an equation that fits a threshold
holds the curve near the control until it and drops afterwards.

The vignette now states that, and draws the general point from it --- an
equation given four per cent of the weight has no effect on the median and still
decides where the interval ends, so a model-averaged interval is read against
the weights rather than on its own.

Thresholds were taken with `nec()` for the equations that fit one and `nsec()`
for the rest, both under `xform = exp`, on the assembled `fit_pam` from the
store the vignette renders from. The estimates above are from the store at
2.1.3.38 with the vignette's earlier sampling settings, and will be re-read from
the run now in progress.

beckyfisher:
The `example4` item of #364 is on this branch as 110740fe, because the passage
it replaces is accurate on `dev` and becomes wrong when this merges.

`vignettes/example4.Rmd.orig:41` said that `bnec` does not allow an interaction
with a fixed factor and recommended independent fits per level as the only
route. It now names both routes — a group-level term in the formula, and
`bnec_group()` — says which question each answers, and links `example8`, which
was added to the vignette's link block as `[e8]`. It also states what
`compare_posterior` is for once group-level terms exist, taking the permutation
argument from the *Case study 3* section of `example8`: the function pairs draws
by a random permutation, so it is valid across separate fits and not between two
levels of one fit.

The rendered `example4.Rmd` was edited to match. No chunk changed, so no refit
was needed.

The other two items of #364 are link faults that predate this branch and are in
#369 against `dev`.

beckyfisher:
[claude]

## What

A documentation pass over the whole package, on a local branch off `dev` that
is not yet pushed, trimmed the help files and vignettes for prose style, length
and statements that go stale. Three passages about grouping stayed in `man/`
because `example8` is not on `dev` yet and this branch could not see it. They
are listed below so this PR can decide whether any of them belongs in the
vignette instead.

Nothing here is a request to change this branch before it merges. The audit
branch does not touch `example8` or any file this PR adds.

## Why it matters

`man/bnec_group.Rd` and `man/crossed_group_weights.Rd` together run to 224 lines
of Rd, and a large part of it is conceptual rather than reference: when the
levels of a factor may be fitted separately at all, what a per-level dispersion
parameter means, and the three readings of a crossed weight table. That is
vignette material by the standard the rest of this audit applied, and
`example8` is the vignette for it. Leaving it in `man/` means a user meets the
argument for `bnec_group()` in a help page and the worked example in a
vignette, with neither pointing at the other.

## Evidence

Three passages, with what each currently says.

### The basis for fitting the levels separately

In `man/bnec_group.Rd`. Factor
levels partition the data disjointly and share no parameters, so the log
likelihood is a sum over levels and the expected log predictive density is
additive; under pseudo-BMA the crossed model weights are then the outer product
of the per-level weight vectors. This is the justification for the function
existing, and it is the same identity `crossed_weights()` rests on for the two
blocks of a hurdle fit.

### Dispersion per level

In `man/bnec_group.Rd`. Separate fits give each level
its own `sigma`, `shape` or `phi`, deliberately, because a shared dispersion
parameter would break the factorisation the crossed weights depend on. A model
with structure spanning levels is a different model and is not what
`bnec_group()` fits. This is precisely the choice between `bnec_group()` and a
group-level term that the PR body says no vignette currently addresses.

### The crossed table and the pooled comparison

In `man/crossed_group_weights.Rd`, about 40 lines. The unrestricted maximum picks
the best equation per level and answers whether the functional form changes
across levels; the diagonal maximum asks which single equation best describes
every level. Passing a `pooled` fit gives a third reading, a like-for-like WAIC
comparison of grouped against pooled, with `se_diff` available only where every
level and the pooled fit settled on a single model.

<details>

### Changes the audit branch has already made here

`man/bnec_group.Rd` fell from 142 to 131 lines. The cuts were:

- the section headed *What this does not do*, whose body read "That is stage 2
  of #33 and lands in the post-processing that the `toxval` migration moves, so
  it is deliberately out of scope here". The fact that a joint refit is not
  performed was kept as a sentence under *Dispersion per level*; the project
  planning around it was not.
- "a tie going to the models, which is what earlier versions did", and the
  paragraph stating that a grouped call no longer reports the estimates earlier
  versions reported. `NEWS.md` holds both.
- "`sample.int()`'s algorithm changed once already, in R 3.6.0" became "has
  changed before". The point is that a regenerated seed cannot be relied on, and
  it does not need the version.
- the headings *Why the levels can be fitted separately*, *What this does not
  do* and *What a plan does and does not reproduce* became noun phrases, and
  "That is a feature, not an oversight" was removed.

`man/crossed_group_weights.Rd` lost the heading *Does the factor matter at
all?*, the bold lead-in on the pseudo-BMA identity, and an announcement idiom.

### If a passage moves

Each of the three is self-contained and would move as a section without
rewriting. The help file should keep a one-sentence statement of the behaviour
and a `vignette("example8")` pointer, which is the pattern the rest of this
audit used: `bnec.Rd` now points at `example1` for the normalisation argument
and `models.Rd` at `example2b` for the `drc` correspondence, rather than
restating either.

If none of them moves, that is a reasonable answer too. They are accurate where
they are, and the argument for moving them is that `example8` will say the same
things at greater length and the two will drift.

### Replying

Name the passage or passages that should move, if any. The move can be made on
the audit branch before it is pushed, or here on this one.

</details>

beckyfisher:
The body has been updated again, for the review comments and for a defect they
led to. The sections that changed are *What*, a new paragraph under *Evidence*,
and the sampling note inside the fold.

The defect is the one worth reading. `example8` reported model-averaged
estimates without ever calling `check_sampling()`. On the single-plate grouped
fit, `ecxexp` held 0.122 of the weight with 159 divergent transitions, and its
`top * exp(-beta * x)` form is unbounded as concentration falls on a logged
predictor: it put that fit's EC10 lower bound at 0.015 mg/L, below the lowest
tested concentration, and drew a credible band reaching four million relative
light units where the data top out near six hundred thousand. A reader queried
the figure, which is how it was found.

Across the five reported single-object fits the weight on equations failing the
screen was 0.170, 0.122, 0.617, 0.000 and 0.030. The 0.617 is `fit_pooled`, the
fit the across-concentration section holds up as the wrong way to do it; the
plate fit that section recommends is clean. All five are now screened, with the
cutoffs and the reasoning deferred to `vignette("example9")`, which already
records that Stan's guidance is that any divergence means the sampler failed and
that the default of ten is a package convention.

Screening changes reported estimates. `fit_pooled` EC10 goes from 0.328 to 0.280
mg/L and its N(S)EC from 0.197 to 0.165, so the across-concentration section now
reports a sixth of a difference in EC10 where it previously reported almost
none. Those numbers are a change of substance rather than of presentation.

Also in this round: the 314-line `summary(fits_herb)` block is replaced by a
heatmap of model weight by equation and herbicide, which shows what that section
asserts --- the seven do not agree on which equation describes them; the
comparison with Jones and Kerswell now sets EC50 against EC50 rather than their
EC50 against the N(S)EC here; and the two `bnec_group()` fits are noted as
unscreened because #376 blocks it.

Issue #230, opened 2026-08-22
pkgdown 'development: mode: auto' resolves master to devel, so both branches publish to /dev/


beckyfisher:
Introduced on `dev` by #220 (#215). Not yet reached `master`, so nothing is broken on the live site today — but it will be the moment the new `pkgdown.yaml` reaches `master`.

## What the design assumes

`_pkgdown.yml` now sets:

```yaml
development:
  mode: auto
```

with a comment stating that this keys off the DESCRIPTION version, so `master` (a "release version", `2.1.3.1`) lands at the site root and `dev` (a "four-component development version") lands at `/dev/`. `pkgdown.yaml` relies on that: it deploys with `clean: false` on pushes to both branches, on the stated grounds that "a build from `master` writes `docs/` at the root and leaves `docs/dev/` untouched".

## What pkgdown actually does

`pkgdown:::dev_mode_auto()` does not distinguish a four-component version by whether the fourth component looks like a development counter. Any version with four or more components is `"devel"` unless the first three are `0.0.0`:

```r
pkgdown:::dev_mode_auto("2.1.3.1")    # master  -> "devel"
pkgdown:::dev_mode_auto("2.1.3.10")   # dev     -> "devel"
pkgdown:::dev_mode_auto("2.1.4")      #         -> "release"
pkgdown:::dev_mode_auto("2.1.3.9000") #         -> "devel"
```

(checked against pkgdown 2.2.1, which is what `setup-r-dependencies` installs)

`master` currently carries `Version: 2.1.3.1` — four components — so it resolves to `devel` too. Both branches would build into `docs/dev/` and publish to the same `/dev/` prefix, each overwriting the other, while the site root is never rebuilt at all and survives only because `clean: false` leaves it alone. That is exactly the failure the `_pkgdown.yml` comment says the design avoids.

The comment's premise is also wrong on its own terms: `2.1.3.1` is not a "release version" by pkgdown's rule.

## Why it looks fine right now

`master` still has the old `pkgdown.yaml`, which triggers on `master` only and predates the `development:` block. And at the 2.1.4 release `master` will carry `2.1.4` — three components — which does resolve to `release`. So the intended behaviour happens to be restored by the release itself.

It breaks again silently the next time `master` carries a four-component version: a `2.1.4.1` patch, or any point where `master` is bumped ahead of a release. Nothing warns; the dev preview is simply overwritten by the release site and vice versa.

## Suggested fix

[...]

PR #233, opened 2026-08-22
Make the precompile workflow runnable, and declare workflow permissions (#230)

beckyfisher:
**Release tier: 2.1.4.** Phase 1 of `notes/implementation/06_review_run.md`. Branched from `dev`, independent of the 224–228 stack — touches only `.github/` and `_pkgdown.yml`.

Closes **#230**.

## The blocking defect

`precompile-vignettes.yaml` could not be run at all. `workflow_dispatch` only takes effect once the workflow file is on the repository's **default branch** (`master`); it was added by #220 to `dev` alone, so GitHub answered a dispatch with a 404 and the workflow was absent from `gh workflow list`:

```
$ gh workflow run precompile-vignettes.yaml --repo open-AIMS/bayesnec --ref dev
HTTP 404: workflow precompile-vignettes.yaml not found on the default branch
```

That matters beyond CI: #190 (the full `precompile.R` run) is a release gate, every vignette number on `dev` is stale, and this workflow is the mechanism #220 added to make that refresh possible without a local Stan toolchain.

**Resolved without touching `master`** — a `push` trigger on a `precompile/**` branch namespace, so the job can be exercised from `dev` now and becomes dispatchable in the ordinary way once it reaches `master` at release.

## The other three defects

- **No `permissions:` block on either workflow.** `peter-evans/create-pull-request` needs `contents: write` and `pull-requests: write`; `JamesIves/github-pages-deploy-action` needs `contents: write`. Both now declare them rather than depending on the repository default, which is read-only on newer repositories. (`ssdtools`, which RF pointed at as the reference setup, declares `permissions: contents: write` for the same reason.)
- **The `~/.cmdstan` cache cached nothing.** The job never installs cmdstanr and nothing in the package sets `backend = "cmdstanr"`, so `brms` uses `rstan`, which compiles in-process and writes nothing there. Removed, with a comment saying so — recompilation genuinely is paid every run, which is what `timeout-minutes` has to accommodate.
- **The `on.exit()` restore never fired.** `shell: Rscript {0}` runs at top level, where `on.exit()` registers against no frame. The held-back `.Rmd.orig` files were not being restored.

  The obvious fix — wrap in `local()` and keep a flag — does not work either, because `precompile.R` opens with `rm(list = ls())`, which sourced at top level clears the global environment and takes the flag with it. The restore is therefore keyed on `dir.exists(".precompile_held")`, which survives it. Verified end to end:

  ```
  held back: example7
  during run,     vignettes/ = example1.Rmd.orig example4.Rmd.orig
  after restore,  vignettes/ = example1.Rmd.orig example4.Rmd.orig example7.Rmd.orig
  held dir removed: TRUE
  ```

## Added: a way to test this without spending hours

The expensive failure mode is a five-hour fitting run that dies at the final step on a permissions problem. So:

- **`dry_run`** (input, or the branch `precompile/dry-run`) skips the fitting entirely and writes a marker file, exercising checkout, dependency install, `permissions:`, and `create-pull-request` — including the repository's *Allow GitHub Actions to create and approve pull requests* setting, which cannot be verified from a config file.
- **Branch-suffix selection**, since a push carries no inputs: `precompile/example1` rebuilds one vignette, `precompile/example1,example4` two, `precompile/all` everything.

Resolution simulated across all seven cases:

```
precompile/example1          -> selection=example1           dry=FALSE
precompile/example1,example4 -> selection=example1,example4  dry=FALSE
precompile/all               -> selection=<all>              dry=FALSE
precompile/dry-run           -> selection=<all>              dry=TRUE
dev (input example2b)        -> selection=example2b          dry=FALSE
dev (no input)               -> selection=<all>              dry=FALSE
dev (input dry_run=true)     -> selection=<all>              dry=TRUE
```

## `_pkgdown.yml` — comment corrected, no behaviour change

The comment stated a rule `pkgdown` does not implement. `pkgdown:::dev_mode_auto()` returns `"devel"` for **any** version with four or more components whose first three are not `0.0.0` — the size of the fourth is irrelevant:

```r
dev_mode_auto("2.1.3.1")   # master today -> "devel"
dev_mode_auto("2.1.3.12")  # dev today    -> "devel"
dev_mode_auto("2.1.4")     #              -> "release"
```

So `master`, at `2.1.3.1`, currently resolves to devel too and would publish into `/dev/` rather than the root. `ssdtools` relies on the same rule deliberately — `main` carries `2.6.0.9002` and publishes to `/dev/`, with the root published on release at a three-component version.

**This is a versioning convention, not a config setting, so it is documented rather than fixed here.** Cutting 2.1.4 as `2.1.4` makes the two sites separate correctly; cutting a future release as `2.1.4.1` would silently collide them again. **A call for RF at release time.**

## Verification

YAML parses; the trigger-resolution logic and the hold/restore logic are both simulated above. The workflow itself is exercised by pushing `precompile/dry-run` — result reported in a comment on this PR.

## Not done here

No R code is touched, so there is nothing to test with `devtools::test()`. `vignettes/precompile.R` itself is unchanged.

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -7,6 +7,13 @@ on:
 
 name: pkgdown
 
+# github-pages-deploy-action pushes to gh-pages, which needs write access to
+# repository contents. Relying on the repository's default workflow permissions
+# leaves the deploy dependent on a setting that is read-only by default on
+# newer repositories. See #230.
+permissions:
+  contents: write
+
 jobs:
   pkgdown:
     runs-on: ubuntu-latest
--- a/.github/workflows/precompile-vignettes.yaml
+++ b/.github/workflows/precompile-vignettes.yaml
@@ -10,6 +10,13 @@
 # PR with the regenerated `.Rmd` files and figures. It deliberately does not
 # push to the branch directly: precompilation changes every number in every
 # vignette, and that is a diff someone should look at.
+#
+# `workflow_dispatch` only works once this file is on the repository's DEFAULT
+# branch (`master` here). Until then GitHub answers a dispatch with
+# "workflow not found on the default branch" and the workflow is invisible to
+# `gh workflow list`. The `push` trigger on the `precompile/*` namespace exists
+# so the job can be exercised from `dev` before it ever reaches `master` --
+# push a branch named e.g. `precompile/test` and it runs. See #230.
 on:
   workflow_dispatch:
     inputs:
@@ -19,6 +26,29 @@ on:
           Leave blank to rebuild all of them -- slow, and what #190 wants.
         required: false
         default: ''
+      dry_run:
+        description: >-
+          Skip the fitting entirely and only exercise the plumbing -- checkout,
+          dependency install, and the pull-request step. Use this to verify the
+          workflow before committing hours of compute to it.
+        type: boolean
+        required: false
+        default: false
+  push:
+    branches: ['precompile/**']
+    # The branch suffix selects what to rebuild, since a push carries no inputs:
+    #   precompile/example1        -> rebuild example1 only
+    #   precompile/example1,ex4    -> rebuild both
+    #   precompile/all             -> rebuild everything (what #190 wants)
+    #   precompile/dry-run         -> skip fitting entirely; see PRECOMPILE_DRY_RUN
+
+# create-pull-request needs both of these, and the repository setting
+# "Allow GitHub Actions to create and approve pull requests" must be on.
+# Without them the final step fails after the whole fitting run has completed,
+# which is the most expensive possible place to discover a permissions problem.
+permissions:
+  contents: write
+  pull-requests: write
 
 name: precompile vignettes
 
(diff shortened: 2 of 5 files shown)


beckyfisher:
## Verified end to end

Pushed `precompile/dry-run` carrying this branch's workflow. **Every step succeeded**, including the one that could only fail after a full fitting run under the old configuration:

```
success  Set up job
success  Run actions/checkout@v4
success  Run r-lib/actions/setup-pandoc@v2
success  Run r-lib/actions/setup-r@v2
success  Run r-lib/actions/setup-r-dependencies@v2
success  Precompile                              <- dry run; precompile.R not sourced
success  Open a PR with the rebuilt vignettes    <- created PR #234
success  Complete job
```

Run: [32579445161](https://github.com/open-AIMS/bayesnec/actions/runs/32579445161).

What that establishes, none of which could be read off a config file:

- the `precompile/**` push trigger fires, so the workflow is runnable from `dev` without waiting for it to reach `master`;
- `permissions: contents: write` + `pull-requests: write` are sufficient;
- the repository's *Allow GitHub Actions to create and approve pull requests* setting **is** enabled — this was the open risk, and it is not a problem;
- `peter-evans/create-pull-request@v6` works against this repo, and its branch naming resolves as `precompile/<ref_name>`;
- `setup-r-dependencies` resolves the package's dependency tree on a clean runner.

The throwaway artifacts have been cleaned up: PR #234 closed, and both `precompile/dry-run` and `precompile/precompile/dry-run` deleted.

**Still unverified, and deliberately so:** the fitting run itself. That is #190's, takes hours, and should be run once, immediately before the CRAN submission. What this PR establishes is that when it is run, it will not fail at the last step.

Issue #232, opened 2026-08-22
define_prior()'s zero guard is a step function on a continuous collapse; worst case is just below the threshold


beckyfisher:
Follow-up to #210 (merged as #222), and referred to from #229.

#210 fixed the case where a quantile of the response is **exactly** zero: `positive_scale()` detects that and falls back to the same quantile of the positive part. But the collapse it is guarding against is continuous, not a step. A quantile that is merely *small* produces an essentially collapsed prior with no guard at all, and the worst case sits just below the threshold where the guard would fire.

## Demonstration

A `nec4param`-shaped `zero_inflated_poisson` response with a true `top` of 40 and a true `bot` of 5, at increasing zero fractions. `gamma(2, rate)` has mean `2 / rate`, so the prior means are directly comparable to the truth:

```r
x  <- as.numeric(rep(1:10, each = 15))
mu <- 40 * exp(-0.35 * pmax(x - 2, 0)) + 5          # true top 40, true bot 5
zi <- function(p) { set.seed(1); as.numeric(rpois(length(x), mu) * rbinom(length(x), 1, p)) }

for (p in c(1, .8, .5, .4, .35, .3, .26, .25, .2, .1)) {
  y  <- zi(p)
  pr <- define_prior("nec4param", validate_family("zero_inflated_poisson"), x, y)
  # ... extract the gamma rate for top and bot
}
```

```
zero_frac q75(raw)  top_prior_mean bot_prior_mean
     0.00      33.5          33.50          10.03
     0.24      27.8          27.75           4.03
     0.54      14.8          14.75           9.03
     0.60      11.0          11.00           9.03
     0.61      11.0          11.00           9.03
     0.67       9.0           9.00          10.03
     0.71       7.0           7.00          10.03
     0.72       6.8           6.75          10.03   <- worst case, guard does not fire
     0.76       0.0          29.00          10.03   <- guard fires, prior recovers
     0.87       0.0          28.25          11.53
```

The `top` prior mean degrades monotonically from 33.5 down to 6.8 as the zero fraction climbs toward 75%, then jumps back to 29.0 the instant `quantile(response, 0.75)` reaches exactly zero. At 72% zeros the prior is centred at roughly one sixth of the true upper asymptote, and nothing in the current code notices.

## Why the tests do not catch it

[...]

Issue #229, opened 2026-08-22
define_prior() errors on responses with no positive values, for families that never use the guarded quantiles


beckyfisher:
Introduced on `dev` by #222 (`positive_scale()`, #210) and made unavoidable by #223 (#207). Both are in the 2.1.4 tier, so this has not reached a release.

## What happens

`define_prior()` now errors for any response with no positive values, **including families whose priors never use the guarded quantities**.

`u_t_g` and `u_b_g` are built unconditionally at `R/define_prior.R:119-125` and `R/define_prior.R:147-157`, but they are only ever consumed for `Gamma`, `poisson` and `negbinomial`. The gaussian, beta, binomial, bernoulli and beta_binomial entries in `y_t_prs` / `y_b_prs` are built from `quantile()` and `sd()` directly and are perfectly well defined on a response that is entirely negative. `positive_scale()` nonetheless runs and calls `stop()`.

An all-negative response is ordinary gaussian input — log-transformed proportions, log ratios, growth increments, anything expressed as a change.

```r
set.seed(1)
x <- as.numeric(rep(1:10, each = 5))
y <- -rexp(50, 1) - 0.5          # gaussian, entirely negative

define_prior("nec4param", validate_family("gaussian"), x, y)
```

Before #222 (at 9e1ef7bc) this returned valid priors, with two harmless `min()` warnings:

```
                                        prior class nlpar
                                 normal(0, 5)     b  beta
 normal(-0.646912064123937, 2.22941542670029)     b   top
  normal(-2.38921489539536, 2.22941542670029)     b   bot
                  gamma(5, 0.363636363636364)     b   nec
```

After #222 (at 20524329):

```
Error : Cannot build priors for "top" and "bot": the response contains no
positive values, so there is no scale to place them on. Check the response
variable, and see ?bnec for the families bayesnec supports.
```

## Why there is no workaround

`add_brm_defaults()` now builds the defaults unconditionally at `R/helpers.R:774`, *before* `validate_priors()` is consulted, so the error fires even when the user has supplied a complete and valid prior set of their own:

```r
gp <- brms::prior_string("normal(0,5)", nlpar = "beta") +
      brms::prior_string("normal(-0.5,2)", nlpar = "top") +
      brms::prior_string("normal(-2,2)", nlpar = "bot") +
      brms::prior_string("gamma(5,0.36)", nlpar = "nec", lb = 1, ub = 10)

[...]

PR #236, opened 2026-08-22
Take the top/bot prior scale at a quantile rescaled by the zero fraction (#232)

beckyfisher:
**Release tier: 2.1.4.** Phase 1 of `notes/implementation/06_review_run.md`. **Stacked on #235** (#229), because it rewrites the same function.

Closes **#232**. Approach chosen by RF, 2026-08-22.

## The defect

#210's guard tested for an *exactly zero* quantile and substituted the same quantile of the positive part. But a quantile of a zero-inflated response is a quantile of the **mixture**, and it is pulled towards zero throughout its range — reaching exactly zero only at the extreme. So the guard was a step function applied to a continuous problem, and it left the worst case, immediately below the threshold, entirely unguarded.

Same setup as the issue: `nec4param`, `zero_inflated_poisson`, true `top` = 40, prior mean = `2 / rate`.

| zero frac | 0.00 | 0.24 | 0.54 | 0.60 | 0.67 | 0.71 | **0.72** | **0.76** | 0.87 |
|---|---|---|---|---|---|---|---|---|---|
| raw q75 | 33.5 | 27.8 | 14.8 | 11.0 | 9.0 | 7.0 | **6.8** | **0.0** | 0.0 |
| **before** | 33.5 | 27.8 | 14.8 | 11.0 | 9.0 | 7.0 | **6.8** | **29.0** | 28.3 |
| **after** | 33.5 | 35.0 | 28.9 | 29.1 | 28.8 | 29.0 | **29.0** | **29.0** | 28.0 |

At 72% zeros the prior was centred at roughly a sixth of the true upper asymptote, then recovered discontinuously by a factor of four one step later.

## The fix

Level `p` of the distribution conditional on being positive sits at level `z + (1 - z)p` of the mixture — a share `z` of the mass has to be passed before any positive value is reached. Equivalently `1 - (1 - p)(1 - z)`. That is the level the quantile is now taken at.

Two properties that matter:

- **It reduces to the identity when `z` is zero**, so a response with no zeros is untouched. Verified with `identical()` at `probs` of 0, 0.25, 0.5, 0.75 and 1 — not `expect_equal`, since the claim is exactness.
- **It degrades smoothly.** No adjacent pair in the sweep above differs by more than a factor of two; the old threshold crossing was a factor of four in one step.

### A correction to the formula as proposed

The issue's comment proposed `1 - (1 - probs) / (1 - zero_frac)`. That inverts an operator: it moves the level the *wrong way*, and goes negative past 75% zeros — the regime this exists for. On a vector that is 50% zeros with positives `1:50`, where the target `quantile(positives, 0.75)` is 37.75:

| | rescaled level | quantile recovered |
|---|---|---|
| `1-(1-p)/(1-z)` | 0.500 | **0.5** |
| `1-(1-p)*(1-z)` | 0.875 | **37.6** |

Recorded on the issue at the time. Correcting algebra to match a stated intent is implementation rather than a modelling decision, so it was self-resolved under the phase rule — but it is a change from what was written, so it is called out here for override.

## ⚠️ This changes results for existing fits

Flagged in its own `NEWS.md` sentence rather than folded into the entry. **`poisson` and `negbinomial` are long-released families** — `origin/master:R/define_prior.R:40-52` — and any count response with a material share of zeros reaches this code. It is *not* confined to the zero-inflated families added in 2.1.4.

A response containing **no** zeros is unaffected exactly, so the blast radius is "count fits with zeros", not "all count fits".

## Tests

Four added; `test-define_prior.R` passes **117/117**, `test-get_priors.R` 38, `test-helpers.R` 37.

Deliberately of a different kind from #210's, which assert only that the rate is finite and positive — **those pass just as well on a prior centred at a sixth of the truth**, which is why this defect survived them. The new ones constrain the prior to be on the scale of the data across the whole zero-fraction sweep, pin the algebra directly against the conditional quantile, and assert smoothness as a bound on adjacent ratios.

## Note on `zero_frac`

Computed as `sum(finite <= 0) / length(finite)`, not `== 0`. Only `Gamma`, `poisson` and `negbinomial` reach this function after #229, and all three are non-negative by construction, so the two are equivalent in practice — `<= 0` is defensive, since a negative value in a count response is invalid input rather than a scale worth trusting.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: bayesnec
 Title: A Bayesian No-Effect- Concentration (NEC) Algorithm
-Version: 2.1.3.12
+Version: 2.1.3.13
 Authors@R: c(person("Rebecca", "Fisher", email = "r.fisher@aims.gov.au", role = c("aut", "cre"),comment = c(ORCID = "0000-0001-5148-6731")), person("Diego R.","Barneche",role="aut",comment = c(ORCID = "0000-0002-4568-2362")), person("Gerard F.","Ricardo",role="aut",comment = c(ORCID = "0000-0002-2220-3971")), person("David R.","Fox",role="aut",comment = c(ORCID = "0000-0002-3178-7243")))
 Description: Implementation of No-Effect-Concentration estimation that uses 'brms' (see Burkner (2017)<doi:10.18637/jss.v080.i01>; Burkner (2018)<doi:10.32614/RJ-2018-017>; Carpenter 'et al.' (2017)<doi:10.18637/jss.v076.i01> to fit concentration(dose)-response data using Bayesian methods for the purpose of estimating 'ECx' values, but more particularly 'NEC' (see Fox (2010)<doi:10.1016/j.ecoenv.2009.09.012>), 'NSEC' (see Fisher and Fox (2023)<doi:10.1002/etc.5610>), and 'N(S)EC (see Fisher et al. 2023<doi:10.1002/ieam.4809>). A full description of this package can be found in Fisher 'et al.' (2024)<doi:10.18637/jss.v110.i05>. This package expands and supersedes an original version implemented in 'R2jags' (see Su and Yajima (2020)<https://CRAN.R-project.org/package=R2jags>; Fisher et al. (2020)<doi:10.5281/ZENODO.3966864>).
 Depends:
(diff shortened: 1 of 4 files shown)


beckyfisher:
This is al implemented as intended.

beckyfisher (commented):
Looks fine.

Issue #231, opened 2026-08-22
get_priors() does not report or round trip a zi prior for the zero-inflated count families


beckyfisher:
Follow-up to #207 (merged as #223). `usable_prior()` was widened to keep the family's own dispersion classes so a supplied prior round trips, and the NEWS entry says the consequence was that "a fit's prior set was a complete record of everything except dispersion". For `zero_inflated_poisson` and `zero_inflated_negbinomial` — added in the same 2.1.4 tier, under #104 — it still is not.

## The rationale in `dispersion_classes()` does not hold for the ZI count families

`R/get_priors.R` documents the omission of `zi` and `hu` like this:

> `zi` and `hu` are deliberately absent: bayesnec models those as a second parameter block, so a prior on them carries class "b" with a prefixed `nlpar`, not a class of their own.

That is true for `hurdle_gamma` and `zero_inflated_beta`, which go through `bnec_hurdle()` / `hurdle_mu_family()` and get a genuine second curve. It is not true for `zero_inflated_poisson` / `zero_inflated_negbinomial`, which are single-block fits — `check_models.R:155-156` admits them without a second block, and D4 records the decision that they must not get one.

For those two families `zi` is an ordinary brms parameter with a class of its own, sitting on the brms default:

```r
fam <- validate_family("zero_inflated_poisson")
bf  <- brms::bf(y ~ top * exp(-exp(beta) * (x - nec) * step(x - nec)),
                top + beta + nec ~ 1, nl = TRUE)
brms::get_prior(bf, data = d, family = fam)[, c("prior", "class", "nlpar", "source")]
#>        prior class nlpar  source
#> 1 beta(1, 1)    zi        default
#> 2                b  beta default
#> ...
```

## The gap

`get_priors()` reports nothing for it, so the returned set is silently incomplete for exactly the families where the zero-inflation probability is the parameter of interest:

```r
get_priors(y ~ crf(x, "nec4param"), data = d, family = "zero_inflated_poisson")
#>                        prior class nlpar
#>                 normal(0, 5)     b  beta
#>  gamma(2, 0.0919540229885057)     b   top
#>   gamma(2, 0.116890707188778)     b   bot
#>   gamma(5, 0.363636363636364)     b   nec
```

and `usable_prior()` drops a supplied one, so it does not round trip — asymmetric with `shape`, which now does:

```r
usable_prior(gp + brms::prior_string("beta(2,5)",   class = "zi"))$class
#> "b" "b" "b" "b"
[...]

PR #237, opened 2026-08-22
Report and round trip a zi prior for the zero-inflated count families (#231)

beckyfisher:
**Release tier: 2.1.4.** Phase 1 of `notes/implementation/06_review_run.md`. Branched from `dev`, independent of #233, #235 and the 224–228 stack — touches only `R/get_priors.R` and a comment in `R/inits_functions.R`.

Closes **#231** (the implementation half — see *Left open*, below). Found by the post-merge review of #223 (#207).

## The gap

#207 widened `usable_prior()` to keep the family's dispersion classes so a supplied prior round trips, and its `NEWS.md` entry says the result was that "a fit's prior set was a complete record of everything except dispersion". For `zero_inflated_poisson` and `zero_inflated_negbinomial` — added in the *same* 2.1.4 tier, under #104 — it still was not.

`dispersion_classes()` justified omitting `zi` and `hu` like this:

> `zi` and `hu` are deliberately absent: bayesnec models those as a second parameter block, so a prior on them carries class "b" with a prefixed `nlpar`, not a class of their own.

**That is true for `hurdle_gamma` and `zero_inflated_beta`, and false for the two count families**, which are single-block fits. For those, `zi` is an ordinary `brms` parameter with a class of its own, on the `brms` default:

```r
brms::get_prior(bf, data = d, family = validate_family("zero_inflated_poisson"))
#>        prior class nlpar  source
#> 1 beta(1, 1)    zi        default
#> 2                b  beta default
```

So `get_priors()` reported nothing for it, and `usable_prior()` dropped a supplied one — asymmetric with `shape`, which #207 made round trip. For these families `zi` is the zero-inflation probability, which is the quantity the family exists to estimate.

## Scope: this is a reporting gap, not a rejection

Worth stating plainly, because the issue's title could suggest otherwise. #223's other half already fixed the *input* side by accident: `make_inits()` filters to `class == "b"`, which drops **every** non-`"b"` class rather than only the dispersion ones, so a supplied `zi` prior already reached `brm()` and correctly got no initial value. It simply could not be discovered or round tripped.

## The fix

- `usable_prior()` keeps `zi` and `hu` alongside `sigma`/`shape`/`phi`.
- `dispersion_classes()` → **`auxiliary_classes()`**, since it is no longer only dispersion, with roxygen distinguishing the two kinds and recording why including `zi`/`hu` is a **no-op** for the two-block families rather than a conflict — their second block carries class `"b"` with a prefixed `nlpar`, so no class-`zi` row is ever generated.
- `source == "user"` still does the work of not reporting a `brms` default as though `bayesnec` had chosen it.
- The `make_inits()` comment is corrected: it described the `class == "b"` filter as being about dispersion, when the filter is general. That generality is the right rule and needs no maintenance as families are added, but the comment claimed something narrower than the code does.

## Tests

Four added to `test-get_priors.R`. Full runs: `test-get_priors.R` **47**, `test-inits_functions.R` **260**, `test-helpers.R` **37**, `test-define_prior.R` **75** — all passing.

- a user prior on `zi` round trips for a zero-inflated count family;
- **a brms *default* on `zi` is still dropped** — the test that would catch over-widening;
- a `zi` prior does not disturb the initial-value search and gets no init;
- **the two-block families are unaffected** — asserts the no-op claim rather than leaving it as a comment.

## Left open, deliberately

Whether `bayesnec` should **generate** a weakly informative `zi` prior rather than leaving `beta(1, 1)` is a modelling decision, not an implementation one, and is escalated rather than settled here — per the phase rule in `06_review_run.md`. #207's hazard note ruled generating a default *dispersion* prior out of scope because it would change every existing fit; the same argument applies less forcefully, since these two families are new in this release and have no existing fits to change. Recorded on #231 for RF.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: bayesnec
 Title: A Bayesian No-Effect- Concentration (NEC) Algorithm
-Version: 2.1.3.11
+Version: 2.1.3.15
 Authors@R: c(person("Rebecca", "Fisher", email = "r.fisher@aims.gov.au", role = c("aut", "cre"),comment = c(ORCID = "0000-0001-5148-6731")), person("Diego R.","Barneche",role="aut",comment = c(ORCID = "0000-0002-4568-2362")), person("Gerard F.","Ricardo",role="aut",comment = c(ORCID = "0000-0002-2220-3971")), person("David R.","Fox",role="aut",comment = c(ORCID = "0000-0002-3178-7243")))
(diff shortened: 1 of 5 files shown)


beckyfisher (commented):
This looks ok

PR #238, opened 2026-08-22
Add example9: a complete analysis workflow (#219)

beckyfisher:
**Release tier: 2.2.0.** Adds `vignettes/example9.Rmd.orig`, a single analysis
run from data to reportable estimate. Closes #219.

## What

The steps of an analysis in the order they occur, worked on `herbicide` simazine,
with a count response following it for the two steps that take a different form
there. The vignette separates the reasons a model fails the sampler screen, which
is the part of the workflow no other vignette shows and the part that decides
whether excluding a model is right.

Three files against `dev`: the vignette source, its NEWS entry, and the version
counter.

## Why it matters

`vignette("example1")` through `example6` each demonstrate one part of the
package. Nothing shows a user the sequence: choose the family, fit the set,
screen the sampler, check the fit, exclude, report.

Nothing distinguishes the reasons a model fails that screen either, and they are
not alike. An equation given too few draws, an equation whose posterior the
sampler cannot traverse, and an equation the design does not support are reported
as the same value in the same column, and the right response to each is
different. Only the first is repaired by more sampling.

## Evidence

Rendered in full against `dev` at `25466e53` --- 84 chunks, none errored, no chunk
cache, `cmdstanr` as `precompile.R` sets. Every count in the prose is an inline R
expression read from the fitted objects, so no statement can go stale against its
chunk output.

The screen returns a different outcome at each of the three fits:

| fit | screen removes | retained |
|---|---|---|
| simazine, `iter = 4000, chains = 2` (1600 draws) | 5 of 11 | 6 |
| simazine, `bnec()` defaults (8000 draws) | 1 of 10, and `ecxll5` fails to fit | 9 |
| `nassarius` A, `bnec()` defaults | 8 of 10 | 2 |

Of the five removed at the reduced settings, three pass once the settings are
raised, `ecxwb2` still fails on divergent transitions, and `ecxll5` fails to fit.
That is the three-way separation, measured rather than asserted.

`ecxll5` is the only five-parameter equation in the `decline` set, and `simazine`
has six distinct concentrations, so with `phi` that is six parameters against six
concentrations. The parameter distinguishing it from `ecxll4` is the asymmetry
exponent, identified from the shoulder of the curve, and the series does not reach
a floor.

On `nassarius` every one of the eight exclusions involves divergent transitions,
from 120 to 5,589. Survival is complete at a dose of 1.25 and near zero at 2.5
with nothing tested between, so the threshold is unidentified across a two-fold
[...]

PR #239, opened 2026-08-23
Correct work queue item 7, and track the #148 spec

beckyfisher:
Notes only — no R code, no tests, nothing that affects the package.

## Why

RF asked me to re-read the #148 thread against `01_work_queue.md` item 7. **The entry was wrong in four ways, and everything downstream inherited it**: PR #226 was built to it, and the phase 2 review validated #226 against the queue rather than against the issue.

1. **"Three parts"** — there are four. Part D (sampler diagnostics and screening) was folded into #148 deliberately, with a stated rationale.
2. **"Decision (d) RF flagged but did not answer"** — RF answered it *in*, in the Final decisions table, and prototyped it in a later comment.
3. **Decision (b) omitted entirely** — a message at the end of `bnec()` **and** a line in `summary()`, thresholding on the ratio.
4. **The `example2` doc fix that rides on (b)** was not mentioned.

## What is in here

- **`01_work_queue.md` item 7 rewritten**, carrying a correction notice, all four settled decisions, the four parts, and the note that Part D is the independent half — it needs no `loo` and does not depend on PR #217, so the spec says build it standalone.
- **`notes/tasks/` is now tracked.** `notes/tasks/148-model-fit-diagnostics.md` is the authoritative 23 KB spec, written at scoping time and referenced from the issue thread — **and it was untracked, so it was not in git**. Neither the queue entry nor the review consulted it. It is the single most useful file for whoever picks this up.
- **`06_review_run.md` gains a START HERE handoff block** — outstanding work, PR state, merge order, and the two things that changed late (the machine is free, so #190 is unblocked; a full-suite run to characterise the `WARN 10` did not finish).

## The lesson, recorded rather than glossed

Four of the five phase 2 findings were gaps between what code did and what was *claimed* about it. The review then made the same mistake one level up — validating PRs against a summary of the requirements instead of the requirements. That is now written into the handoff block.

## Outstanding, all confirmed in scope pre-CRAN by RF

| what | where |
|---|---|
| (b) automatic surfacing | into #226 |
| (d) combined hurdle check | into #226 — prototype already on the thread |
| **Part D (D1–D4)** | its own PR |
| `example2.Rmd.orig:102` doc fix | with Part D |
| rewrite #238 against Part D | after Part D |

### DIFF
--- a/notes/implementation/01_work_queue.md
+++ b/notes/implementation/01_work_queue.md
@@ -276,67 +276,90 @@ which of the two you are building and why.
 
 ---
 
-## 7. #148 — `check_fit()`, `pp_check()` methods, LOO-PIT
+## 7. #148 — `check_fit()`, `pp_check()` methods, LOO-PIT, and Part D
 
-Fully scoped across two long comments, RF's decisions settled 2026-08-21. **Read
-both comments in full before starting** — the scoping comment carries a worked
-demonstration that is the whole justification for the design.
+> **CORRECTED 2026-08-23.** The entry that stood here was wrong in four ways and
+> everything downstream inherited them — PR #226 was built to it, and the phase 2
+> review validated #226 against *this file* rather than against the issue. Read
+> **`notes/tasks/148-model-fit-diagnostics.md`** — the authoritative spec, written
+> at scoping time and referenced from the issue thread — before doing anything
+> here. It is 23 KB and covers Parts A–D, the settled decisions, the hazards and
+> a definition of done. **Neither the previous queue entry nor the phase 2 review
+> consulted it.**
+>
+> What was wrong:
+>
+> 1. **"Three parts"** — there are **four**. Part D (sampler diagnostics and
+>    screening) was folded into #148 deliberately, with a stated rationale.
+> 2. **"Decision (d) RF flagged but did not answer"** — RF answered it *in*, in
+>    the Final decisions table, and then **prototyped it** in a later comment.
+> 3. **Decision (b) was omitted entirely** — a message at the end of `bnec()`
+>    **and** a line in `summary()`, thresholding on the ratio.
+> 4. **The `example2` doc fix that rides on (b)** was not mentioned.
 
-**Settled:** part B in scope and extended to control lack-of-fit; the averaged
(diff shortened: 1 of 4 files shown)


beckyfisher:
## Addendum — Part D was started, then interrupted

Branch **`issue-148-part-d`** is pushed (no PR opened). It carries **D1–D4**:

- `R/check_sampling.R` — one row per candidate: `max_rhat`, `min_ess`, `min_ess_ratio`, `n_divergent`, `failed`. Both spec-mandated routes verified against `manec_example` before use: `nuts_params()` for divergences and `neff_ratio() * ndraws()` for absolute ESS, with a comment at the point of choice saying why **not** `rstan::get_num_divergent()`.
- `R/screen_models.R` — the screening wrapper and the D4 guard. `amend()` is untouched.
- `tests/testthat/test-check_sampling.R`.

**The tests were written but never run — the session was interrupted mid-run. Treat them as unverified.** The functions themselves were exercised by hand and all three cases behaved correctly:

```
check_sampling(manec_example)
      model max_rhat  min_ess min_ess_ratio n_divergent failed
1 nec4param 1.213298 10.13503     0.1013503           0   TRUE
2 ecx4param 1.087736 24.30083     0.2430083           0   TRUE

case 1 (all pass)  -> unchanged, amend() not called
case 2 (some fail) -> drops nec4param, returns a bayesnecfit
case 3 (all fail)  -> "nothing left to average", not amend()'s bare message
```

**D5 is not started**: the `summary()` block covering both axes, the hard-coded 1.05 in `print.manecsummary` (`R/print.R:136`), and making `vignettes/example2.Rmd.orig:102`'s divergence claim true.

### One deliberate deviation from the spec

The spec calls for a new stored fixture, because nothing in the package has a mixed pass/fail model set. The tests instead place the cutoff between the two `manec_example` models' Rhats (1.213 and 1.088), which exercises that case exactly and reproducibly with no fixture to maintain. A real fixture may still be wanted for the #219 vignette, where a realistic demonstration matters. Flagged rather than decided.

beckyfisher (commented):
Updates correct

PR #240, opened 2026-08-23
Add check_sampling() and screen_models(), and align Rhat on 1.01 (#148 Part D)

beckyfisher:
**Release tier: 2.2.0** — opens that heading in `NEWS.md`, see *Merge note* below. Branched from `dev`, independent of the 224–228 stack.

Implements **Part D of #148**, per `notes/tasks/148-model-fit-diagnostics.md`. No open decisions remained on it.

## Why this is a separate PR

The spec says so explicitly: Part D *"touches no posterior predictive machinery, needs no `loo`, and does not depend on PR #217"*, where parts A and B both do. *"Build and merge D first while A and B wait on #217; that also gets screening into the package before #219 is written."*

**Part D was missing from `01_work_queue.md` item 7 entirely**, which is why it was not built with #226 — see #239 for that correction.

## D1/D2 — `check_sampling()`

One row per candidate model:

```
      model max_rhat  min_ess min_ess_ratio n_divergent failed
1 nec4param 1.213298 10.13503     0.1013503           0   TRUE
2 ecx4param 1.087736 24.30083     0.2430083           0   TRUE
```

Both routes were verified against `manec_example` before use:

- **Divergences via `brms::nuts_params()`**, not `rstan::get_num_divergent(x$fit)`. The latter is the obvious thing to reach for and works often enough to look correct, but it reaches past `brms` into the `stanfit` slot and `rstan` is `Suggests`-only — so it fails under `cmdstanr`. There is a comment at the point of choice saying so.
- **ESS via `neff_ratio() * ndraws()`**, giving the absolute rather than a ratio. Since `neff_ratio()` is `min(ess_bulk, ess_tail) / ndraws`, `ess_cutoff = 400` is exactly Vehtari's "both bulk and tail exceed 100 per chain" at four chains, with no arithmetic left to the user. The ratio is reported alongside so "passed because we drew 8000" can be told from "passed efficiently". No `posterior` dependency.

Documented rather than left to be discovered: **a heavily thinned fit can fail a cutoff that a ratio would have passed**, and the correct response is to retain more draws, not to lower the cutoff — thinning lowers ESS by construction, so the tempting fix is the wrong one.

The **divergence default of 10 has no literature behind it** and the help page says so. Stan's guidance is that *any* divergence means the sampler failed to explore the posterior. Ten is pragmatic, from practice with these non-linear models; the honest version is more defensible than an implied citation, and it will be quoted in methods sections.

## D3/D4 — `screen_models()`

Screens on the sampler **only**. A poor `check_fit()` is a modelling result, and dropping on it silently would hide exactly what the user needs to see. All three cases the spec enumerates:

```
all pass   -> returned unchanged, amend() never called
some fail  -> intersect against held models, drop; 2 models -> 1 returns a bayesnecfit
all fail   -> "nothing left to average", not amend()'s bare
              "All models removed, nothing to return"
```

The guard lives in the wrapper; **`amend()` is not modified**, per D4. The drop message names *every* failing criterion, not just the first — a model that failed on two counts is a different situation from one that scraped past on one.

## D5 — Rhat aligned on 1.01, and the summary stops grepping

- `rhat_cutoff` default **1.05 → 1.01** in `rhat.bayesnecfit`, `rhat.bayesmanecfit`, `rhat.bayesnechurdlefit` and `summary()`.
- **`summary()` now computes its verdict** with `rhat()` instead of `has_r_hat_warnings()`, which searched `brms`'s captured warning text for the literal `"some Rhats are > 1.05"`. That made the threshold `brms`'s to set rather than `bayesnec`'s, and it fails **silently**: `brms (>= 2.23.0)` is a floor, not a ceiling, so a reworded warning would make every model report no issue and the summary quietly stop warning. Latent fragility rather than a live bug — but it is the failure mode where silence reads as a pass.
- `print()` reports the cutoff **in use** rather than a hard-coded 1.05.
- `has_r_hat_warnings()` deleted; `summary.R` was its only caller.

### `example2` rewritten, and a sentence corrected

With the default at 1.01 the two chunks contrasting default-against-conservative became **identical calls** and the prose was wrong. The passage now explains 1.01 with the citation, shows the looser 1.05 as the comparison, and points at `check_sampling()`.

The same passage said *"Here we get a message because none of our models failed the default Rhat criterion"* — backwards on both counts: no message appears in the rendered output, and `rhat.bayesmanecfit` only messages when **all** models fail. Removed.

**`@Vehtari2021` was missing from `vignettes/bayesnec.bib`** and has been added. The vignette has been citing the 1.01 recommendation in prose while the code defaulted to 1.05, with no bib entry behind the claim.

## Tests

`test-check_sampling.R` is new (26). `test-bayesmanec_methods.R` 41, `test-helpers.R` 37 — all passing.

**One deliberate deviation from the spec.** It calls for a new stored fixture, because nothing in the package has a mixed pass/fail model set. The tests instead put the cutoff between the two `manec_example` models' Rhats (1.213 and 1.088), which exercises that case exactly and reproducibly with nothing to maintain. **A real fixture may still be wanted for the #219 vignette**, where a realistic demonstration matters rather than a contrived threshold — flagged rather than decided.

**Two changes to existing tests, both deliberate.** `test-bayesmanec_methods.R` pins the exact field names of the summary object, and `rhat_cutoff` is now among them because `print()` needs it. The assertion is updated *and* two more added — that it defaults to 1.01 and that passing `rhat_cutoff = 1.5` is honoured — so the contract is pinned harder than before rather than merely relaxed to fit.

## ⚠️ Merge note — this PR opens the 2.2.0 heading

`dev` has no `# bayesnec 2.2.0` heading yet. **PR #226 also opens one.** Whichever merges first opens it; the second will conflict on a duplicate heading, which is a two-line dedup.

## Gating

The `example2.Rmd.orig` edit lands here but **the rendered `.Rmd` stays stale until the #190 precompile run**, so the published vignette will show the old prose against the new default in the interim.


beckyfisher:

```
```

beckyfisher:
## Independent review

Read the full diff and checked the numerical claims against installed brms 2.23.0 and `manec_example` rather than taking the description on trust.

### What holds up

Verified, not assumed:

- **`min_ess = min(neff_ratio(x)) * ndraws(x)` is exact.** `brms:::neff_ratio.brmsfit` really is `matrixStats::rowMins(cbind(ess_bulk, ess_tail)) / ndraws(draws)`, so the multiply-back is arithmetic, not approximation.
- **The "400 = 100 per chain at 4 chains" framing is right.** `helpers.R:762-773` confirms bnec defaults of `chains = 4`, `iter = 1e4`, `warmup = floor(iter/5)*4 = 8000` → 8000 post-warmup draws, so the "ratio 0.05 at bnec defaults / 0.40 at `thin = 3`" note in the docs checks out.
- **`nuts_params()` over `rstan::get_num_divergent()`** — correct, and worth the comment.
- **`names(x$mod_fits) == success_models`** (`expand_classes.R:253`), so `held` in `screen_models()` is consistent with `check_sampling()`'s row set.
- **`handle_set()` does raise "All models removed, nothing to return"** (`helpers.R:266`), and does silently no-op when every dropped name is absent — the docstring's account of `amend()`'s behaviour is accurate.

---

### 1. The `prior_*` variables are inside the screen

`R/helpers.R:765-766` forces `sample_prior = "yes"` on **every** bnec fit. So every brmsfit carries `prior_b_bot`, `prior_b_top`, … `prior_sigma`, and `sampling_row()` takes `max()`/`min()` over the whole vector. Those are independent draws from the prior — their Rhat is Monte Carlo noise and carries no information about whether the sampler explored the posterior.

Measured on `manec_example`:

```
ecx4param   prior_b_bot   rhat 1.023
            prior_b_beta  rhat 1.012     <- both over the new 1.01 cutoff
            worst real param: b_beta 1.088
```

This is pre-existing in `rhat.bayesnecfit`, not introduced here — but moving the cutoff from 1.05 to 1.01 is exactly what makes it bite, and `check_sampling()` inherits it. A model can now be failed on a variable that is not part of the model. Fix is one line: drop `^prior_` before reducing.

While there, decide `lprior`/`lp__` deliberately rather than by omission. On `nec4param`, `lprior` is the worst on *both* metrics (Rhat 1.213, ESS 10.14) — a hair above `b_bot` (1.211, 10.16), so the numbers in the PR description aren't misleading, but that's coincidence rather than design.

### 2. NA propagation — and #244 makes it reachable

`sampling_row()` has `max(rhat(fit), na.rm = TRUE)` but `min(neff_ratio(fit))` with no `na.rm`. `posterior::ess_bulk()` and `posterior::rhat()` both return `NA` for a constant variable (confirmed directly). Constant variables are exactly what the `constant()` prior work on `issue-244-constant-prior` produces, and it targets the same `dev`.

The failure chain is silent and self-contradictory:

```
min_ess NA -> failed NA
  -> failed <- tab$model[tab$failed]   gives NA_character_
  -> length(failed) == 1, so the "all passed" branch is skipped
  -> intersect(NA, held) is character(0)
  -> message "Dropping 0 of 2 candidate models that failed…" with an empty reason list
  -> amend(x, drop = character(0)) returns the object unchanged
```

Add `na.rm = TRUE`, and force `failed` to be non-NA (`isTRUE(...)`) — or better, report "not assessable" as its own state rather than folding it into a logical.

The same missing `na.rm` is in `rhat.bayesnecfit` (`any(rhat_vals > rhat_cutoff)`). That matters more now than it did: `summary()$rhat_issues` used to come from a grep and was always `TRUE`/`FALSE`; it now comes from `rhat()` and can be `NA`. `print.manecsummary` then does `names(x$rhat_issues[unlist(x$rhat_issues)])`, and an `NA` index yields a length-1 element named `NA` — so the warning reads *"The following model had Rhats > 1.01: - NA"*. `which()` around the `unlist()` fixes the print side.

### 3. `try(amend(...))` is too broad

Any error from `amend()` becomes "Every candidate model failed the sampler screen". Two concrete misdiagnoses:

- A `bayesnecfit` that fails the screen hits `amend.bayesnecfit`'s deliberate *"Cannot drop models from a bayesnecfit: it contains the single model …"* (`R/amend.R:126-131`) — a good message, replaced by a wrong one about averaging.
- `amend_model_set()` recomputes loo weights and predictions; anything that errors in there is reported as a convergence result.

Both numbers are already in hand, so decide it up front and let everything else propagate honestly:

```r
if (length(to_drop) == nrow(tab)) stop("Every candidate model failed …", call. = FALSE)
out <- amend(x, drop = to_drop)
```

### 4. `pull_out()` is an expensive way to reach `$fit`

`sampling_fits()` calls `pull_out()` per model, which reconstructs a whole `bayesnecfit` — predictions, posteriors — to get at the brmsfit. `x$mod_fits[[m]]$fit` is the same object (checked: `identical` rhat from both routes) for a fraction of the cost.

This matters more than it looks, because `summary.bayesmanecfit()` now calls `rhat(x, ...)`, which does the same `pull_out()` loop. `summary()` previously grepped a stored string; on a 5–10 model set it is now materially slower, on an operation users call constantly. Worth fixing in both places.

### 5. Documentation

- **Stale and self-contradicting:** `R/check_sampling.R` Thresholds section (and `man/check_sampling.Rd`) says *"Note `rhat` defaults to the older 1.05."* This PR's own D5 changed it to 1.01.
- `print.manecsummary`'s `if (is.null(x$rhat_cutoff)) 1.01` — an old stored `manecsummary` was computed against the 1.05 grep, so 1.05 is the honest fallback for a missing field.
- `print.manecsummary` still says *"Consider dropping them (see ?amend)"*, and `example2.Rmd.orig` says *"should be dropped with `amend(drop = )`"*. Both should now point at `screen_models()` — that is the PR's own thesis.
- `chk_numeric()` doesn't enforce length 1, but the docs say "vector of length 1". `chk_number()` does, and is what `amend()` uses for scalars.

### 6. Loose ends

- **`extract_warnings()` is now dead code.** `has_r_hat_warnings()` was its only caller; it survives only in a comment. It is also the sole user of `evaluate`, which could come out of `Imports`.
- **`bayesnechurdlefit` doesn't inherit `bayesnecfit`** (`bnec_hurdle.R:219` allots `c("bayesnechurdlefit", "bnecfit")`), so `check_sampling()` will reject a hurdle fit with "requires a bayesnecfit or a bayesmanecfit" — while this same PR updates `rhat.bayesnechurdlefit` to 1.01. Either route it through `hurdle_delegate()` (the pattern is at `bayesnechurdlefit-methods.R:470`) or state in the docs that hurdle fits are not covered.
- The `autoplot.R` tidyselect change is unrelated to Part D and isn't in NEWS.
- DESCRIPTION says `2.1.3.18` while NEWS opens `# bayesnec 2.2.0` above an existing `# bayesnec 2.1.4`. Not a blocker, but the two disagree.

### 7. Tests

The new file is solid on what it covers. Gaps:

- Nothing asserts the **verdict** changed. `rhat_cutoff` is pinned as stored and echoed, but the D5 claim is that `rhat_issues` is now *computed*. Add `summary(manec_example, rhat_cutoff = 1.5)$rhat_issues` all `FALSE` against both `TRUE` at 1.01 — that is the assertion that would catch a regression to grepping.
- No `screen_models()` on a `bayesnecfit` (see finding 3).
- The NA path (finding 2) is untested, and given #244 lands on the same branch, it should be.

The decision to put the cutoff between the two fixture Rhats instead of adding a stored fixture is the right call — same coverage, nothing to maintain.

---

**Overall:** the design judgements are sound and unusually well argued — screening on the sampler only, absolute ESS over ratio, `nuts_params()` over `rstan`, and being explicit that the divergence default has no citation behind it. Findings 1, 2 and 3 are the ones worth addressing before merge; 1 and 2 are the two that can produce a wrong answer silently, which is the failure mode this PR sets out to eliminate elsewhere.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

beckyfisher:
## Review addressed

All twelve findings. Full suite on this branch: **1638 pass, 0 fail**.

**Findings 1 and 2 turned out to be one defect**, and are fixed together in a
single `screenable_pars()` in `R/rhat.R`, which `rhat()` and `check_sampling()`
now both reduce over. It excludes two things:

- **`prior_*` draws.** As measured — `ecx4param` has `prior_b_bot` at 1.023,
  over the new cutoff, while nothing in the model itself is. `lp__` and
  `lprior` are kept, and the reason is now written down rather than left to
  omission: unlike `prior_*` they are deterministic functions of the *posterior*
  draws, so they do carry a convergence signal.
- **Zero-variance parameters**, for which `posterior` returns `NA`. Rather than
  `na.rm`-ing the NA at each reduction, it leaves the screen: a parameter fixed
  at a known value has nothing to converge to, so "not assessable" and "failed"
  never get folded into one logical in the first place. `failed` is non-NA by
  construction in both `rhat()` and `check_sampling()`, and `max_or_na()` keeps
  the degenerate empty case from reading as a pass via `-Inf`.

**This also closes finding 1 of the #246 review** — `rhat()` erroring on a
multi-model fit with a `constant()` prior. Same NA, seen from the other side.
The fix belongs here because every consumer of these diagnostics is in this
diff and none are in that one; putting it there would have meant diagnostics
code in a priors PR and a conflict in `R/rhat.R`. #246 merges second and adds
the end-to-end regression test, which needs both halves to exist.

**3.** Decided from the table up front, as suggested. The per-model reasons now
go into that error too — the all-failed case is the one a methods section most
has to account for, and an error saying only "everything failed" sends the user
back to `check_sampling()` to find out what. A failing `bayesnecfit` now gets
this error rather than `amend.bayesnecfit`'s "Cannot drop models from a
bayesnecfit"; tested both ways.

**4.** `x$mod_fits[[m]]$fit` in `sampling_fits()` and in `rhat.bayesmanecfit()`.

**5.** Stale 1.05 note gone; `print.manecsummary` falls back to 1.05 and points
at `screen_models()`; `chk_number` for every scalar, with a test that a length-2
cutoff is refused. `example2.Rmd.orig` points at `screen_models()` — see the
caveat below.

**6.** `extract_warnings()` deleted and `evaluate` dropped from `Imports`.
`check_sampling()` and `screen_models()` now take a `bayesnechurdlefit`,
delegating to both components via `hurdle_delegate()` / `hurdle_rewrap()` as
the other model-set operations on that class already do. `autoplot.R` is now in
NEWS. On the version disagreement: `master` is 2.1.3.1 and `dev` 2.1.3.17, so
2.1.4 is unreleased and is where dev versions accumulate — the `2.2.0` heading
was the odd one out and has been folded into 2.1.4, leaving `DESCRIPTION` at
2.1.3.18.

**7.** Added: the verdict changes with `rhat_cutoff` (all `TRUE` at 1.01, all
`FALSE` at 1.5 — the assertion that catches a regression to grepping);
`screen_models()` on a `bayesnecfit`, both failing and passing; the NA path via
`screenable_pars()` and via an injected old-style `rhat_issues`; the 1.05
fallback; the `prior_*` exclusion on a real fit; and the hurdle delegation.

### One caveat, not from the review

`vignettes/example2.Rmd` is precompiled and still documents the 1.05 default —
it was already out of sync with this PR's `.orig` edits before I touched it. I
kept my `example2` change prose-only rather than adding a `check_sampling()`
chunk, so as not to widen the gap, but **example2 needs a precompile run before
release**, and it would be worth a worked `check_sampling()` / `screen_models()`
example when that happens. Raising it against #242, which already owns
`precompile.R`.

Issue #206, opened 2026-08-15
Zero-bounded models are excluded for gaussian, but they fit cleanly and model weights reject them when wrong


beckyfisher:
## Summary

`bayesnec` drops zero-bounded models — those with an exponential decay and no
`bot` parameter (`nec3param`, `ecxexp`, `ecxsigm`, `ecxwb1p3`, `ecxwb2p3`,
`ecxll3`) — whenever `family = gaussian()`. The stated reason is that they
"cannot generate predictions of negative response values".

That conflates two different things: the range of the **mean function** and the
support of the **likelihood**. A Gaussian likelihood evaluates `f(y | mu, sigma)`
and the data enter only through the residual `y - mu`; the sign of `y` is never
tested. A mean function asymptoting to zero with Gaussian error is internally
consistent — near the asymptote it *predicts* negative observations at a rate
set by `mu` and `sigma`, and those are ordinary negative residuals.

The practical consequence: the model shape that OECD TG 201 and Ritz, Gerhard &
Streibig (2026, doi:10.1007/s10651-025-00698-y) both recommend for algal
growth-rate data — a curve with a lower asymptote fixed at zero, representing
complete growth inhibition, with Gaussian error — cannot be fitted in
`bayesnec`. That includes `nec3param`, the package's namesake model.

**This was tested rather than argued, and the restriction turns out not to be
protecting anything.**

## What the code does

`R/check_models.R:97` — keyed on the **family**, irrespective of link:

```r
if (fam_tag == "gaussian") {
    use_model <-  model[!model %in% mod_groups$zero_bounded]
    drop_model <- setdiff(model, use_model)
    if (length(drop_model) > 0) {
      message(paste("Dropping the model(s)", ...,
                    "as they are not valid in the case of Gaussian y data."))
    }
    if (length(use_model) == 0) {
      stop("None of the model(s) specified are valid for Gaussian y data.")
    }
```

Behaviour is a **silent drop with a message** — except that a user who names a
single zero-bounded model explicitly gets an **error**, because the set empties.
There is no distinction between `models = "all"` (where dropping is arguably
helpful) and a deliberate, explicit request for `nec3param`.

### The link clause is *not* vestigial — correcting a premise

`R/check_models.R:20` applies a separate exclusion for `logit`/`log` links. It
would be reasonable to assume this is unreachable, since `bnec()` is often
described as forcing an identity link. **It is reachable.** `validate_family()`
forces identity only when `family` is a character string
[...]

PR #243, opened 2026-08-24
Add example7: negative growth rates and the zero boundary (#193)

### LINKED ISSUE #193: Rewrite `example7` as a growth-data case study using the `alga` dataset

beckyfisher:
## What

`example7`, a vignette on responses that can legitimately take negative values — specific growth rate above all — and on what the common practice of removing those negatives does to the toxicity estimate that gets reported.

Closes #193.

## Why it matters

Replacing a negative growth rate with zero is standard practice, in this package and in others, and the reasons are long-standing: a negative value cannot be expressed as a percentage of the control without exceeding 100, and several commonly used equations cannot generate a negative mean at all.

On simulated data, where the truth is known and every convention is fitted the way an analyst would fit it — `bnec()` defaults, model-averaged over the family's declining candidate set — the conventions that keep the measurement recover the truth and the conventions that replace it do not. Across a precision sweep from a 9.6% to a 1.9% control coefficient of variation:

| convention | ErC50 bias, p1 → p4 | ErC50 coverage at p4 | NSEC bias, p1 → p4 |
|---|---|---|---|
| `measured` | -0.4% → -0.2% | 0.97 | +3.5% → +3.2% |
| `censored` | -0.7% → -0.2% | 0.99 | -0.6% → +2.4% |
| `deleted` | -2.4% → -0.0% | 0.99 | -0.5% → +0.7% |
| `floored` | -3.4% → **-2.3%** | **0.24** | -11.9% → **-52.9%** |
| `beta` | -4.4% → **-5.8%** | **0.52** | -1.4% → **-42.7%** |
| `gamma` | +4.3% → +0.2% | 0.96 | +25.0% → **-71.6%** |

The bias of a convention that replaces the measurement does not shrink as the experiment improves, so its interval narrows around a displaced centre and coverage collapses. A laboratory that adds replicates and reduces its control variability produces an analysis that is more confident and no less wrong. `gamma` converges on ErC50 and instead displaces ErC10, from -8.7% to **-25.0%** with coverage of **0.01**, so these conventions cannot be summarised across estimates either.

None of it is visible in the diagnostics. Every one of the 4,200 fits converged and returned all three estimates, and equations failing a convergence check hold 0.4% of the model weight averaged over the study, 5.2% at the worst single cell and convention.

The clearest single result is what the model weights say. Given the measurements as recorded, averaging identifies the generating equation and does so more sharply as the experiment improves — `nec4param` takes 0.51 of the weight at a 9.6% control CV and **0.85** at 1.9%. Under flooring it takes **0.000**, and the six equations whose lower asymptote is fixed at zero take **0.98**. A model weight describes the data supplied to the fit, not the process that produced the measurements, and on floored data those are different questions.

A posterior predictive check says where the Gamma's failure is. These data are generated with a residual standard deviation constant along the curve, so a correctly specified family should predict that value everywhere. `measured` does, and so does `floored`. An identity-link Gamma over-states the spread at the control by a factor of 2.5 and predicts a control-to-floor gradient of 90 to 181 against a true value of one, because it ties the dispersion to the mean. So flooring damages the mean structure and not the variance; under a Gamma the two failures compound.

The vignette also records why nothing in the current defaults requires the substitution. A Gaussian mean is unconstrained, the six zero-asymptote equations are retained under a Gaussian family from #206, and an absolute ECx is reported for them, so the curve shape TG 201 describes can be fitted on the measurements as recorded and weighed against the free-asymptote shapes by the same model averaging as everything else.

And it says where the question does not arise: in the cell whose concentration series stops at the zero crossing, where about two observations in seventy are negative, the four Gaussian conventions span **0.4 percentage points** of ErC50 bias between them.

## Evidence

Simulation and case studies: [open-AIMS/negative-response-conventions](https://github.com/open-AIMS/negative-response-conventions). Seven cells × 100 realisations × 6 conventions = 4,200 model-averaged fits for the simulation, 24 for the case studies, run as SLURM arrays against `bayesnec` at a pinned `dev` commit inside the container `bayesnec`'s own `hpc/image.lock` identifies. No unit failed and no estimate was unidentified. Model weights are `bnec()`'s default, pseudo-BMA with a Bayesian bootstrap.

Every figure in the vignette is transcribed from that compendium's result tables and read out of them by inline code, so a re-run changes the numbers without touching prose.

## Status

Ready. The study has been re-run against the pinned `dev` commit, the vignette's transcribed blocks regenerated from the new tables, and both the source and the generated vignette pushed.

The re-run was 4,224 model-averaged fits with no failures. Only the NSEC changed. #326 seeks the crossing from the control upward, so a draw crossing between the control and the next grid point is no longer discarded: the point estimate changes by 0.1 to 0.5 percentage points and the lower bound by about three times that. Every ErC10, every ErC50 and all 42 rows of the model-weight table are identical to the digit, which is why `results/ppc_spread.csv` was not regenerated.

Two corrections were made while checking the prose against the new tables. Section 6.2 reported the largest mean weight on an equation failing a diagnostic as under half a per cent; that figure is the mean over all 4,200 fits, and the largest for any one cell and convention is 5.2%, under `gamma` in `ctl`. Section 9 now names the NSEC change as a third difference from the CRAN release, alongside the priors and the initial-value search.

`dev` has moved since the pinned commit, and the estimation path has not: `R/ecx.R`, `R/nsec.R`, `R/bnec.R`, `R/fit_bayesnec.R`, `R/inits_functions.R` and `R/priors.R` are all byte-identical between the pin and `dev` at `f54e4bdc`. The changes since are post-fit helpers and RNG restoration, so the merged state reproduces the numbers reported here.

<details>
<summary>What changed in this rewrite, and what was removed</summary>

The study behind the vignette was redesigned (#296). The previous version held the equation fixed at `nec4param` for every arm, which answers a question about the likelihood rather than the question a reader has, and on altered data fitted an equation that did not generate them — so part of what it measured was misspecification the convention did not cause. Every convention now calls `bnec()` with its own default priors and averages over its family's declining set.

Consequences for the document:

- eight lettered arms become six named conventions: `measured`, `censored`, `deleted`, `floored`, `beta`, `gamma`. `deleted` is new and was recorded as an untested practice in the previous version;
- the fixed-equation versus model-averaging split, and the dispersion sub-model section, are removed — the new design has no counterpart for either;
- the `bot` prior sensitivity check is removed: it tested a shared prior the new design does not use;
- the residual scale is generated constant and fitted constant. The previous design generated it rising 8.1-fold, described as calibrated from the real data; re-measured on `alga` with substituted rows excluded, one of the four datasets shows a gradient and the 8.1 figure was a single dose-group ratio on that one dataset;
- a section on the equations the data are made to resemble is added, which is where the model-weight result is reported;
- the case studies moved into the compendium. Four datasets by six conventions, each averaged over thirteen or fourteen equations, is about a fifteen-hour precompile. Section 9 records what that gives up: the case-study estimates no longer track whatever version of `bayesnec` they are published beside.

Sampler settings for the case studies match the simulation — `iter = 4000`, `warmup = 2000`, `adapt_delta = 0.99`, `max_treedepth = 12`. Measured on the two worst-mixing fits, that takes tail ESS from 127 and 57 to 2,327 and 1,892. In the full averaged fit no equation holding any weight fails either diagnostic.

A dispersion sweep was run and is deliberately not reported. It asked whether letting a family's dispersion vary along the curve repairs the bounded families. It does not, the Beta needs no such correction and is made worse by one, and adding one to a Gaussian manufactures a variance gradient the data do not have. The code and tables stay in the compendium, and its `CLAUDE.md` records why they are uncited.

</details>


beckyfisher:
## Independent review

Read the vignette in full rather than the PR description, re-derived the simulation design from the vignette's own code, cross-checked every quoted percentage against the transcribed literals, and verified the package-behaviour claims against `dev`'s source.

**Verdict: the argument is sound and the transcription is clean. Two things block merge, and both are about the branch this lands on rather than about the science.**

---

### Blocking

**1. B2 and B3 do not run on `dev`. This PR depends on #244.**

The pinned-asymptote arms are shown as:

```r
prior_fixed_bot$prior[prior_fixed_bot$nlpar == "bot"] <- "constant(0)"
fit_B2 <- bnec(y ~ crf(x, "nec4param"), data = dat, ..., prior = prior_fixed_bot)
```

`constant()` prior support exists only on `issue-244-constant-prior` (`e153ed2e`) — it is on neither `dev` nor this branch. On `dev`, `make_inits()` hits the hard-coded gamma/normal/beta/uniform table and fails with *"attempt to apply non-function"*. The vignette's prose note says initial values must be supplied and `b_bot` omitted, but the chunk doesn't show them — so a reader copying it gets an error. Once #244 merges, that note becomes wrong.

Either state is incorrect, so this needs deciding rather than deferring: merge #244 first and drop the note, or add the `init` list to the chunk. `eval = FALSE` hides it at build time, which makes it more likely to be missed, not less.

**2. "`bnec()` has no route to a distributional `sigma`" is false on `dev`.**

From *Limitations*:

> Residual error is generated heteroscedastic and fitted homoscedastic … it is present because `bnec()` has no route to a distributional `sigma`.

`dev` has `disp(~x)` and the `"power"`/`"twosided"`/`"loglinear"` variance functions (`R/bayesnecformula.R`, `R/disp_model.R`). The documentation there *specifically* addresses this case — `"power"` takes `log(mu)` and so is refused for a response on the real line, pointing the user at `"loglinear"`, whose form `log(dpar) = c0 + c1*(mu - m)` is close to what `sim_sigma()` generates. The limitation is real but the reason given is not: it should read that the study predates the feature / did not use it, with re-running under `disp("loglinear")` named as the next step.

---

### Scope: this closes #193 but doesn't do what #193 asked

A maintainer decision rather than a fix. #193's plan specified the shipped `alga` dataset ("Nothing in the package uses it yet — it exists for this vignette") and called the dispersion demonstration **"the payoff"**. Neither is delivered: the case studies are transcribed summary numbers, and dispersion appears only as a limitation. Also unlanded from that plan: the TG 201 Annex 5 quotation, the `sgr`/`density` "same measurement on two scales" framing, and the note that `type = "relative"` is not the ecotoxicological definition when `bot` is negative.

What is delivered is arguably stronger — a scored simulation beats an unscoreable case study — but closing #193 leaves `alga` still unused and the dispersion demonstration untracked.

---

### Numeric and factual issues

| issue | detail |
|---|---|
| **NSEC bullet mixes baselines** | "The NSEC is biased high, by **37 to 134%** relative to the intact analysis". 134% is B2 against the *true `nec`* (3.036/1.3); against A it is 112% (ratio 2.125). The low end against A is B1 at 34% (1.337), not E's 37%. Relative to A the range is **34–113%**. |
| **B1's coverage is not zero** | "B1, B3 and E contained the true ErC50 in **none** of the 500". `precision` gives B1 `coverage = 0.002` at cv 1.9 — one dataset. B3 and E are 0.000. |
| **Model-averaging baselines don't match the precision table** | The "before" figures disagree: E −10.3% vs −10.6%, F −5.3% vs −5.6%, F −15.7% vs −15.6%, B3 (δ=8) −7.8% vs −7.6%. Presumably because phase 10 is n = 200 on a subset — but nothing says so, and the section explicitly invites cross-checking. One sentence fixes it. |
| **Beta shift is silent** | "In both cases … `bnec()` reports the shift with a message." `check_data()` messages for `Gamma` only; the `beta` branch nudges without a word. This *helps* the argument — E floors the data silently — so worth correcting rather than dropping. |
| **"biased low by 5 to 10 per cent"** | B1 is −4.9%. Trivial, but it's a headline sentence. |
| **`top_factor = 1.0` framing** | "there are no data in the region that determines it" — at 1.0, `prop_negative` is 1/14, not 0. |
| **B2 vs B3 tension** | The model-averaging preamble says "under B3, E and F the constrained curve visibly cannot follow the data", but the diagnostics section makes the opposite point — B3/E/F sample cleanly, and it is **B2** that "cannot reach its own observations". The mechanism described (flatten `beta`, push `nec` right) is right; the framing borrowed from B2 isn't. |

---

### Figures

`vignette-fig-example-curves-1.png` plots `ex_dat` — the **intact** measurements — in all eight panels, including the four fitted to floored data. The chunk comment says the opposite:

```r
# the actual dataset these eight models were fitted to, rather than a fresh draw,
# so that the points shown are the points the curves were estimated from
```

True for A, B2 and (in effect) C only. The caption's "open points are the nine negative values that the conventions act on" is defensible as a display choice, but a reader will read B1/B3/E/F as failing to fit their own data. Say that the intact measurements are shown throughout for comparison, and fix the comment.

All eight curves are drawn from **posterior-mean parameters**, not the posterior mean curve. For a nonlinear model those differ, and most for B2, whose posterior is widest. Disclosed indirectly ("Posterior means from the eight fits") but worth one clause in the caption.

`real-ratios` renders correctly, including the `rlo = 0` B2 intervals — they run off the panel edge under `scale_x_log10()` rather than vanishing, which reads as intended. `precision-bias` uses `facet_grid(scales = "free_y")`, which frees by row and so keeps the three handling columns comparable — correct.

---

### Housekeeping

- **No `NEWS.md` entry and no version bump.** `dev` is at 2.1.3.17 with a per-PR entry convention, and `example6`'s addition has one.
- **Dead link definitions**: `[e2b]`, `[e3]`, `[e4]`, `[e6]` are defined and never used in `example7`.
- **`example1`'s new `bot` table still has no provenance** — flagged in the PR description. Its numbers are recoverable: a 3-day test, 3871 cells mL⁻¹ initial, counting limit 10, which is `r_salina` from the same compendium (control rate 1.431 d⁻¹ → R = e^4.29 = 73.1). It can take the same treatment as `example7`'s.
- **Pre-existing duplicate bib keys** `ritz2026` and `Ritz2026`, both already on `dev`, same paper. Not introduced here, but `example7` cites one of them.
- **Length**: 14,737 words rendered vs `example6`'s 11,459 — 1.29×, not "half again". It earns the length; the PR description overstates the gap.

---

### Verified correct

Worth recording what held up:

- **The twelve-scenario table reproduces exactly** from the vignette's own functions — `top`, `bot`, `zero_crossing`, `x_max`, ErC10/ErC50, control CV 9.6/6.7/2.8/1.9, `prop_negative` 0 / 1⁄14 / 1⁄7. The `true_ecx()` and `zero_crossing` inversions of `nec4param` are algebraically right, and the scenario-8 concentration grid matches `ex_dat` to the last decimal — the example dataset really is a draw from the stated design.
- **`show_params("decline")` returns 14 models**; `mod_groups$zero_bounded` drops exactly the six named (`nec3param`, `ecxexp`, `ecxsigm`, `ecxwb1p3`, `ecxwb2p3`, `ecxll3`) under Gaussian in `check_models.R:176`; eight remain. One ambiguity: "reduced from the full list to eight" is within the *decline* group — a real `model = "all"` retains more.
- **`ecx_x_absolute()`** does compute `max(y) - max(y)*(x/100)`, matching the stated `f(x) = top*(1 - x/100)` and the scale-invariance argument for approach E.
- **`nsec()`** defaults to `sig_val = 0.01` and takes the reference from the lowest-x column — as described.
- **`check_data()`** shifts to `min(y[y>0])/10` and `1 - 0.001`, as described.
- All seven citation keys resolve; `warne2025` and `negativesgr2026` are both actually cited.
- **#181 is closed**, so the `example6` correction is right and removing the stale reference is warranted.
- `example1.Rmd` and `example1.Rmd.orig` diffs are byte-identical outside chunk bodies — the hand-edit claim checks out.
- The provenance section is the right call, and the pinned commit plus named CSVs make the transcription auditable. The self-imposed "never edit a result in place" comment is a good discipline to have committed.

---

**Recommendation:** approve on substance. Resolve (1) by settling merge order with #244, correct (2) — it is a factual claim about the package in the package's own docs — and take the numeric fixes. (3) is a call for the maintainer, but if #193 closes here, the dispersion demonstration and `alga` need a new tracker.

🤖 AI-assisted review generated with [Claude Code](https://claude.com/claude-code)

beckyfisher:
Note from the #246 review, for whichever of these lands second.

#246 makes a `constant()` prior work when passed straight to `bnec(prior = )`,
which is what `example7`'s pinned-asymptote arms currently work around by
hand-writing an `init` list for every *other* parameter in order to trip the
`!("init" %in% names(brm_args))` escape hatch in `add_brm_defaults()` and skip
the init search entirely.

Once #246 is in `dev`, that workaround can be deleted — set the `constant()`
row in the prior set and let the search run. Worth doing here rather than
leaving a worked example of a workaround in a vignette.

beckyfisher:
## Correction to the review above, and one real finding

Two things in my earlier comment need withdrawing, and one narrower finding survives.

### Withdrawn: "`bnec()` silently accepts a malformed prior"

It does not. `validate_priors()` catches it and says so:

```
Named prior list does not contain priors for model nec4param.
Using bayesnec default priors.
```

I had wrapped every fit in `suppressMessages(suppressWarnings(...))` and never saw it. The
guard from #207 is present and working; the fault was in my test method, not the package.

### Withdrawn: the demonstration behind it

My first round of fits ran against the **installed** bayesnec 2.1.3.7 rather than this
branch's 2.1.3.20 — `library(bayesnec)` takes the library copy and I had not switched to
`devtools::load_all()`. Anything I reported from that round is void.

### What survives: the vignette names the wrong accessor

`pull_prior()` returns `list(object$fit$prior)` — an **unnamed** length-1 list. So in the
Methods chunk:

```r
shared_prior <- pull_prior(fit_A)
prior_fixed_bot$prior[prior_fixed_bot$nlpar == "bot"] <- "constant(0)"
```

`$nlpar` on a list is `NULL`, the subscript matches nothing, `bot` is never pinned, and the
"one shared prior across the six Gaussian arms" control does not happen — each arm falls
back to its own defaults with the message above. `?pull_prior` states the contract
("It is not accepted by `bnec(prior = )` — `get_priors` is the function for feeding priors
back in"), and `tests/testthat/test-get_priors.R:6` says the same.

Verified against 2.1.3.20 with `load_all()`: `get_priors()` returns a bare 4-row
`brmsprior`, and `constant(0)` through it does pin — `b_bot` comes back with sd = 0 and
B2's ErC50 separates from A's.

Why the vignette had it wrong is chronology rather than carelessness:

```
54986ed0  2026-08-14  add get_priors() for both entry points
e45be25f  2026-08-18  add example7: negative response values
```

Later in calendar time, but `git merge-base --is-ancestor` says `get_priors()` is not an
ancestor of the example7 commit — the vignette was written on a branch that had not yet
merged it, and `get_priors(fit_A)` has never appeared in `example7.Rmd.orig` in any commit.
The two have only sat side by side since this PR merged `dev`.

**The study's numbers are unaffected.** An unpinned B2 would be bit-identical to A, and the
compendium reports B2 at 5.81 divergent transitions against A's 0.00, coverage 0.97 against
0.93, and bias -9.5% against +0.6%. The compendium pinned successfully; only the vignette's
rendering of the method into exported-interface code was wrong. Fixed in the Methods prose,
its printed output, and the new case-study code.

### The one thing possibly worth an issue

`pull_prior()` returns an **unnamed** list for a single fit, which is what defeats
`validate_priors()`'s named-list branch. Had it returned `setNames(list(...), model)`, that
branch would have accepted it and the trap would close entirely. `get_priors()` returns a
bare `brmsprior` for a single model, so the two accessors differ in shape as well as in
content. Small, but it is the mechanism by which a plausible-looking line does nothing.

🤖 AI-assisted review generated with [Claude Code](https://claude.com/claude-code)

beckyfisher:
## Handoff — state at `0a8b5d35`

Everything below is committed and pushed. Nothing is held in a scratch directory that matters; the two things that were are now in `notes/`.

### Status

**Mergeable.** `dev` merged in at 35 commits behind (now current), `DESCRIPTION` at 2.1.3.23, `NEWS.md` carries both this branch's entries and dev's. R-CMD-check and pkgdown were still running when this was written — **check them before merging**, they have not been seen green.

The rendered `vignettes/example7.Rmd` is **current**, not stale: it comes from a real local precompile against the branch package. Verified 0 Stan-noise lines, 0 errored chunks, 2167 lines.

### If the vignette needs precompiling again

Two traps, both of which cost time this session:

1. **The installed `bayesnec` is 2.1.3.7 and the branch is 2.1.3.23.** `library(bayesnec)` silently takes the stale library copy, so a precompile against it builds the wrong package. Either `devtools::load_all(".")`, or `R CMD INSTALL --library=<scratch> .` and put that first on `.libPaths()`. Do not install over the main library.
2. **Run it from the repo root.** `fig.path` is relative and `precompile.R` moves figures from the root into `vignettes/`. An ad-hoc knit of one chunk leaves strays at the root — `bb977541` committed six of them, and `.Rbuildignore` has no rule, so they would have shipped. Now covered by a `/vignette-fig-*.png` rule in `.gitignore`.

Local precompile of `example7` alone takes **~137 minutes** (24 `alga` fits). CI takes ~2h56m, inside the 350-minute ceiling.

Per dev's #190 policy a stale rendered `.Rmd` is not a merge blocker, so a re-run is only needed if `example7.Rmd.orig` changes.

### What the precompile exposed, and why it is worth re-reading before trusting a future run

The first precompile exited clean — no errored chunks — and would have shipped three defects. `precompile.R`'s checker passes on all of them:

- 2304 lines of Stan chain progress, because Stan writes to stdout and knitr captures it. Fixed with `results = "hide"` on the fitting chunk, which is what `example1` does.
- The `real-ratios` figure was unreadable: the transcribed B2 lower bound was exactly `0`, which `log10` drops, while a fitted one is tiny but positive, so the axis spanned four decades. Fixed by clipping the bar to the panel.
- The prose still quoted the compendium while the figure came from the fit. All eleven figures are now interpolated from a helper chunk.

**A clean precompile does not mean a good vignette.** Read the rendered output.

### Open items

1. **CI is unverified.** See above.
2. **A follow-up issue was proposed and never opened**, since #193 closes here. It would cover the two things #193 scoped that did not land: a dispersion sub-model demonstration as a vignette section (#193 called it "the payoff"), and re-running the full precision sweep under `disp("loglinear")`. The second matters because the bounded check in `notes/disp_sensitivity_example7.md` shows B2's headline numbers are largely a pooled-`sigma` artefact — one scenario, n = 50, so it revises the vignette's claims without re-scoring the study.
3. **`example7` is ~14,700 words**, the longest vignette. Flagged for a judgement call, never made.

### Things established this session that are easy to lose

- `pull_prior()` returns an **unnamed length-1 list** and its payload carries duplicated `Intercept` rows and a `sigma` default. It is not usable as a `bnec(prior = )` argument and `get_priors()` is. This is deliberate and documented — do not "fix" `pull_prior()` to round-trip; that would turn a caught error into a silently accepted bad prior. `example7` used the wrong one because it was written on a branch predating `get_priors()`.
- `ecx()` defaults to **`ecx_val = 10`**. Three diagnostic scripts here computed ErC10 while labelling it ErC50, which produced a phantom "bistability" and a wrong recommendation. Always pass `ecx_val` explicitly.
- Arm C on data with a detection limit censors a below-limit row at **the counting limit, not zero** — the tighter statement, and the one the test supports. That choice reproduces the compendium exactly on both `r_salina` ErC50s; censoring at zero was 3–30% out.
- The `alga` case studies reproduce the compendium: 35 of 41 ratios within 2%, and the interpolated prose values land at 0.75 / 0.43 / 0.25 / 2.7 / 116 against transcribed 0.74 / 0.43 / 0.26 / 2.9 / 116.

### Two workflow fixes on this branch

Both still absent from `dev` and both needed; they survived the merge into dev's rewritten matrix workflow.

- `BH`, `RcppEigen`, `StanHeaders` added to the precompile workflow's `extra-packages`. They are `LinkingTo` deps of rstan: the binary install does not pull them, but precompiling compiles a Stan model at run time. Without them every `bnec()` call dies instantly with "Boost not found". The workflow added in #215 had never successfully rebuilt a vignette with fitting chunks; `R-CMD-check` has carried `BH` since 2021.
- `precompile.R`'s error check reported a bare `#> Error:` because knitr wraps long messages onto continuation lines and it printed only `hits[1]`. It now carries the wrapped text — which is how the Boost error was found at all.

🤖 AI-assisted; generated with [Claude Code](https://claude.com/claude-code)

beckyfisher:
## Two of the three open items are discharged

Checked while working through the queue; the third is still RF's.

**1. CI is verified.** The handoff note records R-CMD-check and pkgdown as
"still running when this was written -- check them before merging, they have not
been seen green". Both completed **success** on `negsgr-cens-vignette`.

**2. The follow-up issue is opened**: #283, covering the dispersion sub-model
section #193 called "the payoff" and the re-run of the precision sweep under
`disp("loglinear")`. It records that the bounded check in
`notes/disp_sensitivity_example7.md` shows arm B2's headline numbers to be
largely a pooled-`sigma` artefact, and that this qualifies the vignette's claims
rather than re-scoring the study.

**3. The 14,700-word length** is a judgement for RF and is untouched.

## One thing that has changed under this branch since the handoff

`example7` will need re-rendering whatever is decided about its length, because
**#281 redefines every ECx the vignette reports**. Every ECx is now measured from
the control rather than from the maximum of the predicted curve, `type` gains
`range` and redefines `relative`, and an ECx the curve never reaches within the
tested range returns `NA` rather than the nearest grid point. `example7` reports
ErC50 and ErC10 values throughout.

The branch is 66 commits behind `dev` and merges cleanly (`git merge-tree`
reports no conflicts), so the rebase is not the obstacle. Per D13 the full
precompile runs once, after all three vignettes are settled, so **do not
re-render this branch yet** -- #281 and #282 have to land on `dev` first or it
gets rendered twice.

beckyfisher:
## What

A prose and structure pass on `vignettes/example7.Rmd.orig` (2be961e9,
0205cc1a). No
other file is touched, and `vignettes/example7.Rmd` is deliberately left
unrendered.

**No number, estimate or claim is changed.** The numeric tokens in the file
were compared before and after: the only ones removed are those inside deleted
duplicate sentences (`B2`, `B3`, "the last three panels", one "ErC10"). Nothing
was added, rewritten or rounded.

## Why it matters

The vignette is the longest in the package and had accumulated the register
patterns sections 12 and 13 of `CLAUDE.md` now rule against, together with
several passages stating the same point two or three times at different levels
of detail.

## What changed

**Register (section 12).** The five ruled verb cases are corrected wherever
they appeared, leaving none: 16 lines using *carry* (a model does not carry a
term, a fit does not carry an estimate), 27 using *move* applied to a
parameter, an estimate or a bias, both using *cost*, and all three using
*help* or *hurt*. Also
removed: "worth reading carefully", "worth drawing out", "the same story told
by the families", "the ErC10 puzzle", "accuracy bought with width", "the
channel closes", "B2 is largely exonerated", "not chased down", "the sampler
reports this honestly", "so the answer is no, emphatically".

**Naming (section 12).** "the four-parameter NEC model `nec4param`" and "the
three-parameter NEC model" now name the equation without the ambiguous "NEC
model"; "for a `nec` model the NSEC converges on `nec`" is now "for an equation
that includes a `nec` parameter". Four uses of *endpoint* for what the vignette
itself defines as a *toxicity estimate* are corrected, including
`ecx(type = "absolute")`, which the text had called an endpoint two sections
after ruling that it is not one.

**Headings (section 13).** One correction: `## Case studies: four algal growth
tests` (colon appositive) is now `## Four algal growth tests`. Every other
heading already satisfied the rule. Two bold lead-ins that were labels rather
than claims now state a claim.

**Cuts: about 320 words, roughly 2% (16,667 to 16,347).** Every cut removes a
second or third statement of something already established:

- the reading of the twelve-scenario figure, which restated the design table
  immediately above it;
- one of four statements that E is B3 arrived at by a different route;
- one of three statements of the Gamma's constant coefficient of variation;
- one of three statements of "a discrepancy that does not shrink with the noise
  is misspecification";
- the long restatement of the B2 dispersion result in *Credible interval
  coverage*, now a short forward reference to *Limitations* where it is
  established;
- the restatement of the NSEC definition at the head of *Scoring the NSEC
  against the true nec*, now a reference back to the Background;
- the restated reason `nec3param` makes E and F the counterpart of B2 and B3,
  in *Limitations*;
- the closing sentence of the B2 *Limitations* paragraph, which repeated its
  own bold lead-in.

Deliberately not touched: *Provenance of the reported results* (only
`carries` to `contains` and `move them` to `alter them`, with the frozen-figure
disclosure intact) and the case-study statement that its figures are computed
rather than transcribed.

## Claims left alone

None found to be wrong. Two notes for the record rather than for action:

- *ErC10 and ErC50* reports F's high-precision ErC50 bias as -15.6% and *Model
  averaging* reports the same quantity as -15.7%. The difference is the
  500-dataset against the paired 200-dataset baseline, which *Model averaging*
  already explains.
- *Methods* says the candidate set is "reduced to eight declining models" and
  then prints all fourteen. The sentence now reads "the fourteen declining
  models are reduced to eight", which matches what the chunk shows, but no
  figure changed.

<details>
<summary>Mechanics</summary>

**Files.** `vignettes/example7.Rmd.orig` only. `git status` is clean in the
`bayesnec-negsgr` worktree; nothing else was modified in any worktree. The
branch remains 81 commits behind `dev` by design.

**Rendered vignette.** `vignettes/example7.Rmd` is untouched and is now stale
with respect to its source. Per #190 that is not a merge blocker; the
precompile runs once for the whole package.

**Wrapping.** Paragraphs whose text changed were rewrapped to 79 columns.
Paragraphs that did not change were left byte-identical, so the diff is
confined to edited text. Two paragraphs contain inline `` `r ... ` ``
expressions; those were rewrapped by hand so that no inline expression spans a
line break, and the file was checked for that condition afterwards. Code chunk
contents, table rows, the HTML comment and the indented model-form display were
excluded from rewrapping.

**Checks run.** Numeric-token diff (above); markdown heading list diff, which
shows one intended change; fence balance (84 before and after); table row and
bullet counts unchanged. The vignette was not knitted: inline `` `r ` ``
expressions evaluate regardless of `eval = FALSE`, so a knit would require the
case-study fits, which this pass deliberately did not run.

**Not done, and why.** `CLAUDE.md` section 10 would ordinarily open a
`prompts/` log for a change to package documentation. The session instruction
restricted edits to `vignettes/example7.Rmd.orig`, so no log file was created.

</details>

beckyfisher:
The vignette has been rewritten against the landed stack and the PR body now describes the new state rather than the old one.

**What moved.** `dev` at `a7ba018f` is merged in. #206 removed the Gaussian zero-bounded exclusion and the matching absolute-ECx refusal, which made the previous version's central claim false — it argued the defaults were already correct *because* the package excluded those equations. The argument is now that the curve shape and the treatment of the data are separate decisions, and only the second causes the bias. Prose is 8,199 words, from 11,674.

**A new arm.** G fits `nec3param` directly on the intact measurements. It returns 0.32 times the intact ErC50 on `c_proliferum`, against B2's 0.32 for the same constraint imposed through a prior — the same model reached two ways, which is what including it establishes. Both sit well below the intact analysis there, which the vignette attributes to the single residual scale these fits use rather than to the constraint, consistent with B2's simulated bias falling from -12.7% to -1.2% under `disp("loglinear")`.

**Two claims I had written were contradicted by the recomputed figures and are fixed**: the floored analyses sit *below* the intact analysis on `c_proliferum`, not above, and arm G's agreement with B2 does not show that pinning the asymptote is free. The counts of estimates flagged unusable are now computed rather than typed, so they cannot go stale silently.

**The rendered vignette is stale and is not rebuilt here.** A full single-vignette precompile against a 2.1.3.31 build succeeded (341 minutes, 28 fits, no errored chunks) and was discarded because it predated those two corrections. Rebuilding belongs in a full-package precompile rather than a second single-vignette run — see the note in the body about `library(bayesnec)` loading the installed package, which is the trap that made the first two attempts produce a vignette describing the new behaviour while demonstrating the old.

beckyfisher:
The PR body has been rewritten and no longer describes the same study.

The vignette is rebuilt against the redesigned simulation (#296), in which every convention is fitted with `bnec()` defaults and model-averaged over its family's declining set, rather than with the equation held at `nec4param`. The eight lettered arms are now six named conventions, and every figure quoted in the body has changed: flooring's ErC50 bias is -2.3% at the highest precision rather than -5 to -16%, and its interval coverage is 0.24 rather than zero, because the enlarged candidate set absorbs part of the distortion. The NSEC result is larger and in the opposite direction to the previous version, -52.9% rather than +47 to +134%.

Removed from the vignette, and so from the body: the fixed-equation versus model-averaging split, the dispersion sub-model section, and the `bot` prior sensitivity check. Added: a section on which equations each convention makes the data resemble, which is where the clearest result now sits.

Two things a reviewer should know before reading the diff. `vignettes/example7.Rmd` is deliberately not regenerated in this commit, so the source and the generated vignette disagree; the Status section of the body says why, and it is blocked on #309 and #304. And the Limitations section still presents the tight `ec50` prior as an open concern, which #304 has since fixed — that paragraph is known-stale and will be rewritten when the branch takes `dev`.

beckyfisher:
The body has been updated and the Status section rewritten.

Two things had gone stale. The headline table still carried the figures from the first run of the compendium, before it was re-run against the current default priors — for example `measured` on ErC50 read -0.8% → -0.2% where it is -0.4% → -0.2%, and every NSEC figure for the retaining conventions was wrong, `measured` reading +13.8% → +3.9% against an actual +4.0% → +3.3%. The table is now generated from `results/metrics.csv` rather than typed.

And the Status section said `example7.Rmd` was not regenerated and named #309 and #304 as the blockers. Both have merged, the vignette has been rebuilt against the re-run, and source and generated file are in sync. Status now records what actually blocks finalisation: #326 changed `nsec()` so that draws whose NSEC is the control concentration are no longer discarded, which affects every NSEC in the study, and #336 changed initial-value seeding. The compendium needs re-running against a raised pin before this is final.

Two corrections went into the vignette in the same pass. The model weights were described as stacking weights; `bnec()` uses pseudo-BMA with a Bayesian bootstrap unless `loo_controls` is supplied and the compendium never supplies it, so Table 2 and its caption were wrong and the method is now named where the conventions are introduced. And Section 5.3's posterior predictive figures were taken from a single saved fit per cell; they are now transcribed from a 100-realisation measurement, which changed the Gamma's control ratio from 3.5 to 2.5 and its control-to-floor gradient from five thousand to 181.

beckyfisher:
## Outcome of the re-run

The study has been re-run against the pinned `dev` commit and the vignette
refreshed from the new tables. 4,224 model-averaged fits, no failures, 100
usable realisations in every one of the 126 cells. The body above has been
updated to the current state: the NSEC column of its headline table, the
sentence about failing diagnostics, and the Status section.

Only the NSEC changed. #326 seeks the crossing from the control upward, with the
control made the first point of the searched grid, so a draw crossing between
the control and the next grid point is no longer discarded. The point estimate
changes by 0.1 to 0.5 percentage points and the lower bound by about three times
that. Every ErC10, every ErC50 and all 42 rows of the model-weight table are
identical to the digit, so `results/ppc_spread.csv` was not regenerated and
Section 5.3 is unchanged.

| estimate | p1 | p4 | changed by the re-run |
|---|---|---|---|
| `floored` NSEC | -11.9% | -52.9%, coverage 0.00 | -0.4 and -0.1 points |
| `gamma` NSEC | +25.0% | -71.6%, coverage 0.04 | -0.5 and -0.1 points |
| `measured` NSEC | +3.5% | +3.2%, coverage 0.88 | -0.5 and -0.1 points |
| every ErC10 and ErC50 | | | unchanged |

Every conclusion the vignette draws is unaltered. `censored` and `deleted`
remain the only two conventions holding nominal coverage on all three estimates
across the precision sweep, at a minimum of 0.94. All 1,400 Beta and Gamma fits
still returned all three estimates.

## Corrections made while checking the prose

Section 6.2 reported the largest mean weight on an equation failing a diagnostic
as under half a per cent. That figure is the mean over all 4,200 fits; the
largest for any one cell and convention is 5.2%, under `gamma` in the `ctl`
cell. Four definitions were tested against the committed table before concluding
the sentence was wrong rather than the arithmetic. The contrast it supports
stands, since the case studies it is set against sit at 52% and 58%.

Section 9 now names the NSEC change as a third difference from the CRAN release,
alongside the log-scale `nec` and `ec50` priors and the per-chain initial-value
search.

## Evidence

Compendium [open-AIMS/negative-response-conventions](https://github.com/open-AIMS/negative-response-conventions)
at `31b3114`, which `vignettes/bayesnec.bib` now cites. Run as three SLURM
arrays at 200 resident units, about 11.5 hours wall clock.

<details>
<summary>Verification, and what it establishes</summary>

### The regeneration of the transcribed blocks

The blocks were regenerated by script rather than by hand. The script was first
run against the superseded tables and reproduced all three transcribed blocks
byte for byte, so the 132 changed lines are the re-run and nothing else.

### The equivalence test run before the study

All 42 fixed priors are identical between the superseded pin and the new one, so
the Stan source is unchanged and the compile cache stayed warm. Refitting
`p1`/`measured`/iteration 1 under both gave identical ErC10, ErC50 and all 14
per-equation weights, with only the NSEC changing: point estimate -0.0121, lower
bound -0.0423. The lower bound changing three times as far is what #326 predicts.

### The estimation path against current `dev`

`R/ecx.R`, `R/nsec.R`, `R/bnec.R`, `R/fit_bayesnec.R`, `R/inits_functions.R` and
`R/priors.R` are byte-identical between the pinned commit and `dev` at
`f54e4bdc`. What changed since are post-fit helpers and RNG restoration, neither
of which is in the path the study calls, so merging does not stale these numbers.

### The conflict resolved in merging `dev`

The branch had added `BH`, `RcppEigen` and `StanHeaders` to
`precompile-vignettes.yaml` so that `rstan` could compile a Stan program at
render time. #313 has since changed that workflow to sample with cmdstan, which
needs none of them, so `dev`'s version was taken whole.

### The saved posterior draws

The draws behind every estimate are now recorded, 174 KB per unit against 10 MB
for a whole fit. A future change to `nsec()` or `ecx()` then needs only a
re-summary rather than 1,340 core-hours of refitting.

### The record of the pin

The compendium commit `78a46c4` has a subject naming `e226e797` while the lock
file it wrote records `9269bf6d`; `dev` advanced between the fetch and the
commit, and the difference is CI, `NEWS.md` and `precompile.R` only, with no
package code. The lock file is the record. It was already pushed, so the history
was left alone and the discrepancy noted in the following commit message.

</details>

PR #246, opened 2026-08-24
Accept a constant() prior through bnec(), so a parameter can be fixed (#244)

### LINKED ISSUE #244: make_inits() and sample_priors() fail on a constant() prior, so fixing a parameter needs a hand-written init list
`make_inits()` and `sample_priors()` both look a prior's distribution name up in a
hard-coded table of four sampling functions:

```r
fcts <- c(gamma = rgamma, normal = rnorm, beta = rbeta, uniform = runif)
```

A `constant()` prior has no entry, so both fail. Reproduced on `dev` (2.1.3.17):

```r
p <- bayesnec:::define_prior(model = "nec4param", family = gaussian(link = "identity"),
                             predictor = seq(0, 10, length = 20),
                             response = rnorm(20, 1, 0.1))
p$prior[p$nlpar == "bot"] <- "constant(0)"
bayesnec:::make_inits("nec4param", c("b_top", "b_bot", "b_nec", "b_beta"), p, 2)
#> Error in fcts[[fct_i]](1, v1, v2) : attempt to apply non-function
```

## Why this matters

Fixing a parameter at a known value is the request in #84, closed won't-implement
on the grounds that a `constant()` prior already delivers it without new API
surface. That closing comment is right about the recommendation but records this
as the reason the hard-constant route "has to go via `pull_brmsfit()`" rather than
straight through `bnec()`. It does not have to: `bnec()` already has an escape
hatch, because `add_brm_defaults()` skips the init search entirely when the caller
supplies their own initial values (`R/helpers.R:806`):

```r
if (!("init" %in% names(brm_args)) || skip_check) {
```

So `bnec(prior = <with constant()>, init = <list omitting the fixed parameter>)`
works today. That is what `example7` (#193, `negsgr-cens-vignette`) does for its
B2 and B3 approaches, where `bot` is pinned at zero. It works, but it asks the
user to hand-write initial values for every other parameter in order to fix one,
which is a poor trade for what should be a one-line prior change.

## Scope

Four sites, because simply dropping the constant rows before the search is not
enough — the init search evaluates the candidate curve, and the curve genuinely
needs `bot = 0`. The constant must carry its value *through* the search and be
stripped only where the list is handed to `brm()`.

1. **`make_inits()`** (`R/inits_functions.R:44`) — for a `constant(v)` row, set the
   value to `v` instead of sampling. Do this *after* the `check_args` name check,
   which compares the prior names against the model's parameters and must still
   see the fixed one.
2. **Bound rejection in `make_inits()`** — the `while` loops that redraw until the
   value falls inside `lb`/`ub` would spin forever on a constant outside them.
   Skip rejection for constant rows.
3. **`refine_inits()`** (`R/inits_functions.R`) — carries its own copy of `fcts`
   and redraws `b_slope` / `b_d` / `b_beta`. A constant on any of those hits the
   identical error. Skip constant rows when choosing tunable parameters.
4. **`sample_priors()`** (`R/sample_priors.R:35`) — same table, so the exported
   function errors on any prior set containing a fixed parameter, i.e. a user
   cannot inspect the priors they have just written. A constant "samples" as a
   point mass: return `rep(v, n)`. (`check_priors()` is unaffected — it goes
   through `brms::hypothesis()` on the fitted object, not through
   `sample_priors()`.)

Then strip the constant entries from the init list before it reaches `brm()`
(`R/helpers.R:863-872`), because Stan does not declare a parameter whose prior is
constant and will reject an init for it.

The hurdle path (`make_good_hurdle_inits()`) shares `make_inits()`, so it should
follow, but wants its own test.

## Not in scope

No new `bnec()` argument. #84's reasoning stands: fixing an asymptote is usually
the wrong move, and where it is right a `constant()` or tight informative prior
says so explicitly. This issue only makes the route that is already documented
work without a hand-written `init` list.

## Follow-on

- A "Fixing a parameter" subsection in `vignettes/example3.Rmd.orig` (priors),
  which currently does not mention `constant()` at all. `example2b` and `?models`
  cover it only inside the `drc::NEC.2()` comparison.
- Once both are in, a single comment on #84 recording the settled position.
- `example7`'s practical note after approach B2 states that initial values must be
  supplied when `bot` is fixed. That becomes stale with this fix and should be cut
  back to the part that stays true: a user supplying their *own* inits must still
  omit the fixed parameter.

beckyfisher:
Closes #244.

A `constant()` prior now goes straight through `bnec(prior = )`, so a parameter
can be fixed at a known value with a one-line change to the prior set:

```r
p <- get_priors(y ~ crf(x, model = "nec3param"), data = avoid_data,
                family = Beta(link = "identity"))
p$prior[p$nlpar == "top"] <- "constant(0.5)"

fit <- bnec(y ~ crf(x, model = "nec3param"), data = avoid_data,
            family = Beta(link = "identity"), prior = p)
```

`top` comes back as exactly 0.5 with no error, because Stan never declares it.

## The bug

`make_inits()`, `refine_inits()` and `sample_priors()` each look a prior's
distribution name up in a hard-coded table of `gamma` / `normal` / `beta` /
`uniform`. A `constant()` prior has no entry, so the init search stopped with
"attempt to apply non-function".

Fixing one parameter therefore obliged the user to hand-write initial values for
every *other* parameter, in order to trip the `!("init" %in% names(brm_args))`
escape hatch in `add_brm_defaults()` and skip the search entirely. That is what
`example7` (#193) does for its pinned-asymptote arms.

## Why it is four sites and not one

The obvious fix — drop the constant rows before generating inits, as suggested
when #84 was closed — breaks the search. `make_good_inits()` evaluates the
candidate curve against the response range, and `get_init_predictions()`
name-matches the init list against the model's own arguments; a parameter fixed
at `bot = 0` is genuinely part of that curve. Drop it early and every candidate
fails, so the search burns 10,000 trials and falls back to Stan's defaults.

The value is instead assigned rather than sampled, carried *through* the search,
and removed only in `add_brm_defaults()` where the list is handed to `brm()`.
That removal is deliberate hygiene rather than a required fix: Stan moves a
constant parameter out of `parameters` into `transformed parameters`, and both
rstan and cmdstanr accept and ignore an init for one. `bayesnec` does not send
brm() an init for a parameter Stan does not declare, rather than depend on both
backends continuing to ignore it.
Alongside that: bound rejection is skipped for a constant (the `while` loop
would otherwise spin forever, a hang rather than an error), `refine_inits()`
skips a fixed parameter when choosing what to re-draw, and `sample_priors()`
draws the point mass so a prior set containing a fixed parameter can still be
inspected.

`check_priors()` is unaffected — it goes through `brms::hypothesis()` on the
fitted object.

## Also in here

**`sample_priors(plot = NA)`** returns the draws, as documented. The guard
tested `!plot %in% c("ggplot", "base")`, and `NA %in% ...` is `FALSE`, so the
one value documented to return a list of sampled values was the one value
rejected. Unrelated to constant priors, but in the same function and with no
test covering it.

**A "Fixing a parameter" section in `vignette("example3")`.** The priors
vignette did not mention `constant()` at all; the only coverage was inside the
`drc::NEC.2()` comparison in `example2b` and `?models`. The new section is
worked on a simulated two-choice avoidance assay where the control expectation
is 0.5 by design, shows the mechanics, and then makes the two arguments against
reaching for it — the free fit's `top` interval covers 0.5 and is positive
evidence the assay was unbiased at dose zero, which fixing discards; and a
constant asserts the value is known exactly, so nothing propagates into the
*NEC*, whose interval narrows accordingly. An informative prior is offered as
the honest version, and a short subsection closes on the control-normalised
case.

## What has not changed

No `fixed` argument. The reasoning in #84 stands: for control-normalised data
the right intervention is one step earlier, and for a structurally known
asymptote the free estimate is usually the most informative thing the experiment
produces. This only makes the route that comment already recommended work
without a hand-written `init` list.

## Verification

- End to end: `nec4param` with `bot = constant(0)` samples and recovers
  `top` 3.04 / `nec` 4.02 against a truth of 3.0 / 4.0. `b_bot` is absent from
  Stan's `parameters` block, not from the draws — `brms` carries a constant
  parameter through as a zero-variance column, which is what made the `rhat()`
  NA in finding 1 of the review reachable.
- Full test suite: 1262 pass, 0 fail. Twelve new tests across
  `test-inits_functions.R` and a new `test-sample_priors.R`, including a
  regression test for the bound-rejection hang.
- `example3.Rmd` re-rendered from `.orig` against this branch.

## Notes for review

- The unrelated churn in `example3.Rmd` is the `tag` column in `pull_prior`
  output, which post-dates the last precompile run — #190.
- `precompile.R` does not set `options(width = )`, so committed vignette output
  depends on the renderer's `.Rprofile` (the existing files were produced at
  ~115). This render matches that. Worth fixing in #242, which already touches
  that file.
- No conflicts expected with the open stack: nothing else touches
  `R/inits_functions.R`, `R/sample_priors.R` or `example3`, and the seven
  stacked PRs' only `R/helpers.R` change is a deletion at line ~376, well clear
  of the hunks here. `NEWS.md`/`DESCRIPTION` will conflict the usual way.


beckyfisher:
## Independent review

I re-derived the diagnosis from the source rather than from the description, and it holds: four call sites share one hard-coded `gamma`/`normal`/`beta`/`uniform` table and `constant()` has no entry. Carrying the fixed value *through* the init search rather than dropping it early is the right call, and it is the conclusion I reached independently before reading the rationale.

**Verdict: merge, after finding 1.**

Environment for everything below: R 4.6.1, brms 2.23.0, rstan and cmdstanr both installed.

### Verified as fine

- `make_inits()` / `refine_inits()` / `sample_priors()` branches are correct, and the bound-rejection hang test is the right test to have written.
- **Hurdle path.** `make_good_hurdle_inits()` strips and restores the `hu` prefix, so init names come back as `b_hutop` and match `const_pars` in `add_brm_defaults()`.
- **`disp()` parameters.** `define_disp_prior()` builds them with `prior_string(nlpar = )`, i.e. class `b` with `nlpar` set, so taking the filter from `brm_args$prior` does catch them, as the comment claims.
- **Downstream methods on a real fit.** I fitted `nec3param` / Beta with `top = constant(0.5)` through `bnec()` and ran `check_priors()`, `check_chains()`, `summary()`, `ecx()` and `nec()`. All work; `ecx` goes through `posterior_epred()`, so nothing needed teaching.
- `fill_missing_priors()` round-trips the constant row unchanged.
- I expected a fixed `nec` to make `extract_pars()` return `NA`, `expand_nec()` fall through to `mod_class <- "ecx"`, and the reported NEC silently become an NSEC. It does not, because `fixef()` includes constant parameters. Worth a one-line test pinning that, since it is a silent-wrong-answer mode if brms ever changes.
- Both new test files pass. Vignette arithmetic checks out: `beta(6, 6)` sd = 0.1387; NEC interval widths 0.313 fixed against 0.344 free; and `bnec` does warn on a normalised response (`check_normalisation()`, `R/bnec.R:478`).

### 1. Blocking — `rhat()` errors on a multi-model fit with a fixed parameter

brms keeps a constant parameter in the draws as a zero-variance column, so `brms::rhat()` returns `NA` for it:

```
      beta        nec        top        phi
  1.003317   1.017656         NA   1.033616
```

`R/rhat.R:46` is `failed <- any(rhat_vals > rhat_cutoff)`, which is therefore `NA`. `rhat.bayesmanecfit()` then does `if (all(failed))` at `R/rhat.R:69`, and `if (NA)` is a hard error. Reproduced end to end with the idiom the new vignette section teaches, one model argument wider:

```r
p <- get_priors(y ~ crf(x, model = c("nec3param", "nec4param")), data = d,
                family = Beta(link = "identity"))
p <- lapply(p, function(z) { z$prior[z$nlpar == "top"] <- "constant(0.5)"; z })
fit <- bnec(y ~ crf(x, model = c("nec3param", "nec4param")), data = d,
            family = Beta(link = "identity"), prior = p, ...)

rhat(fit)
#> Error in if (all(failed)) { : missing value where TRUE/FALSE needed
```

Single-model fits do not error but return `failed = NA`, which is its own problem for anything that reads it.

Latent before this PR, since the hand-written-`init` escape hatch reached the same state, but this PR is what makes the route documented and reachable in one line. `any(rhat_vals > rhat_cutoff, na.rm = TRUE)` fixes it; dropping zero-variance parameters inside `clean_rhat_names()` is the tidier version. A two-model test with a `constant()` row would pin it.

### 2. The premise for the `add_brm_defaults()` deletion does not reproduce

The comment at `R/helpers.R:872` says "Stan does not declare a parameter whose prior is constant, so supplying an init for one is rejected."

The first half is true — `stancode()` moves `b_top` out of `parameters` into `transformed parameters`. The second half I could not reproduce. I passed an `init` list containing `b_top` to `brm()` under **both** rstan and cmdstanr and both accepted it silently:

```r
p <- prior(constant(0.5), nlpar = "top") + prior(normal(0, 5), nlpar = "beta") +
     prior(gamma(5, 1.43), nlpar = "nec", lb = 0.04, ub = 3)
ini <- lapply(1:2, function(i) list(b_top = as.array(0.5), b_beta = as.array(0.3),
                                    b_nec = as.array(1.1)))
brm(f, data = d, family = Beta(link = "identity"), prior = p, init = ini)
#> accepted (rstan)
#> accepted (backend = "cmdstanr")
```

So the block is defensive hygiene rather than a required fix. Worth keeping — it is the right thing to do — but the comment, the NEWS entry and the PR body all assert a constraint that is not currently binding, and a future maintainer will trust it. Note also that `expect_false("b_bot" %in% names(out$init[[1]]))` pins behaviour to that premise.

### 3. `!is.na(all_priors$nlpar)` is a no-op

`R/helpers.R:882`. brms records an absent `nlpar` as `""`, never `NA` — confirmed against brms 2.23.0. The guard reads as "only non-linear-parameter rows" but excludes nothing. Harmless in effect, since `paste0("b_", "")` matches no init name, but `nzchar()` is what was meant.

### 4. `constant(v, broadcast = FALSE)` is a legal brms prior that now hard-errors

`is_constant_prior()` matches it, `constant_prior_value()` cannot parse it, and the result is `stop("A constant() prior must fix a single numeric value")`. Same for `constant(1/2)`.

```
constant(0.5)                      detected=TRUE  value=0.5
constant( 0.5 )                    detected=TRUE  value=0.5
constant(1/2)                      detected=TRUE  value=ERROR
constant(0.5, broadcast = FALSE)   detected=TRUE  value=ERROR
constant(-1e-3)                    detected=TRUE  value=-0.001
```

Not a regression — both forms failed differently before — but `strsplit(..., ",")[[1]][1]` before the coercion, or `eval(parse())` on the first argument, covers both in about a line. Failing that, a test pinning the current message.

### 5. `sample_priors()` guard, minor

`plot = NULL` gives "missing value where TRUE/FALSE needed"; `plot = c("ggplot", "base")` gives "'length = 2' in coercion to 'logical(1)'". Neither is a regression, but

```r
if (!(length(plot) == 1 && (is.na(plot) || plot %in% c("ggplot", "base"))))
```

handles both. While in that file: the `@return` on `sample_priors` still says "A list containing the initialisation values", copy-pasted from `make_inits`.

### 6. Vignette prose

"Fix `top` at 0.5 and that outcome becomes indistinguishable from one centred on 0.62" sits immediately after the free fit's `fixef()`, which reports `top` = 0.487 [0.457, 0.519]. It is a counterfactual — a biased assay would have looked the same — but as placed it reads as a description of this fit.

### 7. PR description nit

"with `b_bot` absent from the draws" is not the case. brms carries constant parameters into the draws as a constant column; I confirmed `b_top_Intercept` present with a single unique value of 0.5. Only the `parameters` block omits it. That does not change the code, but it is offered as verification, and finding 1 is a direct consequence of it not being true.

### Notes, not findings

- The fence-style churn in `example3.Rmd` (```` ```r ```` to ```` ``` r ````) is knitr-version drift, the same family as the `options(width = )` point already in the description. The repo is already mixed — example1 and example6 are new-style, 2/2b/4/5 old-style — so this does not make it worse. Pinning both in `precompile.R` under #242 would.
- #193's hand-written `init` workaround for its pinned-asymptote arms can be deleted once both land. Worth a note on whichever merges second.

beckyfisher:
## Review addressed

`dev` merged, which brings in #240.

**Finding 1 is fixed there, not here.** It and finding 2 of the #240 review are
one defect seen from two sides: `posterior` returns `NA` for a zero-variance
column, and that `NA` reached `if (all(failed))` in `rhat.bayesmanecfit` as a
hard error, and `failed` in `check_sampling()` as a silent no-op. #240 excludes
zero-variance parameters from the screen rather than `na.rm`-ing them — a
parameter fixed at a known value has nothing to converge to, so "not
assessable" and "failed" never get folded into one logical. Every consumer of
those diagnostics is in that diff and none are in this one, so fixing it here
would have meant diagnostics code in a priors PR and a conflict in `R/rhat.R`.

What this PR adds is the end-to-end regression test, which needs both halves —
the fit needs the `constant()` support added here, the pass needs #240's
exclusion. It fits `nec4param` + `ecx4param` with `top = constant(3)` and
asserts `rhat()` is silent, no verdict is `NA`, `top` is absent from the
reported values, and `screen_models()` is a genuine no-op rather than a silent
one.

**2.** You are right that it does not bind — I did not re-test it, I took the
finding. The comment, the NEWS entry and the PR body now say what is actually
true: Stan moves a constant parameter into `transformed parameters`, both
backends accept and ignore an init for one, and the block is kept as deliberate
hygiene because sending `brm()` an init for a parameter Stan does not declare
is meaningless. The test that pinned the old premise now says it pins a choice.

**3.** `nzchar()`.

**4.** New `constant_one_value()`: the literal fast path first, then `parse()`
on the first argument only. `constant(1/2)`, `constant(0.5, broadcast = FALSE)`,
`constant( 0.5 )` and `constant(-1e-3)` all read; `constant(a)` and
`constant(c(1, 2))` still error. Evaluated in `baseenv()` rather than the
caller's frame — a prior is a specification, not a hook for whatever happens to
be in the session — with a test pinning that.

**5.** Your guard, and the `@return` now describes what the function returns
rather than `make_inits`'s. Tests for `NULL` and length-2.

**6.** Rewritten as the counterfactual it is: the check passes here, and the
point is what it would cost to be unable to ask. Applied to `example3.Rmd` as
well as the `.orig`, since it is prose only and the two would otherwise drift.

**7.** Corrected in the PR body — and, as you say, finding 1 is a direct
consequence of it not being true.

### On your note about #193

Agreed. Once this lands, #243's hand-written `init` list for the pinned-asymptote
arms can go. Flagging it there.

PR #250, opened 2026-08-24
Generate priors and initial values for group-level terms (#245)

### LINKED ISSUE #245: Group-level terms are unusable with bounded families: no prior is generated, and a supplied one is rejected
Three defects that compound. `ogl()`, `pgl()` and `(par | group)` cannot be fitted with a bounded family under the identity link `bnec()` forces: the default prior set omits the group-level parameters entirely, so the fit fails to initialise — and the obvious workaround, supplying the missing prior yourself, is rejected before it reaches `brm()`.

Found while authoring the #6/#33 grouping vignette, which is blocked by it.

## Reprex

Against `dev` (2.1.3.17), on shipped data:

```r
library(bayesnec)
set.seed(17)
data(nec_data)                      # y is a proportion in (0, 1)
nec_data$tank <- factor(rep(1:20, length.out = nrow(nec_data)))

# 1. no group-level prior is generated
get_priors(y ~ crf(x, "nec4param") + ogl(tank),
           data = nec_data, family = Beta(link = "identity"))
#>                       prior class nlpar    lb   ub
#>                normal(0, 5)     b  beta
#>                  beta(5, 2)     b   top     0    1
#>                  beta(2, 5)     b   bot     0    1
#>  gamma(5, 2.283131804990980)     b   nec 0.032 3.22
#>                                       <- nothing for ogl

# 2. the fit cannot start
bnec(y ~ crf(x, "nec4param") + ogl(tank), data = nec_data,
     family = Beta(link = "identity"), iter = 2000, chains = 2, seed = 17)
#> Error : Initialization failed.
#> Failed to fit model nec4param.

# 3. supplying the missing prior is rejected
p <- c(brms::prior(normal(0, 0.05), class = "sd", nlpar = "ogl"),
       brms::prior(normal(0, 0.10), class = "b",  nlpar = "ogl"))
bnec(y ~ crf(x, "nec4param") + ogl(tank), data = nec_data,
     family = Beta(link = "identity"), prior = p, iter = 2000, chains = 2, seed = 17)
#> Error: In model nec4param, user-specific parameter prior names
#>   (ogl, beta, top, bot, nec) do not match expectation (beta, bot, nec, top).
```

## Defect 1 — no group-level prior is generated

`ogl()` enters the model as an extra unbounded non-linear parameter added to the mean:

```
y ~ ogl + bot + (top - bot) * exp(-exp(beta) * (x - nec) * step(x - nec))
ogl ~ 1 + (1 | tank)
```

`get_priors()` returns rows for `beta`, `top`, `bot` and `nec` and nothing for `ogl`, so both its intercept and its group-level standard deviation fall through to the `brms` defaults. For `sd` that default is `student_t(3, 0, 2.5)`.

**On a response bounded in (0, 1) under an identity link, that is catastrophic rather than merely wide.** A tank offset drawn at that scale puts the mean outside the support, where the Beta likelihood is undefined, and Stan cannot find a valid starting point. There is no inverse-link to rescue it — which is exactly the trade `bnec()` makes deliberately, so the responsibility for keeping the mean in range falls entirely on the prior.

`prior_type = "regularizing"` does **not** help, and cannot: it changes only `top` and `bot` (`beta(5,2)` → `beta(5,1)`, `beta(2,5)` → `beta(1,5)`). There is no `sd` row in the set for it to narrow. Verified — it fails identically.

## Defect 2 — a supplied group-level prior is rejected

**Narrowed 2026-08-24 after reading the code on current `dev`.** The `sd` row is
*not* the problem — it is already dropped. `make_inits()` filters the prior to
`class == "b"` (`R/inits_functions.R:134`), which #207/#231 made deliberately
general: anything whose class is not `"b"` is out, and that already covers
`sd`. The comment there names group-level standard deviations explicitly.

What survives that filter is the **`ogl` intercept**, which carries class `"b"`
and `nlpar = "ogl"`. It then fails the exact set-equality test at
`R/inits_functions.R:143`:

```r
check_args <- identical(sort(par_names), sort(fct_args))
```

because `fct_args` holds only the curve's own arguments. That is exactly what
the observed error reports — note `sd` is absent from the listed set, because it
was already filtered:

```
(ogl, beta, top, bot, nec) do not match expectation (beta, bot, nec, top)
```

So the fix is **not** another application of the class filter. It is to exclude
group-level non-linear parameters — the `nlpar` names introduced by `ogl()`,
`pgl()` and `(par | group)` — from the set compared against `fct_args`, in the
same spirit as the `disp_pars()` precedent at `helpers.R:835`:

```r
init_priors <- brm_args$prior
disp_par_names <- disp_pars(disp_spec)
if (length(disp_par_names) > 0) {
  init_priors <- init_priors[!init_priors$nlpar %in% disp_par_names, ]
}
```

They play no part in getting the mean curve inside the response range, which is
all the initial-value search is for, so they should be filtered before the name
check and passed to `brm()` untouched.

Note the #207 gap-filling *does* work here — the warning correctly reports
filling `beta, top, bot, nec` from bayesnec defaults — and then the name check
rejects the very set it just completed.

## Defect 3 — Stan's initial value for `sd` ignores the prior

**Added 2026-08-24.** Defect 1 assumes a better prior fixes initialisation. That
assumption is not safe, and it was never tested, because defect 2 rejected the
hand-supplied prior before it reached `brm()`.

Stan draws its default initial values as `uniform(-2, 2)` **on the unconstrained
scale**. A group-level standard deviation is lower-bounded at zero, so the
transform is `exp()` and the realised initial `sd` lies in `(0.135, 7.39)`
*regardless of what prior is declared*. Under the non-centred parameterisation
`brms` uses, the group offset is `sd * z` with `z` also initialised in
`(-2, 2)`, so the initial offset can reach roughly `±15` on the response scale.

On a response bounded in (0, 1) under an identity link, that is outside the
support whatever the prior says. bayesnec supplies inits only for `class == "b"`
curve parameters, so nothing currently constrains this.

**Consequence for the fix:** a scaled default prior (defect 1) is necessary but
may not be sufficient. Resolving this issue requires *testing* whether the prior
alone lets the chain start, and generating an initial value for the group-level
parameters if it does not. That must be settled empirically before the issue is
closed, not assumed.

## Why this matters more than it looks

- **It is why the identity link "does not work" for grouped models.** [`cr_modelling_training`](https://github.com/open-AIMS/cr_modelling_training)'s `8Factor_covariates_and_groupings.Rmd` records "I had trouble getting this example to fit using the identity link, so here I am calling the family directly", and switches to `Beta()`, which is logit. That is not a stylistic shortcut — under logit an offset of ±2.5 maps harmlessly back into (0, 1), so the fit starts. But `top`, `bot` and `nec` are then logit-scale quantities, and the model is not the one the user thinks they fitted.
- **#6 has been open since 2021 for want of a worked example.** This is a plausible reason the example has been hard to produce.
- Unbounded families are unaffected: a Gaussian response has no support constraint, so a wide `sd` prior is merely wide.

## Suggested resolution

1. **Generate a group-level prior in `get_priors()`** — a `student_t(3, 0, s)`
   for the `sd` class with `s` scaled to the observed response range rather than
   fixed at 2.5, keeping the `brms` shape and heavy tail but making the scale
   appropriate to the data. Scaling by the response avoids family special-casing
   and generalises to gaussian, gamma and binomial alike. The response-scale
   machinery that already sets `top`/`bot` has what it needs.
2. **Exclude group-level `nlpar`s from the init name check** (defect 2), so a
   user-supplied prior round trips.
3. **Determine empirically whether (1) is enough to initialise**, and if not,
   generate an initial value for the group-level parameters too (defect 3).
4. Round-trip `sd` rows through `usable_prior()` / `get_priors()` so they can be
   inspected and amended, as #207 did for dispersion rows.
5. Tests: a bounded family with `ogl()` fits on defaults; a supplied `sd` prior
   is accepted and reaches the fit; an unbounded family is unchanged.

beckyfisher:
**Closes #245.** Targets `dev`. `ogl()` and `pgl()` have shipped since v2.0, so
this is a defect in released behaviour.

## The defect

`ogl()`, `pgl()` and `(par | group)` could not be fitted with a bounded family.
`get_priors()` described none of the parameters those terms add, so a
group-level standard deviation fell through to the `brms` default,
`student_t(3, 0, 2.5)`. Under the identity link there is no inverse link to
return an offset of that scale to the response range, so on a Beta, binomial or
bernoulli response the mean started outside its own support and the fit could
not initialise. Supplying the missing prior directly was rejected before it
reached `brm()`, so no workaround existed either.

## Three parts

**1. Priors.** `define_group_prior()` follows the split `define_prior()` already
makes for the curve's own parameters: a group-level standard deviation is given
one tenth of the observed range of the scale its parameter lives on.

| nlpar | prior |
|---|---|
| `top`, `bot`, `ogl` | `student_t(3, 0, diff(range(y))/10)` |
| `nec`, `ec50` | `student_t(3, 0, diff(range(x))/10)` |
| `beta`, `slope`, `d`, `f` | `student_t(3, 0, 0.5)` |

`student_t(3, 0, .)` keeps the shape of the `brms` default and changes only its
scale. Under that prior there is a probability of about 0.39 that the standard
deviation exceeds the scale, so it is weakly informative rather than tight.
`prior_type = "regularizing"` halves every scale, which it previously could not
do because there was no `sd` row in the set.

The `ogl` intercept is given `normal(0, diff(range(y))/10)`. `ogl` enters as an
offset on the whole curve, so its population intercept is confounded with `top`
and `bot`; `brms` leaves such a parameter flat. This prior is load-bearing —
see the identification note below.

**2. The initial-value name check.** The `ogl` offset carries class `"b"` and so
survived the class filter added by #207/#231, and the exact set-equality test
then rejected the whole prior set. Filtered by class for the `sd` rows and by
name for `ogl`.

**3. Initial values.** Stan draws a lower-bounded `sd` as `exp(uniform(-2, 2))`,
between 0.135 and 7.39, regardless of the prior declared, and the `ogl`
intercept is unbounded and starts anywhere in (−2, 2). `group_inits()` starts
every group-level effect at exactly zero deviation, which is valid for any
family and any link. Confirmed on `nec_data` with `ogl(tank)`, same priors and
seed: scaled priors alone give `Error : Initialization failed.`; scaled priors
plus `group_inits()` sample.

## Divergent transitions: corrected diagnosis

An earlier revision of this branch attributed the divergent transitions on a
grouped bounded fit to funnel geometry and raised `adapt_delta` on that basis.
**That diagnosis was wrong and has been reverted.** The mechanism is a support
boundary. `beta_binomial_lpmf` requires both shape arguments positive, so any
leapfrog step carrying `mu` outside (0, 1) throws and Stan discards the
trajectory:

```
Exception: beta_binomial_lpmf: Second prior sample size parameter[8]
  is -0.195321, but must be positive finite!
```

478 such rejections were logged behind 124 divergent transitions on the
reference fit. The control is decisive: the same curve, the same `ogl(tank)`
grouping and the same prior scales on a **gaussian** response give **0**
divergent transitions at the `brms` default `adapt_delta = 0.8`.

The distinction matters because the two diagnoses imply different remedies. A
funnel implies non-centring, which `brms` already uses — forcing the centred
parameterisation by editing the generated Stan code made it worse, 8.0% against
3.1% — and tighter `sd` priors, which were measured and do not help. A support
boundary implies the parameterisation of the offset, which is #257.

**The residual is not small and not constant.** Across a sweep of the number of
groups and the true group-level standard deviation, at `adapt_delta = 0.99`, it
ranged from **0.03% to 42.9%**. `NEWS.md` and `vignette("example3")` state that
range and tell the user to check.

**It is also not confined to the 0–1 bounded families.** `Gamma(link =
"identity")` and `poisson(link = "identity")` hit the same boundary at the other
end once the curve's tail sits on zero, which is where a declining
concentration-response curve spends most of its predictor range. Measured on
`nec3param` with `ogl(tank)`, 12 tanks, `adapt_delta = 0.99`, with ungrouped
controls on the identical response:

| response | curve's lower asymptote | group term | divergent | max R-hat | min ESS |
|---|---|---|---|---|---|
| Gamma | 0.5 | `ogl(tank)` | 0.38% | 1.004 | 327 |
| poisson | 0.5 | `ogl(tank)` | 0% | 1.001 | 1441 |
| Gamma | 0 | none | 0% | 1.002 | 1699 |
| poisson | 0 | none | 0% | 1.001 | 1434 |
| Gamma | 0 | `ogl(tank)` | 0.00% | **2.03** | **2.7** |
| poisson | 0 | `ogl(tank)` | **87.3%** | 1.014 | 154 |

The ungrouped rows are the control: the group-level offset is the cause. The two
families fail on different diagnostics — Poisson on divergences with an
acceptable R-hat, Gamma on a complete mixing failure with almost no divergences.

## `adapt_delta`, re-gated

`add_brm_defaults()` sets `adapt_delta = 0.99` when a group-level term is
present **and the response distribution restricts the range of `mu`** — every
family except gaussian, under an identity link. An `adapt_delta` the caller
supplied is left unchanged, as is any other `control` entry.

The gate is `mu_is_constrained()`, added to `R/validate_family.R` as the interim
form of the predicate proposed in **#256** and documented as such, so that #256
replaces it rather than adding a second copy. It tests family and link together,
which matters because `validate_family()` forces the identity link only for a
character string: `Beta` and `Beta()` both arrive on a logit link and `Gamma()`
on an inverse link.

Raising `adapt_delta` requires roughly fourteen times the gradient evaluations
per iteration (905 leapfrog steps against 64), so it is applied only where it is
needed.

## An identification problem this exposes

An additive `ogl` offset gives a zero-bounded model a free lower asymptote. For
`nec3param`, `mu = ogl_j + top·E` and `E → 0`, so the asymptote is `b_ogl + r_j`
rather than zero, and the model becomes a reparameterisation of `nec4param`.
Measured on data generated from `nec4param` with a true `bot` of 0.25, on a
predictor range long enough for the curve to plateau:

| | | WAIC | weight | fitted lower asymptote |
|---|---|---|---|---|
| ungrouped | `nec3param` | 522.23 | 0.135 | 0 (fixed) |
| ungrouped | `nec4param` | 513.85 | **0.865** | 0.224 |
| `+ ogl(tank)` | `nec3param` | 506.61 | 0.453 | **0.186** |
| `+ ogl(tank)` | `nec4param` | 506.38 | 0.547 | 0.224 |

Ungrouped, model weighting behaves as intended. Grouped, the two models are 0.23
WAIC apart and split the weight. The zero-centred prior on the `ogl` intercept
slows this — `b_ogl` reached 0.186, about two prior standard deviations from
zero — but does not prevent it. This predates the branch and is not introduced
by it, but the branch is what makes these fits possible on a bounded family.
Recorded in #257, which removes the cause.

## Changes since the earlier revision of this branch

- `dev` merged; version renumbered to `2.1.3.23`.
- `7183b4b1` reverted: the `adapt_delta` raise, its divergence table, and the
  end-to-end tests' 20% divergence bound, which is not a meaningful assertion
  against a measured range of 0.03–42.9%.
- `vignette("example3")` and `NEWS.md` rewritten off the corrected mechanism.
- `adapt_delta` reinstated on the support-based gate.
- `sd_prior_scales()` hardened: it read the last numeric argument as a scale,
  which is a rate for `gamma()` and `exponential()` and a value for
  `constant()`. A user-supplied `gamma(2, 100)` produced a starting value of 100.

## Verified

Full suite under `NOT_CRAN=true`. `devtools::document()` produces no `man/` or
`NAMESPACE` changes.

## Evidence

Simulated data throughout, brms 2.23.0, rstan 2.32.7, cmdstan 2.39.0. The
headline mechanism is `nec3param` with `beta_binomial(link = "identity")`; the
sweep varies the number of tanks (6, 12, 24) and the group-level standard
deviation (0.02, 0.04, 0.15); the replication varies the sampler seed only. The
Gamma, Poisson and identification runs are single fits. **The group effect was
generated additively on the response scale in every case**, which is the
generating mechanism the current parameterisation assumes; #257 records that its
own evidence must cover both.


beckyfisher:
## Independent review

The branch was checked out into a separate worktree and the claims tested under R 4.6.1 with brms 2.23.0. `parse_group_terms()` was exercised against all 23 models and six formula combinations (`ogl` + `pgl`, `disp()` present, an invalid parameter name, `a/b` nesting); it agrees with `add_formula_glef()` in every case. One defect is blocking. The prior-scale rule stated in the PR body, NEWS and roxygen is not the rule implemented.

### 1. `group_inits()` returns an empty list for binomial and beta_binomial

`make_standata(brms_bf, data, family = gaussian())` fails whenever the brms formula carries a `trials()` aterm:

```
Error in terms_ad(formula, family, check_response) :
  Argument 'trials' is not supported for family 'gaussian(identity)'.
```

This is the standard binomial workflow and the documented example at `R/bayesnecformula.R:215`. Reproduced at unit level (`group_inits()` warns, returns `list()`) and end to end: `bnec(y | trials(tr) ~ crf(x, "nec3param") + ogl(tank), family = binomial(link = "identity"))` emits the new "Could not determine the group-level dimensions" warning, receives no group-level initial values, and samples with 177 divergent transitions. It initialised only because Stan retries up to 100 times; part 3 of the fix did not run. Two of the three bounded families named in NEWS are therefore uncovered.

The substitution trades the Beta-with-exact-zeros-and-ones failure for a trials failure. A fallback covers both, where a swap covers neither:

```r
sdata <- try(suppressMessages(make_standata(brms_bf, data = data, family = family)),
             silent = TRUE)
if (inherits(sdata, "try-error")) {
  sdata <- try(suppressMessages(make_standata(brms_bf, data = data, family = gaussian())),
               silent = TRUE)
}
```

An end-to-end binomial fit should be added alongside the two Beta tests.

### 2. Group-level initial values are skipped silently on the `"random"` fallback

`fit_bayesnec()` guards the append with `is.list(brm_args$init)`. `make_good_inits()` returns `"random"` when it gives up (`R/inits_functions.R:323`), so the group-level initial values are dropped on precisely the datasets where the curve search already struggled, and without a message. This is inconsistent with the warning added inside `group_inits()` for a case described as "genuinely unexpected now".

### 3. The scale rule differs from the one stated

"One tenth of the scale the parameter's own prior spans" holds exactly only for `beta`, `slope`, `d` and `f`. Elsewhere the implemented quantity is one tenth of the observed data range, which is not the same thing:

| parameter, family | JSS prior sd | group `sd` scale | median group sd | ratio to prior sd |
|---|---|---|---|---|
| `top`, Beta `beta(5, 2)` | 0.160 | 0.070 | 0.054 | 0.34 |
| `top`, gaussian `normal(q90, 2.5 sd)` | 42.5 | 7.69 | 5.89 | 0.14 |
| `top`, poisson `gamma(2, .)` | 43.2 | 3.40 | 2.60 | 0.06 |
| `top`, Gamma `gamma(2, .)` | 29.8 | 5.43 | 4.16 | 0.14 |
| `nec`, `gamma(5, .)` truncated | 1.85 | 0.90 | 0.69 | 0.37 |
| `beta`, `normal(0, 5)` | 4.99 | 0.50 | 0.38 | 0.08 |

A six-fold spread across families. The choice is defensible, and arguably preferable to the stated rule, since `2.5 sd(y)` is a deliberately diffuse population prior that should not be inherited as a random-effect scale. The documentation should describe it as one tenth of the observed range of the scale the parameter lives on. For `nec` and `ec50` the stated rule does hold if "spans" is read as the truncation bounds, which is the right reading; it should be said rather than implied.

### 4. Bounded population parameters, unbounded group-level deviations

Every JSS default constrains its parameter to the region where the model is defined: `beta(5, 2)` on (0, 1), `lb = 0` for the count and Gamma families, `nec` truncated to `[min(x), max(x)]`. brms cannot bound `r_`, so the group-level deviation is unconstrained on the same scale. The property that the identity link keeps `top`, `bot` and `nec` in range therefore does not extend to a grouped fit, and no choice of `sd` prior restores it. Prior simulation for the Beta case in the PR:

| prior on the group-level sd | P(group-level mean outside (0, 1)) |
|---|---|
| brms default `student_t(3, 0, 2.5)` | 0.70 |
| this PR, `student_t(3, 0, range(y)/10)` | 0.05 |
| plus the `ogl` intercept `normal(0, range(y)/10)` | 0.08 |

This is an order-of-magnitude improvement and accounts for initialisation now succeeding. The residual 8% of prior mass outside the support means rejections during warmup remain possible, consistent with the divergences reported for the coral data. Both end-to-end tests call `suppressWarnings()`, so they would not surface them. The asymmetry warrants a paragraph in `example3.Rmd`: a reader of the JSS paper will assume the in-range property still holds.

Related: `range(x)/10` is 300 on a 0.1 to 3000 ug/L series, applied to the sd of an unbounded deviation with `student_t(3)` tails, while the population `nec` is hard-bounded to the tested range. The default is reasonable on a log-transformed predictor and loose on a raw one. The "Not done" section flags the consequence for `group_inits()`; the same magnitude enters the prior, which is the half that affects inference.

### 5. `prior_type` does not reach `define_group_prior()`

NEWS states that `prior_type = "regularizing"` could not help because there was no `sd` row to narrow. There is one now and it is still not narrowed, so the lever a user reaches for when a grouped fit misbehaves remains inert on the parameter causing the trouble. `range/20` under `"regularizing"` would close this and keep the two-tier design consistent.

### Minor

- The hurdle branch passes the raw survivor response where the non-hurdle branch passes the link-scaled one. A no-op under the identity link `bnec()` forces, but the two branches now differ without a stated reason.
- The hurdle group priors are tested at `define_prior()` level only. The justification refers to `vignette("example8")`, which does not exist on `dev` (vignettes end at example6), so this path ships without an end-to-end fit.
- `example3.Rmd` enumerates the default prior set by section and is not updated for a new class of default, and `get_priors()` output changes for every grouped model.
- `sample_priors()` filters `class == "b"` and `check_priors.bayesnecfit()` iterates `fixef()`, so an `sd` row cannot be displayed by either. The parameter the PR describes as hardest to get right is the one the package's prior diagnostic will not show. The `ogl` intercept displays correctly.

### Verified as working

- `parse_group_terms()` across all models and formula combinations, including `disp()` exclusion and silent dropping of invalid parameter names.
- The `class == "b"` and `nlpar != "ogl"` init filtering: `add_brm_defaults()` returns initial values for the curve parameters only, and the group-level priors reach `brm()` intact.
- `usable_prior()` retains the generated `sd` row (`source == "user"`) and drops the two brms defaults carrying `group = "tank"`.
- The workaround round trip. `get_priors()`, edit the `sd` row, `bnec(prior = ...)`: the fit runs and the edited `student_t(3, 0, 0.02)` appears in the fitted prior slot.

### Recommendation

Item 1 is a regression against the PR's own stated scope and should be fixed before merge. Items 2, 3 (wording) and 5 are small and belong here. The vignette text and the hurdle end-to-end test can follow separately.

beckyfisher:
## Review points addressed in 0d4c3b20

**1. `group_inits()` returned an empty list for binomial and beta_binomial.** The `make_standata()` query used `gaussian()` unconditionally, and `trials` is not a valid aterm for `gaussian`, so every fit of those two families took the error path. The fit's own family is asked first now, with `gaussian()` as the fallback for a response the family will not accept yet. Neither case can take the other's failure. Covered by a unit regression on a `trials()` formula and an end-to-end binomial `ogl()` fit.

**2. The `"random"` fallback dropped the group-level initial values silently.** `brms` takes `init` as one list per chain or as a keyword, so where the curve search gives up there is nothing to append to. `fit_bayesnec()` now says so.

**3. The scale rule is restated.** One tenth of the observed range of the scale the parameter lives on, which is what the code does. The two readings coincide only for the dimensionless parameters; for `nec` and `ec50` they agree anyway because those priors are truncated to the predictor range. Roxygen, NEWS and the PR description updated.

**4. The bounds asymmetry is documented.** A group-level deviation cannot be bounded the way `top`, `bot` and `nec` are, so a grouped fit does not inherit the in-range guarantee and no scale choice restores it. `vignette("example3")` gains a **Priors for group-level parameters** section carrying this, the rule above, and the `adapt_delta` remedy. Prose only, outside every chunk, and no vignette fits a grouped model, so the precompiled `.Rmd` was edited to match rather than re-knitted.

**5. `prior_type` now reaches `define_group_prior()`.** `"regularizing"` halves every generated scale; the degenerate 0.5 fallback is not narrowed, since it stands in for a scale that could not be measured.

**6. Minor.** The hurdle branch link-scales the survivor response, matching the non-hurdle branch. The hurdle path gains an end-to-end `hurdle_gamma` + `ogl()` fit rather than prior-level tests only.

### Verified

|---|---|---|---|---|
| full suite, `NOT_CRAN=true` | **0** | 2 | 0 | all |

The two warnings are in `test-disp_model.R:412` (a `p_waic` diagnostic) and `test-get_priors.R:239` (the declining-response message). Both are in tests this PR does not touch, and both fit models with no group-level term, so every path changed here returns before doing anything. `devtools::document()` produced no `man/` or `NAMESPACE` changes.

### Left as follow-ups

`sample_priors()` filters `class == "b"` and `check_priors()` iterates `fixef()`, so neither can display an `sd` row. Fixing that means teaching `sample_priors()` to draw from `student_t` and `check_priors()` to reach outside `fixef()`, which changes a user-facing plot for every grouped fit; it is an enhancement rather than a defect of this PR. `group_inits()` also still starts every `sd` at the median of the generated scales, as the description's **Not done** section records.

beckyfisher:
## Group-level terms on a bounded family: what changed, and what is still open

Posted as a handover. A session working on #228 could not fit a `beta_binomial` with group-level terms; this records what that turned out to be, what is fixed, and what a later session would need to carry it further. Everything below is measured on this branch, not inferred.

### It was real, and it was not `beta_binomial`

`binomial + ogl()` fails identically, so it is every bounded family under the identity link. Ruled out first: the brms 2.22.0 `beta_binomial` defect in `notes/beta_binomial_varying_trials.md` is not involved — that note's "constraint proposed but not yet applied" line is **stale**, `brms (>= 2.23.0)` is in `DESCRIPTION` on `dev`, this branch and the #228 branch. Worth correcting in that note.

Divergent post-warmup transitions, `beta_binomial` `nec3param` with `ogl(tank)`, at `bnec()`'s own `iter = 1e4, warmup = 8000`:

| variant | divergent | max R-hat |
|---|---|---|
| no group-level term | **0%** | 1.009 |
| `adapt_delta = 0.8` (the brms default, i.e. this PR before `7183b4b1`) | **46.7%** | 1.0045 |
| `adapt_delta = 0.99` (now the default) | 9.0% | 1.0041 |
| `adapt_delta = 0.999` | 8.2% | 1.0038 |
| `0.99` + `constant(0)` on the `ogl` intercept | 5.1% | 1.0020 |
| `pgl()` instead of `ogl()`, `0.99` | 4.3% | 1.0029 |

Three properties made this dangerous rather than merely annoying. **Max R-hat is 1.002–1.005 in every row including the 46.7% one**, so the usual convergence check reports a healthy fit over a posterior missing the neck of its own funnel. **Running longer does not help** — the same fit is 52% divergent at `iter = 800` and 47% at `iter = 10000`. And **every prior-side remedy failed**: `constant(0)`, a tenfold tighter intercept prior and a tenfold tighter `sd` prior all landed within 320–415/800 against the default's 415, because the geometry is a step-size problem, not a prior one.

### What changed

`7183b4b1` — `add_brm_defaults()` sets `adapt_delta = 0.99` when, and only when, the formula carries a group-level term. Ungrouped fits are untouched (it costs runtime); a caller's `max_treedepth` is preserved and augmented; a caller's own `adapt_delta` wins, including a lower one. `0.99` is the knee of the curve — `0.999` buys 0.8pp for a large runtime cost.

The four end-to-end grouped tests asserted only `ndraws > 0`, so **they would have passed at 46.7% divergent**. They now assert a divergence rate through a shared `expect_few_divergences()` helper, bounded at 20% — generous enough for the residual, tight enough to catch a return to the old regime.

`NEWS.md` and the new **Priors for group-level parameters** section of `vignette("example3")` carry the table and the two user levers.

Earlier commits on this branch, for completeness: `0d4c3b20` fixed `group_inits()` returning an empty list for binomial and beta_binomial (`trials` is not a valid aterm for `gaussian`, and the `make_standata()` query used `gaussian()` unconditionally), routed `prior_type` to `define_group_prior()`, and restated the scale rule. `6120b7b3` merged `dev` and added a `rate()` regression, `rate()` having exactly the same aterm problem as `trials()`.

### What is NOT resolved

**A grouped fit on a bounded family still carries ~4–9% divergences.** That is a mitigation, not a cure. The residual comes from putting an unbounded group-level offset on a mean that must stay in (0, 1): `brms` declares `r_` unconstrained, so a grouped fit does not inherit the in-range guarantee the curve's own bounded priors give. No choice of prior or sampler setting removes it.

**Every number here is from one simulated dataset** — `beta_binomial`, `nec3param`, 12 tanks, a deliberately weak group effect (true tank sd 0.04 against a prior scale of ~0.09). No sweep across models, group counts or datasets was run. Real data may sit either side of 9%.

### Options a later session could try

1. **Make `constant(0)` the default for the `ogl` intercept.** Halves the residual (9.0% → 5.1%). The intercept is genuinely unidentified — a constant added to `ogl` comes straight back out of `top` and `bot` — so this is arguably the more correct model. It was considered and rejected earlier in this PR on the grounds that it changes the model rather than the prior; that judgement was left standing deliberately rather than reversed, and the numbers now exist to revisit it.
2. **A logit-link escape hatch for grouped fits.** The whole problem is the identity link `bnec()` forces. Under a logit link there are no bounds to leave and the funnel is ordinary. It would cost the direct interpretability of `top`, `bot` and `nec` that the identity link exists to preserve, so it would have to be opt-in — but it is the only route that removes the residual rather than shrinking it.
3. **Steer users toward `pgl()` / `(par | group)` and away from `ogl()`.** `pgl` puts the grouping on parameters that are individually identified; `ogl` adds a free offset duplicating `top`. Measured 4.3% against 9.0%. Could be documentation, or a message when `ogl()` meets a bounded family.
4. **Sweep the parameter space before trusting the 9% figure** — vary the number of groups, the true group sd, the model (`nec4param` has a `bot` for the offset to be confounded with as well), and the family.

### Reproducing

Self-contained, no scratch files needed:

```r
library(bayesnec)
set.seed(245)
n_tank <- 12
x  <- rep(seq(0, 4, length.out = 15), each = 4)
tk <- factor(rep(seq_len(n_tank), length.out = length(x)))
tr <- sample(20:120, length(x), replace = TRUE)      # varying trials
mu <- 0.05 + (0.9 - 0.05) * exp(-exp(-0.4) * pmax(x - 2, 0)) +
      rnorm(n_tank, 0, 0.04)[as.integer(tk)]
mu <- pmin(pmax(mu, 0.02), 0.98)
p  <- rbeta(length(mu), mu * 12, (1 - mu) * 12)      # real overdispersion
y  <- rbinom(length(mu), tr, p)
d  <- data.frame(x = x, y = y, trials = tr, tank = tk)

fit <- bnec(y | trials(trials) ~ crf(x, "nec3param") + ogl(tank), data = d,
            family = beta_binomial(link = "identity"),
            iter = 1e4, warmup = 8e3, chains = 2, seed = 245)

np <- brms::nuts_params(fit$fit)
sum(np$Value[np$Parameter == "divergent__"]) / brms::ndraws(fit$fit)
```

Add `control = list(adapt_delta = 0.8)` to see the pre-fix behaviour. **Note the overdispersion**: generating with `rbinom` alone leaves `phi` unidentified and inflates the divergence count for an unrelated reason — an early version of this investigation made that mistake and its numbers were discarded.

### Verified

Full suite, `NOT_CRAN = true`: **0 failures, 0 errors**. Two warnings, `test-disp_model.R:412` and `test-get_priors.R:239`, both confirmed pre-existing by running the same files at the pre-review commit. R CMD check was green on all four platforms plus pkgdown as of `6120b7b3`; `7183b4b1` has just been pushed and its CI has not reported yet.

beckyfisher:
## Closing review

The branch has been reduced against the corrected diagnosis. This records what
changed, what was verified, and what was deliberately left for the two follow-up
issues.

### What was removed, and why

The earlier revision attributed the divergent transitions on a grouped bounded
fit to funnel geometry, raised `adapt_delta` on that basis, and documented the
result in `NEWS.md`, a new `vignette("example3")` section and a 20% divergence
bound in the end-to-end tests. `7183b4b1` has been reverted and those claims
rewritten.

The mechanism is a support boundary. `beta_binomial_lpmf` requires both shape
arguments positive, so a leapfrog step carrying `mu` outside (0, 1) throws and
the trajectory is discarded — 478 logged rejections behind 124 divergent
transitions on the reference fit. The control is decisive: the same curve, the
same `ogl(tank)` grouping and the same prior scales on a gaussian response give
0 divergent transitions at the `brms` default `adapt_delta = 0.8`.

Three claims did not survive replication and have been removed rather than
softened:

- **"A few percent typically remain."** Across a sweep of group count and true
  group-level standard deviation the measured range is 0.03% to 42.9%.
- **"`constant(0)` roughly halves the divergences."** Across three sampler seeds
  it reduces them by about a quarter, 7.7% to 5.8%, against a baseline
  seed-to-seed spread of 5.6–9.0%. The quoted 9.0% → 5.1% compared the worst
  baseline run with the best `constant(0)` run.
- **The `pgl()` over `ogl()` recommendation.** The direction replicates (4.1%
  against 7.7%) and is retained as an observation, not as advice.

### What was kept

`parse_group_terms()`, `group_inits()`, the `class == "b"` / `nlpar != "ogl"`
filtering that admits a hand-supplied group prior to the initial-value search,
the `make_standata()` family selection, `"sd"` in `auxiliary_classes()`, and the
`amend()` and `get_priors()` routing. All of it is required whatever
parameterisation the offset ends up with, and zero deviation remains the correct
starting point under any of them.

`define_group_prior()` keeps its response / predictor / log-scale split. Its
roxygen has been rewritten: the prior is what allows such a fit to initialise,
and it does not prevent the sampler reaching the boundary afterwards, because
the posterior for the standard deviation is an order of magnitude smaller than
the distance from the fitted mean to that boundary.

### What was corrected

`adapt_delta = 0.99` is reinstated on a support-based gate rather than on the
presence of a group-level term. `mu_is_constrained()` tests family and link
together, which matters because `validate_family()` forces the identity link
only for a character string: `Beta` and `Beta()` both arrive on a logit link and
`Gamma()` on an inverse link. It is documented as the interim form of the
predicate proposed in #256, which should replace it rather than add a second
copy.

The gate is correct for every family `bnec()` accepts, verified directly:
`Beta`, `beta_binomial`, `Gamma`, `poisson` and `hurdle_gamma` under identity
are raised; `gaussian` under identity, `gaussian(link = "log")` and
`Beta(link = "logit")` are left at the `brms` default.

`sd_prior_scales()` read the last numeric argument of a prior as a scale. That
is a rate for `gamma()` and `exponential()` and a value for `constant()`, so a
user-supplied `gamma(2, 100)` produced a group-level starting value of 100. It
now matches on the distribution name.

### Two findings recorded for the follow-up issues rather than fixed here

**The `(0, Inf)` identity families are affected at least as severely.** Once the
curve's lower asymptote sits on zero — the region a declining
concentration-response curve occupies over most of its predictor range —
`poisson(link = "identity")` with `ogl(tank)` reaches 87.3% divergent, and
`Gamma(link = "identity")` shows a complete mixing failure (R-hat 2.03, bulk ESS
2.7) with almost no divergences at all. Ungrouped controls on the identical
responses are clean. The two families fail on different diagnostics, so neither
statistic alone is sufficient. This is why the gate is support-based; #256
carries the predicate.

**An additive `ogl` offset gives a zero-bounded model a free lower asymptote.**
For `nec3param`, `mu = ogl_j + top·E` and `E → 0`, so the asymptote is
`b_ogl + r_j` rather than zero and the model becomes a reparameterisation of
`nec4param`. On data generated from `nec4param` with a true `bot` of 0.25,
ungrouped weighting assigns 86.5% to `nec4param`; adding `ogl(tank)` leaves the
two models 0.23 WAIC apart with the weight split 45/55, `nec3param` having
recovered an asymptote of 0.186. This predates the branch. The zero-centred
prior on the `ogl` intercept slows it — `b_ogl` reached about two prior standard
deviations from zero — but does not prevent it. #257 removes the cause; the
`check_models()` classification is deliberately left alone until then, so that
the classification and the parameterisation are corrected together.

### Verification

| | result |
|---|---|
| full suite, `NOT_CRAN = true` | **0 failures** |
| warnings | 2, both pre-existing |
| `devtools::document()` | no `man/` or `NAMESPACE` changes |
| `vignette("example3")` | prose identical in `.Rmd` and `.Rmd.orig`; no chunks touched, so no re-knit required |
| CI | 9 of 9 green — `R-CMD-check` on ubuntu release and devel, macOS and windows, on both the push and pull_request runs, plus `pkgdown` |

Note that `NOT_CRAN` is not set in `R-CMD-check.yaml`, so every `skip_on_cran()`
test is skipped on CI — including the four end-to-end grouped fits this branch
adds. CI green is therefore necessary but not sufficient here, and the local
`NOT_CRAN = true` run above is what covers them.

The two warnings are `test-disp_model.R:412` (a `p_waic` diagnostic) and
`test-get_priors.R:239` (the declining-response message). Both are in test files
this branch does not modify and both fit models with no group-level term, so
every path changed here returns before doing anything.

### Evidence, and what it does not cover

Simulated data throughout, brms 2.23.0, rstan 2.32.7, cmdstan 2.39.0. The
headline mechanism is `nec3param` with `beta_binomial(link = "identity")`; the
sweep varies the number of tanks (6, 12, 24) and the group-level standard
deviation (0.02, 0.04, 0.15); the replication varies the sampler seed only. The
Gamma, Poisson and identification runs are single fits.

The group-level effect was generated additively on the response scale in every
case, which is the mechanism the current parameterisation assumes. #257 records
that its own evidence must cover both generating mechanisms before the
transformed parameterisation is accepted.

No real dataset was used at any point.

PR #252, opened 2026-08-25
Fan out vignette precompilation, and fix the step that discarded every run (#251)

### LINKED ISSUE #251: precompile vignettes: a partial rebuild is unusable — PR step discards the run, and the job is serial
The final `create-pull-request` step lists a file in `add-paths` that only exists on a **dry run**:

```yaml
add-paths: |
  vignettes/*.Rmd
  vignettes/*.png
  vignettes/.precompile-dry-run
```

`git add` fails the entire pathspec when any entry matches nothing, so on a real run:

```
[command]/usr/bin/git add -- vignettes/*.Rmd vignettes/*.png vignettes/.precompile-dry-run
fatal: pathspec 'vignettes/.precompile-dry-run' did not match any files
```

Nothing is staged, the commit reports `no changes added to commit`, and the action exits with a bare `##[error]Unexpected error:` — no message, which is why the cause is not obvious from the summary.

**The fitting itself succeeds.** Run [32794834438](https://github.com/open-AIMS/bayesnec/actions/runs/32794834438) knitted example1 cleanly and had the full result in the working tree — `example1.Rmd`, eight regenerated figures, and two new ones — before throwing it away:

```
M  vignettes/example1.Rmd
M  vignettes/vignette-fig-exmp1-{bbinom,beta,binom,disp-bands,disp-width,gammabase,negbinbase,poisson}-1.png
?? vignettes/vignette-fig-exmp1-poisson-od-1.png
?? vignettes/vignette-fig-exmp1-rate-1.png
```

So ~35 minutes of compute is spent, correctly, and then discarded at the last step.

## Why it was not caught

The workflow has only ever been exercised as a **dry run** (`precompile/dry-run`, and the `precompile-rehearsal-2026-08-24` branch) — which is precisely the path where the marker file exists, so `git add` succeeds. The comment above `permissions:` anticipates the general shape of this ("the most expensive possible place to discover a permissions problem") but the rehearsal could not reach it.

## Fix is not quite one line

Simply deleting the marker from `add-paths` breaks the dry run, whose entire purpose is to exercise this step and which would then produce no diff and create no PR. Options:

1. Have the dry run write its marker somewhere already covered by an existing glob, e.g. `vignettes/precompile-dry-run.png`, and drop the third `add-paths` entry.
2. Write the marker on **every** run and delete it after a successful real run, before the PR step.
3. Replace `create-pull-request`'s staging with an explicit `git add` using `--ignore-unmatch`, then hand it `add-paths: .`.

(1) is the smallest change that keeps both paths honest.

## Worth adding regardless

Upload the regenerated files as a build artifact **before** the PR step. Any future failure in PR creation then costs nothing — the output is recoverable rather than lost with the runner.

Found while precompiling example1 for #224; that vignette is being regenerated locally and committed directly in the meantime. Related: #230, #215, #190, #248.

beckyfisher:
Release tier **2.1.4** (CI/housekeeping). Closes #251.

Precompiling one vignette on a runner takes about 20 minutes — `example1` was
20m37s in run 32794834438. That should make a targeted rebuild a routine thing
you reach for, and instead the practice on this stack has become a full local
rebuild of every vignette, hours per branch, before merging anything that edits
one. This PR is about closing that gap.

## The gap

Three things had to hold for a partial rebuild to be usable and none of them
did.

**1. The PR step discarded every successful run.** `add-paths` listed
`vignettes/.precompile-dry-run`, which exists only on a dry run. `git add`
stages *nothing* when any single pathspec matches nothing — verified: exit 128,
empty index — so run 32794834438 knitted `example1` correctly, had the rebuilt
vignette and ten figures sitting in the working tree, and then failed at
`git add` with a bare `##[error]Unexpected error:` and lost all of it. The dry
run is the one configuration where that file exists, so it was also the only
configuration that had ever passed: the bug was invisible to its own test.

**2. The errored-chunk guard judged vignettes it had not touched.**
`precompile.R` globbed every `^example.*\.Rmd$` in the directory rather than
the outputs of the `.Rmd.orig` files it had just knitted. The workflow's
hold-back mechanism hides `.Rmd.orig` files; it cannot hide rendered `.Rmd`
files, which have to stay in place for a partial diff to make sense. Latent
today — no committed `.Rmd` carries `#> Error` — but it meant one known-bad
vignette in the tree would fail *every* partial rebuild, at the last step,
after the compute.

**3. The job was serial.** One leg knitted every `.Rmd.orig` in sequence, so a
full rebuild was the sum (~3–4 h) and one failure aborted the lot — which is
what run 32795516512 did.

## What changed

- **`add-paths` reduced to the two globs that always match.** The dry run now
  writes a genuine PNG at `vignettes/vignette-fig-precompile-dry-run-1.png`,
  which the existing `vignettes/*.png` entry already stages. The marker lives
  *inside* `add-paths` rather than alongside it, which is the whole fix.
- **Fanned out, one job per vignette**, `fail-fast: false`. Wall clock is the
  slowest vignette rather than the sum, and a vignette that fails is confined
  to its own leg. This is also what #190 needs: the release rebuild becomes
  ~35 minutes rather than an evening.
- **Each leg uploads what it produced as an artifact before the PR step**, so
  any future failure there costs the review, not the compute. Selected with
  `git ls-files --others --modified` over the same two globs — by asking git,
  rather than guessing at the figure naming convention, which is chunk-derived
  and does not carry the vignette's name.
- **A `plan` job resolves and validates the selection.** A name with no matching
  `.Rmd.orig` now fails in seconds; previously it held back every vignette, so
  `precompile.R` knitted nothing, the guard passed vacuously, and the run
  reported success having rebuilt nothing.
- **`base` is an explicit optional input**, defaulting to the branch dispatched
  against — which is already what you want on the dispatch path. The `push`
  trigger is kept but documented as the fallback it is: it cannot choose base
  and selector independently, and it opened its PR against `precompile/<name>`
  rather than the feature branch. The PR branch moved to `precompiled/` so it
  cannot itself match the `precompile/**` trigger.
- **The guard is scoped to `file_path_sans_ext(orig_files)`** — what this run
  knitted.
- A `concurrency` group, queueing rather than cancelling: a run 40 minutes into
  fitting should not be thrown away by a second dispatch.

Fanning out also retires the hold-back/restore dance from #230 — the one keyed
off a directory because `on.exit()` never fires at `Rscript {0}` top level and
`precompile.R` opens with `rm(list = ls())`. With one vignette per leg and a
throwaway checkout there is nothing to restore, and `precompile.R` is now run as
`Rscript vignettes/precompile.R` rather than sourced into a shell step.

## The policy point

`notes/implementation/00_protocol.md` gained a paragraph, because the machinery
was only half the problem. **A precompile is not a precondition for merging.**
`dev` has no branch protection and no required checks
(`gh api repos/open-AIMS/bayesnec/branches/dev/protection` → 404), and since
\#242 the rendered vignettes are display-only markdown that `R CMD check` builds
in ~16s. The protocol already said to edit the `.Rmd.orig`, note the staleness
in the PR body, and rebuild at release under #190 — the drift into per-branch
full rebuilds happened because the cheap path did not work. It does now.

## Verification

Locally, against logic extracted from the committed YAML:

- The `plan` step over eleven trigger shapes — dispatch with no selection, one
  name, two names with whitespace, a typo (fails, exit 1), `dry_run` in three
  casings; push to `precompile/example1`, `precompile/example1,example4`,
  `precompile/all`, `precompile/dry-run`, and `precompile/nope` (fails).
- The guard, on a simulated partial rebuild: a bad `example8.Rmd` present in the
  tree but not rebuilt is correctly ignored; breaking the vignette that *is*
  being rebuilt is still caught.
- Hold-back → pack → upload → unpack → `git add` end to end in a scratch repo:
  the collect job ends with exactly the leg's own modified `.Rmd`, its modified
  figure and its new figure staged, and `git add` exits 0.
- `git add -- <existing> <nonexistent>` confirmed to exit 128 and stage nothing,
  which is the mechanism behind (1).

**On CI: dry run [32797611555](https://github.com/open-AIMS/bayesnec/actions/runs/32797611555)** — result noted in a comment below. It exercises `plan` → one
matrix leg → artifact upload → download → unpack → PR creation, i.e. every step
except the fitting itself. The throwaway PR and branch it creates should be
deleted.

**Not verified before merge:** a real fan-out. That costs ~20 minutes of runner
time per vignette and needs this file on a branch to dispatch against. The
natural first real use is the `example1` rebuild that #224/#247 wants; if it is
worth doing before merging, say so and I will dispatch it against this branch.


beckyfisher:
**Dry run [32797611555](https://github.com/open-AIMS/bayesnec/actions/runs/32797611555) passed — all three jobs green.**

The step that mattered:

```
[command]/usr/bin/git add -- vignettes/*.Rmd vignettes/*.png
Created pull request #253 (open-AIMS:precompiled/issue-251-precompile-fanout => issue-251-precompile-fanout)
```

No `fatal: pathspec` anywhere in the job — that is the fix, on the exact line that discarded run 32794834438. PR #253 contained exactly one file, `vignettes/vignette-fig-precompile-dry-run-1.png`, so the dry-run marker is picked up by the existing `vignettes/*.png` glob as intended, and the PR opened against `issue-251-precompile-fanout` — the dispatch ref, which is the base you want — from the new `precompiled/` namespace.

Everything except the fitting itself is now exercised end to end on CI:

| job | | |
|---|---|---|
| `plan` | success | 4s; resolved the dispatch to `matrix=["dry-run"]` |
| `precompile (dry-run)` | success | 3m08s; marker written, packed, uploaded as `precompiled-dry-run` (5.4 kB) |
| `open-pr` | success | 8s; artifact downloaded, unpacked, staged, PR created |

Note the wall clock the fan-out cannot help with: the three jobs ran at 01:28, 01:33 and 01:48 — about 20 minutes of *queueing* for runners, because this PR's own R-CMD-check matrix and the other open PRs were competing for them. On a real nine-vignette run that queueing is paid once per leg in parallel rather than once, so it is noise against ~20 minutes of fitting; on a one-vignette rebuild it is worth knowing about.

#253 closed and its branch deleted — it is the throwaway the dry run exists to produce.

beckyfisher:
Merged current `dev` (with #224) in. The two conflicts were the ones predicted and both mechanical:

- `DESCRIPTION` — one line. `dev` is at 2.1.3.21, so this branch takes 2.1.3.22.
- `NEWS.md` — both sides inserted at the top of the same `# bayesnec 2.1.4` heading. Union, with #224's package-behaviour entries ahead of this branch's CI entry.

Nothing else overlapped: #224 touches no workflow and not `precompile.R`; this branch touches no `R/`, no tests and no vignette sources. Net diff against the new `dev` is exactly the five files this PR is about, and #224's `rebuild precompiled example1` commit and its figures are intact in the merged tree.

Waiting on R-CMD-check.

Issue #256, opened 2026-08-26
A single statement of the permitted range of mu, and of which parameters can take it outside that range


beckyfisher:
# A single statement of the permitted range of `mu`, and of which parameters can take it outside that range

## Summary

`bayesnec` decides in at least five places whether a model is admissible for a
given response, and each place restates the rule rather than deriving it. The
rules are correct, individually. They are not written down anywhere as one
statement, they cannot be queried, and they have already drifted: #170 was opened
because `?models` and `check_models()` disagreed about the same question, and
#206 quotes a line number for the Gaussian exclusion that has since moved.

Two artefacts would replace all five, and both are needed by work already in
flight — #206 from one direction and the group-level terms of #245/#250 from the
other.

**(a) A support predicate.** One function returning the interval `mu` must lie in,
given family **and link as actually fitted**.

**(b) A model range table.** For each model in `models()`, the analytic range of
`mu` given the parameter constraints, and for each parameter, whether a deviation
on it can take `mu` outside that range.

## What the code does now

| gate | `R/check_models.R` | condition | drops |
|---|---|---|---|
| 1 | 76 | link ∈ {logit, log} | `mod_groups$zero_bounded` |
| 2 | 92 | identity ∧ {bernoulli, beta, binomial, beta_binomial} | `bounded_linear_drops()`, then `bounded_power_drops()` |
| 3 | 115 | identity ∧ hurdle family | as gate 2, applied to the mu block |
| 4 | 153 | identity ∧ {Gamma, poisson, negbinomial, zero_inflated_poisson, zero_inflated_negbinomial} | `neclin`, `neclinhorme`, `ecxlin`, `nechormepwr01` |
| 5 | 176 | `fam_tag == "gaussian"`, any link | `mod_groups$zero_bounded` |

These five gates are not all answering the same question, and separating them is
most of the work this issue asks for. There are three distinct questions:

| | question | gates |
|---|---|---|
| **Support** | can the model's mean leave the interval the likelihood defines? | 2, 3, and most of 4 |
| **Reachability** | can the mean function cover the range the link scale requires? A mean function bounded at zero cannot produce the negative values a `log` or `logit` linear predictor needs | 1 |
| **Appropriateness** | is the shape or the ceiling of the mean function meaningful for this response, irrespective of validity? | `nechormepwr01` in gate 4; gate 5 |

Gates 2 and 4 differ only in which end of the interval is at risk — above 1 for
[...]

Issue #170, opened 2026-07-30
check_models() and ?models disagree on slope-bearing equations for 0,1 bounded identity families


beckyfisher:
## Summary

`?models` and `check_models()` disagree about which equations are excluded for
0, 1 bounded families under an identity link. The documentation is the stricter
of the two. I have corrected the documentation to match the code (see the
hurdle-Gamma work), but the underlying question is worth a broader look: **the
restriction may not be needed at all**, and the code may already be more
permissive than necessary rather than less.

## The discrepancy

`?models` currently states:

> Models with a linear decay or hormesis linear increase (all models with
> parameter "slope") are not suitable for modelling families that are 0, 1
> bounded (binomial, beta, beta_binomial).

Ten equations carry a `slope` parameter:

`ecxhormebc4`, `ecxhormebc5`, `ecxlin`, `nechorme`, `nechorme4`,
`nechorme4pwr`, `nechormepwr`, `nechormepwr01`, `neclin`, `neclinhorme`

`check_models()` drops only three of them (`neclin`, `neclinhorme`, `ecxlin`) —
i.e. the linear-decay models, not the hormesis ones. The other seven remain
available.

## Why the code may be right and the docs wrong

The rationale for excluding `slope` models from 0, 1 bounded families is that a
term like `top + exp(slope) * x` is unbounded above, so `mu` can leave (0, 1)
under an identity link. But:

1. `nechormepwr01` is explicitly the 0, 1 bounded hormesis form —
   `1/(1 + ((1/top) - 1) * exp(-exp(slope) * x))` — and is bounded in (0, 1) by
   construction. A blanket "drop all `slope` models" rule would wrongly exclude
   the one equation designed for this case.
2. In a Bayesian setting an out-of-support `mu` is a rejected proposal, not an
   invalid model. It costs sampling efficiency and can make initialisation hard,
   but it does not make the fit incorrect.
3. #168 is direct evidence that initialisation is no longer the obstacle it was.
   It tested all 17 models valid for `beta_binomial(link = "identity")` across
   both prior types, including the hormesis models `nechorme`, `nechorme4`,
   `nechormepwr01`, `ecxhormebc4` and `ecxhormebc5`, and asserted that
   `make_good_inits()` finds valid inits with **all per-chain predictions
   strictly within (0, 1)**. `nechorme` in particular was previously falling back
   to Stan's random initialisation and is now reliably rescued by the new
   `refine_inits()`.

[...]

Issue #257, opened 2026-08-26
Apply group-level deviations on a scale where the mean cannot leave its support


beckyfisher:
# Apply group-level deviations on a scale where the mean cannot leave its support

*Depends on #256 (the support predicate and model range table).*

## Summary

A group-level term adds an offset that `brms` declares unconstrained. Under the
identity link `bnec()` uses, the mean it is added to is often not
unconstrained — and every leapfrog step that carries `mu` outside the
likelihood's support is rejected by Stan and counted as a divergence. Applying
the deviation on a transformed scale instead makes those excursions impossible
by construction, without changing what `top`, `bot`, `nec` and `beta` mean.

Measured against the current additive offset: **0 divergent transitions** on
every dataset and sampler seed tried, the highest effective sample size of
anything measured, and **15× fewer gradient evaluations per iteration**.

## The mechanism

`beta_binomial_lpmf` requires both shape arguments positive, so `mu` outside
`(0, 1)` throws. Running the reference fit through cmdstanr with
`show_exceptions = TRUE` logs 478 rejections behind 124 divergent transitions:

```
Exception: beta_binomial_lpmf: Second prior sample size parameter[8]
  is -0.195321, but must be positive finite!
```

`(1 - mu) * phi < 0` is `mu > 1`. The same thing happens at the other end of the
interval for the `(0, Inf)` identity families — see #256 for the
Gamma and Poisson measurements, where the effect is worse, not better.

This is not a funnel. `brms` already uses the non-centred parameterisation;
forcing the centred one by editing the generated Stan code made it worse (8.0%
against 3.1%). Nor is it a prior problem: a penalised-complexity prior, a
half-normal, an upper-bounded `sd`, `prior_type = "regularizing"` and
`constant(0)` on the intercept were all at or worse than the default. The
posterior `sd` is around 0.07 while the fitted mean sits about 0.17 from the
boundary — it is the length of the leapfrog trajectory that leaves the support,
not the width of the posterior.

## The change

Write the transform in **collapsed form, never as the literal sandwich**:

```
(0, 1):    mu = m·e^o / (1 − m + m·e^o)        # odds(mu) = odds(m) · exp(o)
(0, Inf):  mu = m·e^o
```

`logit(m)` underflows to `-Inf` once the decay term exceeds about 709, and
`inv_logit(-Inf + o)` is exactly 0, which fails the likelihood's positivity check
just as surely as `mu > 1` does. The collapsed forms are stable as `m → 0`, which
[...]

PR #259, opened 2026-08-26
State the permitted range of mu in one place (#256)

beckyfisher:
**Part of #256.** Targets `dev`. This is the first of two changes.

The two artefacts it adds change no behaviour: `check_models()` is untouched and
a new test asserts its gates agree with the table. One consumer does change.
`mu_is_constrained()`, added in #250 and now derived from those artefacts, is
what `add_brm_defaults()` uses to decide whether a grouped fit gets
`adapt_delta = 0.99`, and correcting it to ask family and link together moves 46
of the 87 family/link combinations `validate_family()` accepts — 4 of them
because a `log` link does **not** keep a 0–1 bounded mean in range, which is a
defect this PR fixes. See the `NEWS.md` entry for the enumeration.

## What this adds

**`mu_support(family, dpar)`** — the interval `mu` may occupy, as a property of
the response distribution. `(-Inf, Inf)` for gaussian; `(0, 1)` for bernoulli, beta, binomial,
beta_binomial and the `mu` block of `zero_inflated_beta`; `(0, Inf)` for Gamma,
poisson, negbinomial, the zero-inflated counts and the `mu` block of
`hurdle_gamma`. `dpar = "hu"` or `"zi"` gives the second block of a two-block
family, always `(0, 1)`.

**`mu_is_constrained(family, dpar)`** adds the link, and asks family and link
**together**. `brms` applies the inverse link before the likelihood, so what
decides it is whether the range of that inverse lies inside the support.
`Beta(link = "log")` is the case showing neither decides it alone: `exp()` is
positive but unbounded above, so on a 0-1 response it can still hand the
likelihood an invalid mean, under a link that is safe for every count family.
For a two-block family the link asked is the block's own -- `link_hu` or
`link_zi`.

`mu_is_constrained()`, added in #250 as the interim form of this predicate, is
now derived from it via `link_range()`, and its separate definition in
`R/validate_family.R` has been removed. There is one
statement, not two.

**`model_mu_ranges()`** — one row per model, with **two kinds of flag, kept
separate**:

| flag | question | models |
|---|---|---|
| `below_zero` | support: the mean is unbounded below | `neclin`, `neclinhorme`, `ecxlin` |
| `unscaled_excess` | support: the mean can exceed one through a term with **no coefficient** | `nechormepwr`, `nechorme4pwr` |
| `zero_asymptote` | reachability: the mean decays onto zero, so it cannot produce the negative values a `log`/`logit` linear predictor needs | the 11 in `mod_groups$zero_bounded` |
| `ceiling_at_one` | appropriateness: the mean saturates at exactly 1 by construction | `nechormepwr01` |

## Why two kinds of flag rather than one

`check_models()` excludes `nechormepwr01` from the count and Gamma families.
On support grounds it is safe there, so this looks like a discrepancy. It is
not. The hormetic term is
`1 / (1 + ((1 / top) - 1) * exp(-exp(slope) * x))`, which saturates at exactly
1:

| `top` | factor at x = 0, 0.5, 1, 2, 4, 10 |
|---|---|
| 0.3 | 0.300 → 0.414 → 0.538 → 0.760 → 0.959 → 1.000 |
| 20.0 | 20.000 → 2.360 → 1.537 → 1.148 → 1.018 → 1.000 |

For `top` below 1 it rises towards 1, which is the intended hormetic increase.
For `top` above 1 it *falls* towards 1, expressing a decline where hormesis is
intended, and cannot represent a mean above `top`. `mu` stays positive
throughout, so the model is valid and unsuitable at the same time. **A
support-only table would have reinstated it for the count families and
introduced a defect.**

Two further entries contradict the abbreviated rules in circulation, and are
recorded in the roxygen:

- **`slope` is exponentiated in all ten models that carry it**, so a deviation
  on it can never make `exp(slope)` invalid. It sets a *level* — the height of
  the hormetic peak, or the rate of linear decline — and that is what leaves the
  interval. "Exponentiated, therefore safe" is the wrong test.
- Of the 21 equations carrying **`beta`**, it is harmless in 20 and harmful in
  `neclinhorme` alone, where it enters as a subtractive linear term rather than
  inside a decay factor bounded in (0, 1]. It is harmless in exactly one of the
  ten equations carrying `slope`, `nechormepwr01`.

## Nothing derives from the table yet, and that is the point

`check_models()` is unchanged. A new test asserts that its gates agree with
the table, for all twelve families in `mod_fams`:

| gate | derived set |
|---|---|
| identity ∧ {beta, binomial, bernoulli, beta_binomial} | `below_zero ∪ unscaled_excess` |
| identity ∧ {Gamma, poisson, negbinomial, zero_inflated_*} | `below_zero ∪ ceiling_at_one` |
| link ∈ {log, logit} | `zero_asymptote` |
| `fam_tag == "gaussian"` | `zero_asymptote` |
| identity ∧ two-block family | union over both blocks |

All reproduce exactly, so the table alters no model set. A later change makes
the gates read the table, and any difference found then is a decision rather
than a regression — which is the staging recorded on #256. The behaviour that
does change is `mu_is_constrained()`'s, described above and in `NEWS.md`.

The agreement test is what stops the table becoming a second source of truth.
#170 was opened because `?models` and `check_models()` had drifted apart; a
table written down and never checked is the same failure with an extra file.

## The table is regenerated from the formulas, not trusted

`test-mu_support.R` evaluates each model's mean over a grid of parameter values
inside the constraints a 0-1 bounded family imposes, and reads the flags back:

- `below_zero` is asserted to be exactly which models produce a negative mean;
- `unscaled_excess` is asserted by driving `slope` to −25, so `exp(slope)` is
  effectively zero and any excess carrying a coefficient disappears — what still
  exceeds 1 is the excess with none. The four coefficiented hormesis models are
  separately asserted to exceed 1 at `slope = 2`, so the distinction is shown to
  be necessary rather than assumed;
- `zero_asymptote` is derived structurally (no `bot`, and not unbounded below),
  asserted numerically as a ratio at a large predictor value, and asserted to
  reproduce `mod_groups$zero_bounded`;
- `ceiling_at_one` is asserted through the factor table above.

`mu_support()` is asserted for every family in `mod_fams`, so a family added
there without a decision here is a test failure rather than a silent fallback.

## Deliberately not in this change

**The link policy.** #256 records the decision that `validate_family()` should
assign the identity link where the user did not choose one, honour an explicit
`identity`, `log` or `logit`, and refuse anything else — because `Beta` and
`Beta()` currently both arrive on a logit link and `Gamma()` on an inverse link,
silently. That is a genuine behaviour change touching several entry points into
`validate_family()` and changing what `bnec()` fits, rather than how hard it
samples, so it is kept separate. It follows immediately, stacked on this branch.

Until it lands, `mu_support()` gives the correct answer for the family object it
is handed; whether that object carries the link the user intended is the other
change's problem.

## Observed while doing this, not fixed here

`validate_family("beta")` fails with `unused argument (link = "identity")`,
because `mod_fams` maps the family tag `beta` to the constructor name `Beta` and
`get("beta")` resolves to `base::beta()`. The tag is the one brms reports and
the one the package uses internally, so a user reading a fitted object's family
and passing it back gets a confusing error. One line in the link-policy change.

## Verified

`test-mu_support.R`: 145 assertions, all passing. `devtools::document()`
produces no `man/` or `NAMESPACE` changes, every function being `@noRd`.
Full suite and CI below.


beckyfisher:
## Review round 1 addressed in 810870b6

Eight of the nine findings are accepted and fixed. The ninth is accepted and
deferred with a reason. One finding led somewhere the suggested fix did not.

### 1. Two-block families and gate 3 — fixed

Correct, and the PR body was wrong to claim "every family". `mu_support()` gains
a `dpar` argument: `"mu"` is the block the curve's parameters live on, `"hu"`
and `"zi"` are probabilities and are on (0, 1) whatever the `mu` block is.

`hurdle_gamma` is the case that needs it — (0, Inf) on `mu` and (0, 1) on `hu`
at once — and its drop set is now derived as the union of what each block would
drop alone, which is exactly the three-way union observed. `zero_inflated_beta`
is (0, 1) on both blocks, so the union collapses and it behaves as an ordinary
bounded family. Both are in the agreement test, and a further test asserts the
covered set equals `unname(mod_fams)`, so a family omitted in future fails
rather than passing silently.

### 2. Non-identity links — fixed, and the diagnosis was deeper than the finding

The finding is right that `probit`, `cloglog`, `cauchit` and `sqrt` got the
wrong answer. The cause is not that `mu` is the linear predictor under those
links; it is that **brms applies the inverse link before the likelihood**, so
the mean is valid by construction whatever the sampler proposes. Confirmed from
the generated Stan code:

```
identity : target += binomial_lpmf(Y | trials, mu);
logit    : target += binomial_logit_lpmf(Y | trials, mu);
probit   : mu = Phi(mu);
cloglog  : mu = inv_cloglog(mu);
```

So the whitelist was the wrong shape. The function is now factored into two
questions: `mu_support()` is a property of the response distribution and does
not depend on the link at all, and `mu_is_constrained()` adds the link.

That also corrects a case the finding did not raise. **`inverse` is constrained**,
not guaranteed — `Gamma(link = "inverse")` emits `mu = inv(mu)` then
`gamma_lpdf(Y | shape, shape ./ mu)`, and `inv(eta)` is negative wherever `eta`
is. Only `identity` and `inverse` fail to guarantee; an unrecognised link is
treated as not guaranteeing, which is the conservative direction since the only
consequence is raising `adapt_delta` where it may not be needed.

### 3. The `unsafe` column — withdrawn rather than extended

The counterexamples are correct and reproduce:

| | | max mu |
|---|---|---|
| `nechorme`, `top` 0.9, `slope` log(0.05), `beta` 0 | `nec` = 1 | 0.950 |
| | `nec` = 4 | **1.100** |
| `ecxhormebc4`, `top` 0.9, `slope` log(0.05), `beta` log(2) | `ec50` = 1 | 0.793 |
| | `ec50` = 5 | **1.032** |

Before adding `nec` and `ec50` I wrote the completeness test the finding asks
for, deriving the column by perturbing each parameter from a fixed baseline. It
disagreed with both the table and the suggested correction:

```
model          derived                            table
neclin         slope,top                          nec,slope,top   <-- DIFFERS
nechormepwr    beta,nec,slope,top                 slope,top       <-- DIFFERS
nechorme4pwr   beta,bot,nec,slope,top             bot,slope,top   <-- DIFFERS
```

Two distinct problems. `neclin` lost `nec` because the baseline `slope = -5`
makes the decline too slow for any `nec` to drive the mean below zero — with
`slope = 0` it does. And `nechormepwr` gained everything because its baseline is
*already* outside (0, 1), which is what `unscaled_excess` means, so perturbing
any parameter trivially leaves the range.

**The column is not a per-parameter property.** Whether a deviation on `nec`
takes `nechorme` above 1 depends on where `slope` and `top` are, so any such
column needs an existential definition over the remaining parameters, and a
derivation test that searches for an in-range baseline first — which does not
exist for two of the models. Shipping a column that took three attempts to
define, that nothing consumes and that nothing tests is worse than not shipping
it. It is withdrawn, the observations that motivated it are kept as prose in the
roxygen, and #257 is recorded as the place it belongs, since that is where it
would be consumed and where the definition can be tied to what the transform
needs.

### 4. `ceiling_at_one` not regenerated — fixed

Correct: it was the one flag verified against a hand-written copy of the term.
It is now driven through `eval_mu("nechormepwr01", ...)` with `nec` placed
beyond the evaluated predictor range so the decay factor is 1 and the hormetic
term is what remains. Editing `bf_nechormepwr01` now breaks it. The test also
asserts `mu > 0` throughout, which is the support-versus-appropriateness
argument itself, and that no `slope` value takes the factor above 1.

### 5. The "every family" test was vacuous — fixed

Correct; `c(0, Inf)` satisfied both assertions, so the fallback the comment
claimed to guard against would have passed. Replaced with an expected-value list
keyed on the family tag plus a set-equality assertion against
`unname(mod_fams)`. `zero_inflated_negbinomial` now has an explicit expectation.

### 6. `below_zero` is conditional on a non-negative predictor — fixed

Stated in the roxygen under a new **Scope** heading, together with the
observation that `check_models()` has a sixth gate for a predictor containing
negative values which the table does not yet describe.

### 7. Two prose claims — fixed

`slope` is unsafe in nine of the ten equations carrying it, `nechormepwr01`
being the exception because its factor is bounded by `max(top, 1)`; the test now
asserts that directly. And 21 equations carry `beta`, of which 20 are safe and
`neclinhorme` is not — `neclin` and `ecxlin` have no `beta` parameter. Both
corrected in the roxygen and in the PR body.

### 8. `unscaled_excess` criterion — fixed

Stated in the roxygen as specific to a (0, 1) bounded response rather than
against the model's own level, which is what the assertion tests. The thirteen
equations without `slope` are now asserted too, rather than left untested on the
grounds that they are obviously fine.

### 9. Housekeeping — two fixed, one deferred

The duplicate `mu_is_constrained` test has been removed from
`test-inits_functions.R`, with a pointer left to where the coverage now lives.
The list-column naming issue is moot, the column having been withdrawn.

The open-versus-closed interval point is accepted and **not** fixed. `mu_support()`
documents open intervals and returns closed endpoints with no way to distinguish
them. Nothing compares against the endpoints yet, so the change would be
untestable here; it belongs with #257, which is the first consumer that would
compare. Recorded there rather than left implicit.

### Verification

`test-mu_support.R`: 145 assertions, all passing. Full suite and CI below.

beckyfisher:
## Review round 2 addressed in d91ac5b6

All eleven findings accepted. Finding 1 was a genuine defect in the round-1 fix
and its correction changes the design; the rest follow from it or are prose.

### 1. BLOCKING — `unguarded_links()` treated guarantee as a property of the link alone. Fixed.

Correct, and the counterexample reproduces:

```
Beta/logit  : mu = inv_logit(mu);  target += beta_lpdf(Y | mu .* phi, (1 - mu) .* phi);
Beta/log    : mu = exp(mu);        target += beta_lpdf(Y | mu .* phi, (1 - mu) .* phi);
```

`exp(eta)` is positive but unbounded above, so on a 0–1 response a `log` link
hands `beta_lpdf` a negative second shape parameter — while guaranteeing a valid
mean for every count family. Neither the family nor the link decides it.

Replaced with the containment test the finding proposes. `link_range()` gives
the interval an inverse link maps the linear predictor into, and
`mu_is_constrained()` is true when that interval is not inside
`mu_support()`. The family/link split survives, as predicted:

| | | |
|---|---|---|
| `Beta` + `logit`, `probit`, `cloglog`, `cauchit` | (0,1) ⊆ (0,1) | guaranteed |
| `Beta` + `log` | (0,Inf) ⊄ (0,1) | **constrained** |
| `Gamma` + `log`, `sqrt` | (0,Inf) ⊆ (0,Inf) | guaranteed |
| `Gamma` + `inverse` | (−Inf,Inf) ⊄ (0,Inf) | constrained |
| any + `identity` | (−Inf,Inf) | constrained unless the support is unbounded |

### 2. SHOULD FIX — the "conservative" claim was backwards. Fixed by the same change.

Correct: the old code returned `FALSE` for anything outside
`c("identity", "inverse")`, so an unrecognised link suppressed the raise. Under
containment an unrecognised link returns `c(-Inf, Inf)`, which never lies inside
a bounded support, so it now reports the mean as reachable. The prose and the
code agree, and a test asserts it for a fabricated link on both a bounded and an
unbounded family.

### 3. SHOULD FIX — `dpar = "hu"` asked the wrong link. Fixed.

Correct. `mu_is_constrained()` now reads `family[[paste0("link_", dpar)]]` for a
non-`mu` block. A test covers a `hurdle_gamma` with `link = "log"`: the `mu`
block is unconstrained, the `hu` block is still constrained, and reading
`family$link` for both would have answered the `mu` question twice.

### 4. SHOULD FIX — not behaviour-neutral. Fixed by declaring it.

Correct, and the delta is larger than the finding states and runs in both
directions. Measured against the version merged in #250, over every family and
link `validate_family()` accepts:

- **2 combinations `FALSE` → `TRUE`**: `binomial(link = "log")` and
  `Beta(link = "log")`. These are the defect in finding 1 — they now receive the
  raise and previously did not.
- **18 combinations `TRUE` → `FALSE`**: `probit`, `cloglog`, `cauchit` and
  `sqrt` across the bounded and count families. No proposal under those links
  can reach an invalid mean, so the raise was costing gradient evaluations for
  nothing.

`NEWS.md` now carries this as its own entry describing a behaviour change with
the reason, rather than under "No behaviour change". Grouped fits only;
ungrouped fits are unaffected in every case.

### 5. SHOULD FIX — `NEWS.md` described the withdrawn column. Fixed.

The entry is rewritten. The withdrawn column, the superseded "against its own
level" wording and the "five gates" count are all gone.

### 6. SHOULD FIX — the superseded explanation survived in the consumer. Fixed.

Correct, and the more serious of the prose findings, since it is the failure
mode this work exists to prevent. `R/helpers.R` and
`test-inits_functions.R` both said "under a log or logit link `mu` is the linear
predictor and is likewise unconstrained". Both now state the containment rule
and name `Beta(link = "log")` as the case that shows "any link other than
identity is safe" to be false.

### 7. MINOR — the Scope section declared per-parameter safety unwritable and then made two per-parameter claims. Fixed, and the withdrawal argument restated.

The assessment is accepted: an existential definition is well formed, and
restricting the domain to equations that survive the bounded gate dissolves the
empty-baseline objection, since `nechormepwr` and `nechorme4pwr` are exactly the
two that gate excludes. The roxygen no longer argues the column cannot be
defined. It is withdrawn on the ground given last — nothing consumes it, so it
cannot be tied to a requirement — and #257 is named as where it belongs.

The two observations are kept and are now **asserted over the admissible set**
rather than left as prose. Of the five slope-bearing equations admissible on a
(0, 1) response, a deviation on `slope` alone takes the mean above 1 in four and
cannot in `nechormepwr01`; a deviation on `beta` alone leaves the mean in range
in every admissible equation; `neclinhorme` is the exception and is excluded on
`below_zero`; and 21 of 23 equations carry `beta`, `neclin` and `ecxlin` being
the two without.

### 8. MINOR — gate 1 tests only `logit` and `log`. Recorded.

Correct: the reachability argument applies equally to `probit`, `cloglog` and
`cauchit`, and `dropped(Beta(link = "probit"))` is empty. The table encodes the
existing gap. Recorded in the Scope section alongside the sixth-gate note rather
than fixed, since changing which models `check_models()` drops is the behaviour
change this PR is staged to avoid.

### 9. MINOR — the negative-predictor note joined two disjoint sets. Fixed.

Correct and confirmed. The Scope section now states both sets separately: four
hormesis equations return a negative mean at a negative predictor and **no gate
excludes them**; the negative-predictor gate drops a different four, for a
different reason — a fractional power of a negative base being undefined.

### 10. MINOR — `expected_supports()` keys. Fixed.

Correct; the keys are the `mod_fams` values, which is the form
`validate_family()` accepts. The comment says so, and notes that the tag `beta`
differs from the constructor `Beta` and that `validate_family("beta")` errors,
which is recorded on #256 and is not this test's subject.

### 11. MINOR — `mu_support(NULL, dpar = "hu")`. Fixed.

The null guard now precedes the `dpar` branch.

### Verification

`test-mu_support.R`: 182 assertions, all passing. `devtools::document()`
produces no `man/` or `NAMESPACE` changes. Full suite and CI below.

beckyfisher:
## Review round 3 addressed in ed18cee1

All findings accepted. **Three of them are places where the round-2 response
claimed a fix that had not been made.** That is worse than the defects
themselves in a PR whose purpose is to stop prose and code drifting apart, and
the cause was that I edited by string replacement without asserting the match.
Every edit in this round asserts, and each fix below was verified after applying
it rather than at the point of writing.

### SHOULD FIX 1 — `link_range()` omitted `softit` and `squareplus`. Fixed.

Correct. Both are now in the switch. Rather than extend it by inspection again,
the enumeration is asserted: a new test constructs every family in `mod_fams`
against thirteen candidate links, collects those accepted, and requires that the
only ones falling to the `c(-Inf, Inf)` default are `identity`, `inverse` and
`1/mu^2`. A link added to brms that any accepted family takes will fail that
test rather than being silently misclassified.

`1/mu^2` falling to the default is correct and is now stated: its inverse is
`eta^(-1/2)`, undefined for a non-positive linear predictor.

### SHOULD FIX 2 — the `dpar` fix had no discriminating test. Fixed.

Correct, and verified: reverting `mu_is_constrained()` to read `family$link`
unconditionally left all assertions passing. `hurdle_gamma` cannot discriminate
because its `mu` link is never one that maps into (0, 1).

The case suggested is now in the file. Re-running the same reversion against it:

```
── 1. Failure ('test-mu_support.R:103:3'): a two-block family is asked about the
```

### SHOULD FIX 3 — the superseded rule survived in `test-inits_functions.R`. Fixed.

Correct, and the round-2 response was wrong to say otherwise. The replacement
targeted a lowercase "an unconstrained mean" while an earlier edit in the same
round had capitalised it, so the substitution matched nothing and I did not
check. The comment now states the containment rule and names
`Beta(link = "log")` as the case disproving "any link other than identity is
safe".

### SHOULD FIX 4 — the behaviour delta did not reproduce. Restated from a full enumeration.

Correct on both counts, and the omission was in my enumeration: it varied
`link` only, and held a fixed list of constructors that excluded `bernoulli` and
the zero-inflated families. `zero_inflated_beta(link = "log")` alone is refused
by `validate_family()`, which requires `link_zi = "identity"`; supplied together
it is accepted and does change.

Re-measured over **87 combinations** — every family in `mod_fams` against every
link it constructs, with `link_hu` and `link_zi` at the identity `bnec()`
requires:

- **4 `FALSE` → `TRUE`**: `binomial(link = "log")`, `bernoulli(link = "log")`,
  `Beta(link = "log")`, `zero_inflated_beta(link = "log")`.
- **42 `TRUE` → `FALSE`**, across `probit`, `probit_approx`, `cloglog`,
  `cauchit`, `softit`, `sqrt`, `softplus` and `squareplus`.

`NEWS.md` now carries the counts with the enumeration that produced them, since
as the finding notes a bare count depends on the constructor path.

The PR body's "deliberately behaviour-neutral" opening is replaced. It now
separates the two artefacts, which alter no model set, from `mu_is_constrained()`,
which is a consumer and does change.

### MINOR — all fixed

- The gate-1 link reach is now genuinely in the Scope section, as a fourth item:
  `zero_asymptote` states a property of the mean function while the gate
  consuming it tests `logit` and `log` only, so `probit`, `cloglog`, `cauchit`
  and `softit` drop nothing. Left unfixed deliberately, since closing it changes
  which models a fit uses.
- The duplicated `Gamma(link = "log")` assertion is replaced by the two
  remaining `log` cases, `bernoulli` and `zero_inflated_beta`.
- `sqrt` is documented as `[0, Inf)`, reaching zero at `eta = 0`, which is
  outside the open support the count and Gamma families require though on a set
  of measure zero.
- The `is.null(link)` branch for a `dpar` the family does not carry now says so
  and why returning `TRUE` is the conservative answer.
- The blank line before the following `test_that()` is restored.

### On the round-1 deferral being stale

Noted and agreed: `mu_is_constrained()` does now compare against endpoints, so
the open-versus-closed question is live rather than deferred. `sqrt` is the only
link whose range touches an endpoint, the closed comparison gives the right
answer, and it is documented above. If #257 derives anything else from the
interval the distinction will need representing properly rather than by comment.

### Verification

`test-mu_support.R`: 187 assertions, 0 failed. `test-inits_functions.R` and
`test-check_models.R` re-run clean. `devtools::document()` produces no `man/` or
`NAMESPACE` changes. Full suite and CI below.

beckyfisher:
## Resume state — review cycle paused after round 3

Recorded here rather than in a notes file so that the state and the work are in
the same place. Nothing is running; the machine can be rebooted.

### Where the cycle stopped

Three review rounds have run, each in a separate session, each with a response
comment above. **Round 4 was not launched.** The cycle limit agreed for this
work is five rounds, or until findings are resolved, or until something needs a
user decision.

Round 3's recommendation was "not yet, but close", contingent on two prose
findings — a stale comment asserting the rule this PR disproves, and a PR body
claiming behaviour neutrality that the NEWS entry contradicted. Both are fixed
in `ed18cee1`, along with the two code findings (`link_range()` omitting
`softit` and `squareplus`, and the `dpar` fix having no discriminating test).

### State at `ed18cee1`

| | |
|---|---|
| CI | 9 of 9 green on this commit |
| `test-mu_support.R` | 187 assertions, 0 failed |
| `test-inits_functions.R`, `test-check_models.R` | re-run clean |
| `devtools::document()` | no `man/` or `NAMESPACE` change |
| working tree | clean, pushed |
| **full suite under `NOT_CRAN=true`** | **outstanding — see below** |

### The one outstanding verification

The full suite has **not** been run against this commit. An earlier run was
started when the PR was opened and is invalid: `testthat::test_dir()` reads test
files at run time while `devtools::load_all()` had loaded the pre-round-2 code,
so it reported nine failures in `test-mu_support.R` that are entirely that
mismatch. The log has been deleted rather than left to be misread.

This matters because `R-CMD-check.yaml` does not set `NOT_CRAN`, so CI skips
every `skip_on_cran()` test. CI green is necessary but not sufficient.

To resume:

```
cd /mnt/c/Rworking/wt-256
NOT_CRAN=true nohup Rscript -e 'suppressMessages(devtools::load_all(".", quiet = TRUE)); testthat::test_dir("tests/testthat", reporter = "summary", stop_on_failure = FALSE)' > ignore/test_full.log 2>&1 &
```

Two pre-existing warnings are expected and are not this branch's:
`test-disp_model.R:412` and `test-get_priors.R:239`.

### Next steps, in order

1. Run the full suite above. If clean, launch review round 4.
2. If round 4 returns nothing substantive, merge #259.
3. Open the link-policy PR, stacked on `dev` after #259 merges. It is the second
   half of #256 and is deliberately separate because it changes what `bnec()`
   fits rather than how hard it samples. Scope is recorded in the #259 body
   under **Deliberately not in this change**, and the decision itself is on #256
   under **(a) The support predicate**: assign identity where the user did not
   choose a link, honour an explicit `identity`, `log` or `logit`, refuse
   anything else. `?bnec` documents the opposite policy and needs rewriting with
   it. One line in that change also fixes `validate_family("beta")`, which
   currently errors with `unused argument (link = "identity")` because
   `mod_fams` maps the tag `beta` to the constructor `Beta`.
4. Then #257, the transform, stacked on the finished foundation.

### Worktrees

`/mnt/c/Rworking/wt-256` is this branch. `/mnt/c/Rworking/wt-245` still holds
the merged #245 work and can be removed with `git worktree remove` once nothing
else needs it. Both carry an untracked `ignore/` directory, which is
gitignored; the fitting scripts and result objects behind the #245 measurements
are in `wt-245/ignore/scratch-250/`.

beckyfisher:
## Review round 4 addressed in dc51e5d5

All four round-3 items were verified fixed by the reviewer against file contents
and by re-deriving the evidence, which is the right way to check them given that
rounds 2 and 3 both contained claimed fixes that had not landed. The independent
reproduction of the behaviour delta — 87 accepted combinations, 4 `FALSE`→`TRUE`,
42 `TRUE`→`FALSE`, with the 42 splitting as cauchit 7, cloglog 7, probit 7,
sqrt 5, probit_approx 4, softit 4, softplus 4, squareplus 4 — matches `NEWS.md`
combination for combination.

### SHOULD FIX — two sentences in the PR body stated the rule this PR disproves. Fixed.

Correct, and both were pre-round-2 text that survived the round-3 body edit
because that edit targeted the opening and the derivation section, not
**What this adds**.

- "a proposal can only be invalid under `identity` … or `inverse`" is replaced
  by the containment statement, with `Beta(link = "log")` named as the case
  showing neither family nor link decides it alone.
- "the coarsest reading of it — `any(is.finite(mu_support(family)))`" is
  replaced by "derived from it via `link_range()`", which is what the
  implementation does.

The body is the merge record, so this is worth correcting even though no file
changes.

### MINOR — "Three restrictions" followed by four items. Fixed.

The round-3 edit added the gate-reach item without updating the lead-in.

### MINOR — the enumeration claim was stronger than the test. Fixed.

Correct. The roxygen said the enumeration was "every link the twelve families in
`mod_fams` accept, obtained by constructing each", while the test intersects
family acceptance with a fixed thirteen-element candidate list. The text now
says exactly that, and records that the list is complete for brms 2.23.0 but
that a link added to an existing family in a later version would be missed
rather than caught, since brms does not expose the accepted set
programmatically.

Engineering a programmatic enumeration was considered and rejected on the
reviewer's reasoning: there is no supported way to obtain it, and the failure
mode is a link being reported constrained when it need not be, which costs
gradient evaluations rather than correctness.

### Items correctly identified as not findings

Recorded here so they are not re-raised. The `sqrt` open/closed question is
handled — `link_range("sqrt")` returns `c(0, Inf)`, the containment test admits
it, and the roxygen records that the image is `[0, Inf)` and reaches zero on a
set of measure zero. The `hu`/`zi` branch is exercised only by tests because the
single production caller passes the default `dpar`; whether a grouped fit on a
hurdle block should consult the second block's support is unchanged from #250
and belongs to the follow-on. The gate-1 link reach is recorded in Scope and
deferred, since closing it changes which models a fit uses.

### Verification

`test-mu_support.R`: 187 assertions, 0 failed. `test-check_models.R` and
`test-inits_functions.R` re-run clean. `devtools::document()` produces no `man/`
or `NAMESPACE` change. The full suite under `NOT_CRAN = true` is running against
this branch and will be reported before merge, since `R-CMD-check.yaml` does not
set `NOT_CRAN` and CI therefore skips every `skip_on_cran()` test.

beckyfisher:
## Merging

Four review rounds, each in a separate session. Round 4 verified the round-3
fixes against file contents and independently reproduced the behaviour delta,
and recommended merge.

| | result |
|---|---|
| full suite, `NOT_CRAN = true` | **0 failures** |
| warnings | 2, both pre-existing (`test-disp_model.R:412`, `test-get_priors.R:239`) |
| CI on `dc51e5d5` | **9 of 9 green** |
| `test-mu_support.R` | 187 assertions |
| `devtools::document()` | no `man/` or `NAMESPACE` change |

The full-suite run matters independently of CI: `R-CMD-check.yaml` does not set
`NOT_CRAN`, so CI skips every `skip_on_cran()` test, including the end-to-end
grouped fits #250 added.

### What the review cycle changed

Recorded because two of the four rounds found errors in the previous round's
fix, which is the argument for having run it.

- **Round 1** established that the `unsafe` column was untested and incomplete.
  Extending it was attempted, and the derivation showed the column is not a
  per-parameter property; it was withdrawn and recorded against #257.
- **Round 2** found the link rule wrong. The round-1 fix named `identity` and
  `inverse` as the only unguarded links, which is false for `Beta(link = "log")`
  — `exp()` is positive but unbounded above, so a log link does not keep a 0-1
  bounded mean in range. Replaced with a containment test.
- **Round 3** found `link_range()` still missing `softit` and `squareplus`, the
  `dpar` fix untested, and three places where the round-2 response claimed a fix
  that had not landed. The cause was string replacement without asserting the
  match; every edit since asserts.
- **Round 4** verified all of the above and reproduced the behaviour delta to
  the combination.

### Follow-on

#256 is **not** closed by this. Its second half — the link policy — follows in a
stacked PR: assign identity where the user did not choose a link, honour an
explicit `identity`, `log` or `logit`, refuse anything else. That change also
fixes `validate_family("beta")`, which currently errors with
`unused argument (link = "identity")` because `mod_fams` maps the tag `beta` to
the constructor `Beta`.

Three items are deliberately deferred and are recorded in the `model_mu_ranges()`
Scope section rather than left implicit: the sixth `check_models()` gate for a
negative predictor, the gate-1 link reach (`zero_asymptote` is consumed against
`logit` and `log` only, so `probit`, `cloglog`, `cauchit` and `softit` drop
nothing), and the open-versus-closed interval question, which becomes live if
#257 derives anything further from the returned interval.

PR #260, opened 2026-08-27
Assign the link unless the caller chose one (#256)

beckyfisher:
**Completes #256.** Targets `dev`, stacked after #259 which landed the support
predicate and the model range table.

This is the behaviour change #256 records under **(a) The support predicate**,
and it is deliberately separate from #259 because it changes what `bnec()` fits
rather than how hard it samples.

## The defect

Which link a fit used depended on how the family was written, and the difference
was silent:

| supplied as | link used before | link used now |
|---|---|---|
| `"Beta"` | identity | identity |
| `Beta` | **logit** | identity |
| `Beta()` | **logit** | identity |
| `Gamma()` | **inverse** | identity |
| `Beta(link = "logit")` | logit | logit |
| `Beta("logit")` | **logit, unread** | logit, read |
| `binomial(link = "probit")` | probit | **refused** |
| `binomial("probit")` | **probit** | **refused** |
| `hurdle_gamma()` | **error** on `link_hu` | identity on both blocks |
| `hurdle_gamma(link_hu = "identity")` | **log** on the mean | identity on the mean |
| `hurdle_gamma(link = "identity")` | **error** on `link_hu` | identity on both blocks |
| `"beta"` | **error** | identity |
| `"gamma"`, `fit$family` for a Gamma fit | **error** | identity |
| `do.call(bnec, list(family = Gamma(link = "log")))` | log, unread | log, read |
| `wrapper(family = Beta())` where `wrapper` forwards `...` | **logit** | identity |

In the `Beta` and `Gamma()` rows the curve was fitted to a transform of the mean
while `top`, `bot` and `nec` were reported as though they were on the response
scale. That is the property the identity link exists to preserve, and it is what
the JSS article describes.

`validate_family()` forced identity only for a character string
(`R/validate_family.R:23`). The function branch called `family()` with the
family's **own** defaults, and a constructed object passed through untouched.

## The rule

**Naming a family and nothing more leaves the link to `bayesnec`.** Writing a
link argument makes it yours.

```r
bnec(..., family = "Beta")                 # identity
bnec(..., family = Beta)                   # identity
bnec(..., family = Beta())                 # identity
bnec(..., family = hurdle_gamma())         # identity on both blocks
bnec(..., family = Beta(link = "logit"))   # logit, honoured
bnec(..., family = Gamma(link = "inverse"))# refused
```

Only `identity`, `log` and `logit` are fitted on. The first is what `bayesnec`
assigns; the other two are honoured when asked for because `?models` and
`vignette("example2b")` describe them as supported in about ten places and
`check_models()` has a gate for them. Everything else is refused with an error
naming the family and pointing at the fix.

## Reading intent needs the expression

`Beta()` and `Beta(link = "logit")` produce **identical** family objects, so the
evaluated value cannot distinguish them. `family_link_source()` reads the
unevaluated expression, and each entry point captures it:

| entry point | how |
|---|---|
| `bnec()` | `substitute(list(...))[-1]$family` |
| `get_priors()` | `substitute(family)` |
| `bnec_group()` | `substitute(family)` |
| `bnec_hurdle()` | `substitute(family_growth)` |

`substitute()` rather than `match.call()` at every one of them, including
`bnec()`, because `match.call(expand.dots = FALSE)` records a `...` forwarded
from a wrapper as the placeholder `..1`. `substitute()` follows the promise to
the expression the caller wrote, through any depth of forwarding.

The reading is **per link argument**, not one flag for the family. A single
flag left the block the caller did not write at the brms default, so
`hurdle_gamma(link_hu = "identity")` fitted the mean on `log`.

`update()` on a fitted object is covered as well. `update.bnecfit()` passed the
family **positionally** into `...`, so `names(list(...))` was `NULL`,
`"family" %in% names(named_list)` was `FALSE` in `retrieve_valid_family()`, and
the family was re-derived from the data; `...` then went straight to
`brms::update()`, so even a validated family would not have been the one the
refit received. `update(fit, family = Beta(), force_fit = TRUE)` therefore
refitted on **logit**. The family is now validated at that entry point with the
same `substitute()` capture, and the refit call is assembled so that the
validated family is what `brms` receives. `amend()` is not on this route at
all: it has no `...` and no `family` argument, and reads the family off the fit
it is amending.

A symbol is the one case intent cannot be read from — it is either the
constructor (`Beta`) or a variable holding an object
(`fam <- Beta(link = "logit"); bnec(family = fam)`). `validate_family()`
separates them because it has the value: a constructor names a family, while an
object's own link is honoured and, where that link is not the identity, a
message says which one was taken. Silently replacing a link someone may have
meant is the worse failure of the two.

Anything not written as a constructor call is read the same way, because it
cannot be read at all: `fit$family`, `do.call(bnec, list(family = ...))`, and an
evaluated family inlined into a call.

## Two defects this also fixes

**`validate_family("beta")` errored.** `mod_fams` maps the tag `brms` reports to
the constructor name, and the two differ for one family, so `get("beta")`
resolved to `base::beta()` and gave `unused argument (link = "identity")`. A
user reading the family off a fitted object and passing it back hit this. The
tag is now mapped in `bnec_default_family()`, in one place rather than at each
call site.

**`family = hurdle_gamma()` errored.** `link_hu` defaults to logit, and the
guard requiring identity there fired on a link the user had not chosen. Naming
the family now assigns both links, which is the point of the rule above.

## Existing tests that changed, and why

`test-check_models.R` constructs link-carrying families to exercise the link
gates — 18 call sites now pass `link_source = "chosen"`, which is what they
mean. Three assertions in `test-hurdle_family.R` asserted the old behaviour;
the two that guard a non-identity `link_hu` or `link_zi` now say the caller
chose it, so the guard is still exercised, and the case that previously errored
is asserted as working.

## Verified

`test-validate_family.R` is new: 114 assertions covering the expression reading,
each supply form, the refusal, the message on a variable-held family, every tag
in `mod_fams`, and an end-to-end check through `get_priors()` that the prior
returned records the link actually used. `devtools::document()` regenerated
`man/bnec.Rd` for the new **The link** section. Full suite and CI below.

## Not in this change

`?models` and `vignette("example2b")` describe `log` and `logit` as supported,
which is why they are honoured rather than refused.


beckyfisher:
## Review

Five cycles, each with an independent session re-checking the previous round's
work against the branch rather than accepting it. Fifteen defects were confirmed
by running code and have been fixed on this branch (47718c02, 9773278e,
856d9c06, b98755d7, 195f8837). The findings below are grouped by what they
break.

Every claim here was reproduced against the branch before it was acted on, and
against `dev` where the question was whether something is a regression.

---

### The link was still substituted silently, by five routes

The defect #256 records is a curve fitted to a transform of the mean while
`top`, `bot` and `nec` are reported on the response scale. That fit samples
cleanly and reports healthy R-hat, so every one of these was reachable without
a symptom.

**A positional link was not read.** `link` is the first formal of every family
constructor, so `binomial("probit")` states a link as plainly as
`binomial(link = "probit")` does. Only named arguments were read, so the two
forms got opposite outcomes — one refused, one silently refitted on identity.
The call is now matched against the constructor's own formals before its
argument names are read.

**One flag for the whole family left the other block at the brms default.**
`hurdle_gamma(link_hu = "identity")` fitted the **mean on `log`**. Writing
`link_hu` says nothing about the mean, and writing `link` says nothing about
`hu`, so the reading is now per link argument: each block the caller did not
write is still assigned. `hurdle_gamma(link = "identity")`, which previously
errored on a `link_hu` default the caller had not chosen, now works.

**The constructor was resolved by value lookup, not function mode.** R resolves
the callee of a call in function mode, skipping non-function bindings; `eval()`
does not. An ordinary `Beta <- 0.5` in the caller's workspace therefore made
`family = Beta()` unreadable and beta's own **logit** link was honoured.
Verified end to end: the `top` prior changed from `beta(5, 2)` to a logit-scale
normal, and with `nec3param` the same shadowing turned a working call into
`None of the model(s) specified are valid for a logit link`.

**A family forwarded through `...` was misread.** `match.call(expand.dots =
FALSE)` records a forwarded `...` as the placeholder `..1`, which reads as a
symbol, so `wrapper(family = Beta())` fitted on logit where
`bnec(family = Beta())` fits on identity, and `wrapper(family = Gamma())`
errored on the inverse link. All four entry points now use `substitute()`,
which follows the promise to the expression the caller wrote through any depth
of forwarding.

**Anything not written as a constructor call fell through to "assign".**
`do.call(bnec, list(family = Gamma(link = "log")))` inlines the evaluated
family into the call it builds, and `fit$family` is how a family is read back
off a fitted object. Neither can be read for intent, so both are now treated as
a variable holding a family — honoured, with a message — rather than having the
link replaced in silence.

### A dispersion sub-model was returned to the log scale

`mod_fams` maps `gaussian` and `Gamma` to the **stats** constructors, which
take only `link`. Filtering the rebuild's arguments to what the constructor
accepts silently dropped `link_sigma` and `link_shape`. Visible in the
generated Stan code:

```
supplied  link_sigma = "identity" : vector[N] sigma = rep_vector(0.0, N);
after validate_family()           : vector[N] sigma = rep_vector(0.0, N);
                                    sigma = exp(sigma);
```

This regressed `dev` for `gaussian`, and `disp()` route B models exactly that
parameter. The un-takeable arguments now go through `brms::brmsfamily()`, whose
output is byte-identical at the Stan level to the family the caller supplied.
`Gamma` is the exception: `brmsfamily` reports the tag `gamma`, and rewriting
that tag breaks brms' own dispatch (`.family_Gamma not found`), so a dispersion
link written on `Gamma` is refused with a message pointing at `disp(~x)`. That
input errored on `dev` too, by a different route. A dispersion link that is
only the family's own default is not refused — it is nothing to carry.

### A validation bypass, and a message repeated per group level

`bnec_group()` and `bnec_hurdle()` validate the family and then hand the object
to `bnec()`, where the expression is only the symbol holding it. The link was
read a second time and announced — once per group level — advising the caller
to drop a link they had chosen. `validate_family()` is now idempotent.

The first form of that fix returned early on an already-validated family,
**before** the checks, so a family carrying the marker could be edited into an
unsupported link and passed back. Only the rebuild and the message are skipped
now; every check still runs. The marker is private and is stripped before the
family reaches brms or a returned fit object.

### `fit$family` for a Gamma fit was refused

brms reports the tag `gamma` where `stats::Gamma` reports `Gamma`, and
`mod_fams`, `check_models()` and `mu_support()` all key on `Gamma`. So both
`bnec(family = fit$family$family)` and `bnec(family = fit$family)` were refused
for a Gamma fit — the exact "read the family off a fit and pass it back" case
the `"beta"` mapping in this PR was written for. Worse than the refusal: that
tag also defeats `check_models()`, which returned all 23 equations instead of
the 19 valid on the identity link. A family arriving under a non-canonical tag
is now rebuilt under the canonical one, keeping every link it holds.

### Two remaining classes of input gave the wrong error

`validate_family("t")` gave `unused argument (link = "identity")` — the same
error this PR fixes for `"beta"`, one branch away — because `exists(ctor, mode
= "function")` admits any function in scope. `bnec(family = mean)` gave that
function's own argument error. Both now say the family is not implemented, and
an unimplemented family is reported before an unsupported link.

### Documentation that had become untrue

`@section The link:` was inserted mid-`@details` with nothing reopening the
block, so roughly 290 lines of `?bnec` — the raw-response advice, the
hurdle-versus-zero-inflation discussion, the dispersion section and the NA
handling — rendered under the heading "The link".

Also corrected: `?bayesnecformula`'s claim that `bnec()` *forces* the identity
link, which contradicts `log` and `logit` being honoured; `?bnec`'s statement
that the hurdle families "must be specified" with both links set explicitly,
which the same help page now contradicts; the `disp()` route-B rationale and
its error text; `vignettes/example1.Rmd(.orig)`, which still instructed
supplying the link alongside an explicit family; `vignettes/example2.Rmd(.orig)`,
which attributed the dropped models to a `logit` link in a vignette that
supplies no family and has always been fitted on identity; and four internal
statements that `bnec()` forces identity.

---

### Verified

`devtools::test()` on the branch at 195f8837: 2022 assertions, no failures, no
errors, two pre-existing warnings. `test-validate_family.R` grew from 48
assertions to 114, covering the positional form, per-block reading,
the dispersion links on both constructor families, the `gamma` tag, constructor
shadowing, `do.call` and `fit$family`, `...` forwarding, idempotence and the
marker's privacy, and the expression capture at `bnec()`, `bnec_group()` and
`bnec_hurdle()` — none of which needs sampling, since a refused link errors
before any model is fitted.

Four of the fifteen defects were introduced by an earlier round of this review
and caught by a later one; they are not listed separately above, since only the
final state matters.

### Not in this change

~~`amend()` and `update.bnecfit()` do not apply the link policy at all, and this
PR does not change that.~~ **Superseded.** This has since been done on the
branch, and the attribution to `amend()` was wrong: `amend()` has no `family`
argument. See the follow-up comment below.

`?models` and `vignette("example2b")` describe `log` and `logit` as supported,
which is why they are honoured rather than refused.

beckyfisher:
## What

`update()` on a fitted object now reads the family exactly as `bnec()` does.
This was listed as out of scope in the review comment above; it has since been
done on this branch.

**Correction to that comment and to the body.** Both named `amend()`. `amend()`
has no `...` and no `family` argument (`R/amend.R:40`); it reads the family off
the fit it is amending (`R/amend.R:228`), which is already validated. The route
is `update.bnecfit()` alone, and its own roxygen example passes
`family = Beta(link = "identity")`.

## Why it matters

`update(fit, family = Beta(), force_fit = TRUE)` refitted on **logit** where
`bnec(family = Beta())` fits on identity — the same silent failure this PR
exists to remove, reached by the one route the PR had left open. An unsupported
link such as `Beta(link = "probit")` was accepted without comment on that route
as well.

## Evidence

Captured at the `brms` entry point by tracing `brms:::update.brmsfit`, so it is
what `brms` is actually handed rather than what a fit reports afterwards:

```
update(manec_example, family = Beta(), force_fit = TRUE)

before   brms receives  beta, link = logit
after    brms receives  beta, link = identity
```

Three defects, each confirmed by running code on this branch:

**The family was passed positionally.**
`has_family_changed(object, data_to_check, dot_args$family)` put it into `...`,
so `names(list(...))` was `NULL`, `"family" %in% names(named_list)` was `FALSE`
in `retrieve_valid_family()`, and the family was re-derived from the data. The
guard could not detect a family change at all: it only ever compared the
data-derived family against the fitted one.

**`...` went straight to `brms::update()`.** Even a validated family would not
have been the one the refit received, so naming the argument alone fixes
nothing.

**The comparison never matched.** Only visible once the family was read.
`all.equal()` was applied to the whole family object; `brms` stores a
`brmsfamily` in the fit while `mod_fams` builds gaussian, Gamma, binomial and
poisson from `stats`, so the two carry different components and different
closures. `update(fit, family = "gaussian")` on a gaussian fit therefore asked
for `force_fit = TRUE`. Measured on `manec_example`, which is gaussian:

```
has_family_changed(o, dat, validate_family("gaussian"))   before TRUE   after FALSE
has_family_changed(o, dat, validate_family("Beta"))       before TRUE   after TRUE
has_family_changed(o, nec_data)                           before TRUE   after TRUE
```

`devtools::test()` on the branch at c7be9ecf: 2282 assertions, no failures, the
same two pre-existing warnings. The PR body has been updated: the paragraph
saying this route was not covered now describes what it does, and the
`test-validate_family.R` assertion count was still the pre-review figure.

<details>

**`R/bnecfit-methods.R`, `update.bnecfit()`.** The family is validated at the
entry point, before the change guard, using the same capture as `bnec()`:

```r
if ("family" %in% names(dot_args)) {
  link_source <- family_link_source(substitute(list(...))[-1]$family,
                                    env = parent.frame())
  dot_args$family <- validate_family(dot_args$family,
                                     link_source = link_source)
}
```

`substitute()` rather than `match.call()` for the reason given in the body: a
`...` forwarded from `update.bayesnechurdlefit()` is recorded by `match.call()`
as the placeholder `..1`. The validated family is passed to the guard still
carrying the idempotence marker, so `validate_family()` does not run a second
time with no link source and rebuild a caller-chosen link back to identity. The
marker is dropped before the refit.

The refit call is assembled rather than forwarded, because `...` cannot be
rewritten:

```r
upd_call <- as.call(c(quote(update), quote(fit_i),
                      list(formula. = NULL, newdata = quote(newdata),
                           recompile = quote(recompile)), dot_args))
object[[i]]$fit <- try(eval(upd_call), silent = FALSE)
```

`as.call()` rather than `do.call()` so that `newdata` stays a **symbol**.
`brms::update.brmsfit()` calls `substitute_name(newdata)` and stores the result
as the `data_name` attribute it prints, truncated to 50 characters, so an
inlined data frame would replace the name with 50 characters of the frame
itself. Verified unchanged: the tracer reports `newdata` in both the before and
after runs.

**`R/helpers.R`.** `has_family_changed(x, data, family = NULL)` takes the family
as a named argument. The loop variable it built was also called `family` and
shadowed it; renamed `fam`. New `family_signature()` reduces a family to its tag
and its character-valued `link*` components, sorted, which is what "a different
family" means here.

**Tests.** New `tests/testthat/test-bnecfit-methods.R`, 20 assertions, none
requiring sampling: `family_signature()` across the stats and `brms`
constructors; `has_family_changed()` with a supplied family and with none; the
refusal of an unsupported link before any refit; the change guard; and what
`brms` receives, captured by tracing `update.brmsfit` and halting the refit at
that point. The last covers the assigned link, the honoured link written both
named and positionally, the absence of the marker, and the `newdata` symbol.

**Not changed.** `update.bayesnechurdlefit()` forwards `...` to
`update.bnecfit()` for each component, so it inherits the policy without its own
change; the `substitute()` capture follows the promise through that forwarding.

</details>

beckyfisher:
This looks good.

Issue #196, opened 2026-08-14
ecx() and nsec() drop inner arithmetic when back-transforming an inline crf() transformation


beckyfisher:
## Summary

When a transformation is written inline in `crf()`, `ecx()` and `nsec()`
back-transform their result by substituting into the **first argument slot** of
the parsed call. Any arithmetic inside that slot is silently discarded, so
`crf(log(x + 1))` is back-transformed as `log(x)`.

The model fit is unaffected. Only the reported `ecx` and `nsec` values are
wrong, and they are wrong silently: nothing errors, and the value returned is
plausible.

## Root cause

`R/ecx.R:198-201`, and the identical pattern at `R/nsec.R:173-175`:

```r
x_call <- str2lang(eval(parse(text = x_str)))
if (inherits(x_call, "call")) {
  x_call[[2]] <- str2lang("ecx_out")
```

`x_call[[2]]` is the call's first argument. For `log(raw_x)` that argument is
the symbol `raw_x`, and replacing it is correct. For `log(raw_x + 1)` the
argument is the expression `raw_x + 1`, and replacing the whole expression drops
the `+ 1`.

No fitting required to see it:

```r
for (s in c("log(raw_x)", "log(raw_x + 1)", "log(raw_x/2)", "sqrt(raw_x)")) {
  x_call <- str2lang(s)
  x_call[[2]] <- str2lang("ecx_out")
  cat(sprintf("%-16s -> %s\n", s, deparse(x_call)))
}
#> log(raw_x)       -> log(ecx_out)     correct
#> log(raw_x + 1)   -> log(ecx_out)     "+ 1" dropped
#> log(raw_x/2)     -> log(ecx_out)     "/2" dropped
#> sqrt(raw_x)      -> sqrt(ecx_out)    correct
```

Affected: any inline transformation whose argument is not a bare variable.
Unaffected: `log(x)`, `sqrt(x)`, and any pre-transformed column.

## Demonstration

`prop_data` from `vignette("example1")`, `nec4param`, `seed = 333`. The same
model is fitted two ways: the transformation inline, and the transformation
pre-computed into a column.

```r
prop_data$log_x <- log(prop_data$raw_x + 1)
a <- bnec(resp ~ crf(log(raw_x + 1), model = "nec4param"), data = prop_data, seed = 333)
b <- bnec(resp ~ crf(log_x, model = "nec4param"), data = prop_data, seed = 333)
```

[...]

Issue #160, opened 2026-01-16
NEC on autoplot and base plots not plotting correctly when a function is called for x


beckyfisher:
``` r
library(tidyverse)
library(bayesnec)
#> Loading required package: brms
#> Loading required package: Rcpp
#> Loading 'brms' package (version 2.22.0). Useful instructions
#> can be found by typing help('brms'). A more detailed introduction
#> to the package is available through vignette('brms_overview').
#> 
#> Attaching package: 'brms'
#> The following object is masked from 'package:stats':
#> 
#>     ar
#> 
#> Attaching package: 'bayesnec'
#> The following object is masked from 'package:stats':
#> 
#>     step
library(patchwork)

data(nec_data)
nec_data <- nec_data |> 
  mutate(log.x = log(x))

exmp_a <- bnec(y ~ crf(log(x), model = "nec4param"), data = nec_data, chains = 2)
#> Finding initial values which allow the response to be fitted using a nec4param model and a beta distribution.
#> Compiling Stan program...
#> Start sampling
#> 
#> SAMPLING FOR MODEL 'anon_model' NOW (CHAIN 1).
#> Chain 1: 
#> Chain 1: Gradient evaluation took 0.000237 seconds
#> Chain 1: 1000 transitions using 10 leapfrog steps per transition would take 2.37 seconds.
#> Chain 1: Adjust your expectations accordingly!
#> Chain 1: 
#> Chain 1: 
#> Chain 1: Iteration:    1 / 10000 [  0%]  (Warmup)
#> Chain 1: Iteration: 1000 / 10000 [ 10%]  (Warmup)
#> Chain 1: Iteration: 2000 / 10000 [ 20%]  (Warmup)
#> Chain 1: Iteration: 3000 / 10000 [ 30%]  (Warmup)
#> Chain 1: Iteration: 4000 / 10000 [ 40%]  (Warmup)
#> Chain 1: Iteration: 5000 / 10000 [ 50%]  (Warmup)
#> Chain 1: Iteration: 6000 / 10000 [ 60%]  (Warmup)
#> Chain 1: Iteration: 7000 / 10000 [ 70%]  (Warmup)
#> Chain 1: Iteration: 8000 / 10000 [ 80%]  (Warmup)
#> Chain 1: Iteration: 8001 / 10000 [ 80%]  (Sampling)
#> Chain 1: Iteration: 9000 / 10000 [ 90%]  (Sampling)
[...]

Issue #93, opened 2021-11-12
Auto-correct when check_data applies shifts from 0 and 1


beckyfisher:
The check_data function will apply a shift to ether the x or y data under certain scenarios. Ideally predicted values (and associated NEC, NSEC and ECx estimates) should be corrected for this shift before being returned.

beckyfisher:
Note this might be tricky for NEC as it is a returned parameter, not a predicted value.

beckyfisher:
An alternative to providing corrected values would be to simply add the applied correction as an attribute to the bayesnecfit or bayesmanecfit objet.

beckyfisher:
#270 removes the predictor correction entirely, so the half of this issue that
the second comment above identifies as the important one is now moot. There is
no predictor shift to reverse: a zero concentration reaches `brm()` as recorded,
and `nec()`, `ecx()` and `nsec()` report on the scale the user supplied.

What remains is the response correction. `check_data()` makes three, and they
differ in one respect that matters here.

- A zero under `Gamma` is shifted to `min(y[y > 0]) / 10`, **with a message**
  naming the substituted value and recommending `family = hurdle_gamma()` where
  the zeros are meaningful.
- A zero under `beta` is shifted the same way, **silently**.
- A one under `beta` or `zero_inflated_beta` is reduced by 0.001, **silently**.

Measured on `dev`, a `Beta(link = "identity")` response of
`c(rep(1, 5), rep(0.6, 5), rep(0.3, 5), 0, 0, 0.05, 0.05, 0.05)`, capturing every
condition `check_data()` raises:

```
messages emitted: none
y in    1.000 1.000 ... 0.000 0.000 0.05
y out   0.999 0.999 ... 0.005 0.005 0.05
```

So the argument #270 makes against the predictor correction — that it was
applied without notice — applies unchanged to two of these three.

**None of the three is an offset that can be reversed.** A shifted predictor was
an offset on an axis, so subtracting it from a `nec` or an ECx would have
recovered the value on the recorded scale. A substituted response is a modelling
decision: the fitted curve estimates the mean of the data as substituted, and
there is no quantity to subtract from a prediction to undo it. The correct
treatments are the ones the `Gamma` message already points at —
`hurdle_gamma()` or `zero_inflated_beta()` where the zeros are real, or `cens()`
where they are a detection limit, which #181 added and `vignette("example1")`
documents.

The proposal in the issue body, that estimates be corrected for the shift, has
[...]

PR #264, opened 2026-09-02
decide the check_data write-back per variable, not per formula (#258)

### LINKED ISSUE #258: check_data() boundary corrections are discarded when a transformation is written inline in crf(), so the fit fails on values the package reports as repaired
## Summary

`check_data()` corrects a response or predictor that sits on a boundary its
family cannot represent — a zero under `Gamma` or `beta`, a one under `beta`.
`fit_bayesnec()` then writes those corrected values back into the data frame
passed to `brm()` **only when no population variable is transformed inside
`crf()`**. Where an inline transformation is present, the correction is computed,
reported to the user, and then discarded, and `brm()` receives the original
boundary values.

The fit fails outright with a `brms` error naming a condition the package has
already stated it repaired. The transformation and the correction need not
concern the same variable: a `log()` on the **predictor** discards a shift
applied to the **response**.

This is not a regression. `master` (2.1.3.1, the current CRAN release) fails
identically.

## Reproduction

Paired runs on packaged data, differing only in where the logarithm is applied.
R 4.6.1, `brms` 2.23.0, cmdstanr 0.9.0 / cmdstan 2.39.0, `dev` at 2.1.3.22,
2 chains, 2000 iterations, `seed = 17`.

`herbicide` records one exact zero in `fvfm` for hexazinone, which `check_data()`
shifts to 0.001.

```r
data(herbicide)
hex <- herbicide[herbicide$herbicide == "hexazinone", ]

# (a) transformation applied before bnec() -- the form used in the JSS paper
hex_a <- hex
hex_a$concentration <- log(hex_a$concentration)
bnec(fvfm ~ crf(concentration, model = "nec3param"), data = hex_a, seed = 17)
#> bayesnecfit; nec = 0.8715 [0.7809, 0.9447]; max rhat 1.014
#> min(fit$fit$data$fvfm) == 0.001    the shift reached brms

# (b) transformation written inline -- identical model, identical data
bnec(fvfm ~ crf(log(concentration), model = "nec3param"), data = hex, seed = 17)
#> Error : Family 'beta' requires response greater than 0.
```

In both cases `check_data()` returns `mod_dat` with `min(y) == 0.001` and no
remaining zeros. Only in (b) is that result discarded.

A minimal case, on a `Gamma` response, shows the package reporting a repair it
does not perform:

```r
d <- data.frame(x = rep(c(0.1, 1, 10, 100), each = 5))
d$y <- c(rep(c(8, 6, 3), each = 5), rep(0, 5))
bnec(y ~ crf(log(x), model = "nec3param"), data = d, seed = 17)
#> Your response contains zeros, which a Gamma distribution cannot represent.
#> They have been shifted to 0.3 (one tenth of the smallest non-zero value). ...
#> Error : Family 'gamma' requires response greater than 0.
```

The message is emitted from `check_data()`, which always runs; whether the shift
is retained is decided afterwards, in `fit_bayesnec()`.

## Root cause

`R/fit_bayesnec.R:40-52`:

```r
trans_vars <- find_transformations(bdat)
# if no transformations are applied via formula (including on trials),
# use the output of check_data
if (length(trans_vars) == 0) {
  data[, y_var] <- y
  data[, x_var] <- x
  if (family$family == "binomial" || family$family == "beta_binomial") {
    data[, t_var] <- tr
  }
}
```

`find_transformations()` returns every population variable that does not appear
by name in the model frame, across `x_var`, `y_var` and `trials_var` together.
The guard is therefore all-or-nothing: a transformation on any one of the three
suppresses the write-back for all three.

The family selected by `check_data()` is assigned at `R/fit_bayesnec.R:37`,
before the guard, and is retained. Only the data corrections are lost.

All six corrections in `check_data()` are affected: the zero shifts on `x` for
`Gamma` and `beta` (lines 156, 178), the zero shifts on `y` for `Gamma` and
`beta` (169, 182), and the shifts away from one on `x` and `y` for `beta` and
`zero_inflated_beta` (185, 192).

## Scope

Affected: any fit combining an inline `crf()` transformation with a response or
predictor value on a boundary the family excludes. Within the packaged data this
covers the ametryn, irgarol and hexazinone subsets of `herbicide`, which record
12, 10 and 1 zero `fvfm` values respectively.

Unaffected: transformations pre-computed into a column, and data that touch no
boundary. The JSS paper applies `log()` with `dplyr::mutate()` before calling
`bnec()` (`article/article.Rmd:585`), so none of its fits encounter this,
including the ametryn analysis of Section 6.

## Relation to existing issues

The defect is distinct from, but shares a cause with, three open issues:

- **#93** — the shift is applied but reported estimates are not corrected back
  for it. The converse case: there the shift reaches the fit, here it does not.
- **#196** — `ecx()`/`nsec()` mishandle an inline transformation when
  back-transforming.
- **#160** — plots mishandle an inline transformation on the predictor.

In all four, an inline `crf()` transformation is tracked by name-matching or
positional substitution and each consumer handles it independently. A shared
representation of the transformation, recorded once on the model frame, would
address the class rather than the instance, and is worth considering before a
fourth separate patch.

This one differs from its neighbours in causing a hard fit failure rather than a
silent numerical error.

## Consequences elsewhere

- `vignette("example1")` (line 106) recommends writing transformations inline in
  `crf()`, and states that `bayesnec` substitutes `min(x[x > 0]) / 10` where it
  must. A reader following that guidance with a zero-bounded response receives an
  error rather than the substitution described.
- The `example9` draft on #238 fits `fvfm ~ crf(log(concentration), ...)`. Both
  the reported failure of the whole `decline` set on irgarol and the reported
  failure on hexazinone are instances of this defect, not of the `Beta` zero
  boundary that the draft attributes them to. That attribution should be
  corrected when #238 is reworked.

## Suggested fix

Decide the write-back per variable rather than for the formula as a whole: a
transformation on `x` should not suppress a correction to `y`. Where the
transformed variable is itself the one corrected, the correction must be applied
before the transformation is evaluated, since applying it afterwards would shift
a value on the transformed scale rather than the recorded one.

Failing that, the minimum acceptable behaviour is to raise an error naming the
conflict at `check_data()` time, rather than allowing `brm()` to fail on a
condition the user has been told was repaired.

## Tests

`tests/testthat/` has no case combining an inline transformation with a boundary
value, which is why this passes the check matrix. Two are needed: a `beta`
response containing a zero and a `Gamma` response containing a zero, each fitted
with `crf(log(x), ...)`, asserting that the values reaching `brm()` carry the
same corrections as the equivalent pre-transformed call.

beckyfisher:
Closes #258.

## What

A boundary correction computed by `check_data()` now reaches `brm()` whether or
not a population variable is transformed inside `crf()`. The decision is made
one variable at a time instead of for the formula as a whole.

## Why it matters

`check_data()` shifts a response off a boundary its family cannot represent — a
zero under `Gamma` or `beta`, a one under `beta`. The shift was written back
into the data frame `brm()` is given only when **no** population variable was
transformed inside the formula, so a `log()` on the **predictor** discarded a
shift applied to the **response**. The correction was computed, reported to the
user, and then dropped, and the fit failed with a `brms` error naming the
condition the package had just said it had repaired.

This affects any fit combining an inline `crf()` transformation with a response
on a boundary the family excludes, which includes the ametryn, irgarol and
hexazinone subsets of the packaged `herbicide` data. It is not a regression:
`master` (2.1.3.1) fails identically.

## Evidence

On `dev` at 5f6ffe72, R 4.6.1:

```r
hex <- herbicide[herbicide$herbicide == "hexazinone", ]
bnec(fvfm ~ crf(log(concentration), model = "nec3param"), data = hex, seed = 17)
#> Error : Family 'beta' requires response greater than 0.
```

`check_data()` returns `mod_dat` with `min(y) == 0.001` in that call; only the
write-back discards it. With this branch the same call hands `brm()` a response
whose minimum is `0.001`, identical to the equivalent pre-transformed call.
`tests/testthat/test-fit_bayesnec.R` asserts that equality for a `beta` response
containing a zero and a `Gamma` response containing a zero, in both cases fitted
with `crf(log(x), ...)`, which is the pair of cases the issue asks for. For the
`Gamma` case the assertion is made of the data frame `brm()` is handed by
`fit_bayesnec()` itself.

Two further defects were found while making the change and are fixed here, and
three further changes follow from the decision:

- Any `NA` in a population variable made the write-back fail with `replacement
  has 19 rows, data has 20`. The model frame drops incomplete cases, so it is
  shorter than the data frame `brm()` is handed. Rows are now matched by name.
- `trials()` carrying arithmetic fitted every observation against the wrong
  number of trials. `y | trials(n * 2) ~ crf(x, ...)` had the doubled values
  written into the user's `n` column before `brm()` saw it, because the aterm
  matches the bare column name; `brm()` then evaluated `trials(n * 2)` against
  those, so a recorded 10 became 40. The write-back is removed — `check_data()`
  never corrects trials, so it had nothing to carry.
- A fit that previously ran without error can now be given a different
  predictor. Where the response is transformed inline, the predictor's zero
  shift used to be discarded along with everything else, so it reached the
  priors and not the data; it now reaches both. The two agree for the first
  time — on `dev` the `nec` prior was bounded below by the shifted value while
  the data still carried the zero. Recorded in NEWS, since a fit of that shape
  gives different estimates when refitted.
- `get_priors()` and `update()` run the same check, so a formula whose response
  is transformed inline onto a boundary now raises the new error there too,
  where `get_priors()` previously returned a prior table for a fit that could
  not be run. Recorded in NEWS.
- `vignette("example1")` stated that `bayesnec` substitutes `min(x[x > 0]) / 10`
  for a zero concentration written as `crf(log(x))`. It does not: the
  transformation is evaluated first, so `log(0)` arrives at the data check as
  `-Inf` and the fit stops. The paragraph now says which form gets the
  substitution.

<details>

**`R/helpers.R` — new `pop_var_is_transformed(data, var)`.** `find_transformations()`
answers the question for the formula as a whole, which is what made the guard
all-or-nothing. The new function answers it for one variable, comparing the
model frame column at the variable's position in `bnec_pop` against the
variable's own name. The comparison is positional because that is how
`retrieve_var()` and every other consumer of `bnec_pop` finds a variable, so the
two cannot disagree about which column is meant; a name-membership test would
also call `y` transformed in `y ~ crf(log(y))`. `clean_aterms()` is applied
first, so `trials(n)` is not counted as a transformation.

**`R/fit_bayesnec.R` — new `write_back_checks(data, bdat, var, values)`**,
replacing the `length(trans_vars) == 0` guard. It is called for the response and
for the predictor. There is no trials call: `check_data()` never corrects the
trials variable, so where the aterm names a bare column the write-back was a
no-op, and where it carries arithmetic it was wrong — `clean_aterms()` maps
`trials(n * 2)` back to `n`, so the doubled values were written into the user's
`n` and `brm()` evaluated the aterm against those. It returns `data` unchanged where the variable is transformed, and
otherwise assigns the checked values into rows matched by
`match(rownames(bdat), rownames(data))` rather than wholesale, which is what
fixes the `NA` case above.

**`R/check_data.R` — the six correction sites.** A transformed variable is no
longer corrected at all, so nothing is left for the write-back to lose:

- *Response, transformed, on a boundary the family excludes.* New
  `check_inline_boundary()` raises an error naming the response as written, the
  family, the boundary and the remedy. It is called from `bnec()` before the
  model loop, beside `check_normalisation()`, which is hoisted out of
  `check_data()` for the same reason: `check_data()` runs once per model, so a
  model set would otherwise print the conflict for every member and end on the
  generic all-models-failed advice. `check_data()` keeps the call as a backstop
  for `get_priors()` and for a direct `fit_bayesnec()`. One consequence of the
  `bnec()` position: the conflict is now raised before `check_cens_support()`,
  so a response that is both transformed onto a boundary and censored there
  reports the transformation conflict from `bnec()` and the censoring conflict
  from `get_priors()`. Both are true and both must be resolved. The correction cannot be carried through:
  `brm()` re-evaluates the transformation from the recorded column, so the shift
  would be discarded and the fit would fail regardless. This is the minimum
  behaviour the issue asks for, in place of a `brms` error about a repair the
  package reported making. The `Gamma` zero case keeps the existing
  `hurdle_gamma()` suggestion.
- *Predictor, transformed, at zero.* Left where it is. A zero on the transformed
  scale is not evidence of a boundary artefact on the recorded scale — `log(1)`
  is zero for a concentration of one — so shifting it would replace a legitimate
  value, and no family constrains the support of a predictor. The corrected
  value is no longer used for the priors either, since `check_data()` no longer
  produces one, which removes an inconsistency between the priors and the data
  the fit sees. The visible consequence is the `nec` prior's lower bound: on a
  `crf(sqrt(x))` with `x` taking 0, 1, 10 and 100 it is `0` where it was `0.1`.
  Recorded in NEWS.

**Scope.** The shared representation of an inline transformation proposed in the
issue, which would also address #93, #196 and #160, is not attempted here.

**Not fixed, raised separately.** `check_data()` tests `x_type == "beta"` at two
sites while `set_distribution()` returns `"Beta"`, so the predictor corrections
for a 0-1 bounded predictor are unreachable. Verified: a predictor of
`c(0, 0.25, 0.5, 1)` comes back from `check_data()` unchanged. Raised as #265;
both lines carry the guard this change adds and a comment saying they are
unreachable, so that whichever way #265 resolves it is one change and not two.

Also raised: #266, on `make_good_inits()` taking 561 seconds for one model on a
twenty-row dataset before falling back to Stan's own initialisation. It is
unrelated to this change, but it decides how the tests here are written.

**Tests.** `tests/testthat/test-fit_bayesnec.R`, fourteen cases at three
levels. Four drive `fit_bayesnec()` itself with `brm()` mocked, which is what
proves the correction reaches the fit; they supply `init` so that
`add_brm_defaults()` skips the initial-value search of #266. One drives `bnec()`
over the `decline` set to assert that the conflict is raised once rather than
per model. Six run `check_data()` and `write_back_checks()` in the order
`fit_bayesnec()` calls them, which is cheap enough to cover the boundary cases
exhaustively; the remaining three assert on `check_data()` or on
`pop_var_is_transformed()` alone.

</details>

https://claude.ai/code/session_013Wfabsoz4FyAQmFz6ShMkK


beckyfisher:
## Review record

Six rounds of independent review were run against this branch, each by a session
with no part in writing it. The first found that the fix had no test; the last
found nothing. What each round found, and what it changed:

**Round 1 — the fix was untested.** The test file asserted against a
reimplementation of the three `write_back_checks()` calls rather than against the
calls themselves, so deleting the `y_var` call left the suite passing. The round
also established that `fit_bayesnec()` can be driven in under a second by mocking
`brm()` and supplying `init`, which is what made an integration test affordable.
Removing the `y_var` call now fails a named test.

**Round 2 — the new error fired once per model.** `check_data()` runs per model,
so a model set printed the conflict twelve times and then ended on the generic
all-models-failed advice, with the cause far above it. The check was factored
into `check_inline_boundary()` and is raised once from `bnec()`, beside
`check_normalisation()`, which the repository hoists for exactly this reason.
The round also established that the trials write-back was a provable no-op, and
it was dropped.

**Round 3 — dropping the trials write-back was not inert.** `clean_aterms()`
maps `trials(n * 2)` back to `n`, so the write-back overwrote the user's `n` with
the doubled values and `brm()` then evaluated the aterm against those: a recorded
10 fitted as 40. That is a silent wrong answer on `dev` that this branch fixes,
and it now has its own NEWS bullet and a regression test. The round also found an
assertion that could not fail, since the text it looked for is raised by `stop()`
rather than `message()`.

**Round 4 — a fit that previously ran without error is given a different
predictor.** Where the response is transformed inline, the predictor's zero shift
used to be discarded along with everything else, so it reached the priors and not
the data. It now reaches both, and the two agree for the first time. The reach is
any family, and the shift is not always small, so it has its own paragraph in
NEWS rather than following from the general statement.

**Round 5 — measurements in NEWS that were true only of one dataset.** The `nec`
lower bound was given as `0.1` where it is one tenth of the smallest positive
transformed value; the reach of the predictor shift was overstated, since it
fires only where the predictor spans values above one; and the enumeration of
what `check_data()` corrects omitted `zero_inflated_beta` at one. The round also
found that #267, raised in round 4, was a duplicate of #260, already merged to
`dev` six hours earlier. It is closed, and `dev` is merged in.

**Round 6 — no findings.** Every round-5 item was re-measured on the merged tree
rather than read. The merge left no semantic conflict: #260's rewritten
`has_family_changed()` is intact and both of its routes reach the new check.

## State

The full `R CMD check` matrix — macOS, Ubuntu release and devel, Windows — and
pkgdown pass on the merged head. `test-fit_bayesnec.R` is fourteen cases and
thirty-four assertions.

Four issues were raised from this work and are not addressed here:

- **#265** — `check_data()` compares `x_type` against `"beta"` while
  `set_distribution()` returns `"Beta"`, so the two predictor corrections for a
  0-1 bounded predictor are unreachable. Both lines carry a comment saying so.
- **#266** — `make_good_inits()` spends up to ten thousand attempts before
  falling back to Stan's own initialisation: 561 seconds for one model on a
  twenty-row dataset, per model.
- **#268** — `plot()` and `autoplot()` use the same all-or-nothing guard this
  change replaces, so a transformation on the response suppresses `xform` on the
  predictor axis.
- **#267** — closed as a duplicate of #260.

https://claude.ai/code/session_01NwkNQ4QftY6brsHqU4bCto

beckyfisher:
This all looks good.

PR #270, opened 2026-09-03
remove the predictor zero correction, and take the nec prior scale from the concentration series (#269)

### LINKED ISSUE #265: check_data() compares x_type against "beta" while set_distribution() returns "Beta", so the predictor boundary corrections for a 0-1 bounded predictor never fire
## What

`check_data()` decides whether to correct a predictor sitting on a boundary by
testing `x_type == "beta"`, but `set_distribution()` — the function that produced
`x_type` — returns `"Beta"` with a capital B. The two predictor corrections for a
0-1 bounded predictor are therefore unreachable, and have been since they were
written.

## Why it matters

A predictor recorded as a proportion — a dilution fraction, a proportion of a
reference concentration — reaches the fit with its zeros and ones intact, while
the equivalent predictor on a wider scale has its zeros shifted to
`min(x[x > 0]) / 10`. Neither behaviour is announced, so two datasets that differ
only in the units of the predictor are treated differently with no indication
that they were.

Whether the correction should fire at all is the open question. No family
constrains the support of a predictor, so unlike the response corrections these
are not needed for the fit to run, and the `Gamma` branch that does fire silently
modifies the user's recorded predictor. The choice is between making the two
branches reachable and removing all three.

## Evidence

`dev` at 5f6ffe72, R 4.6.1:

```r
x0 <- c(0, 0.25, 0.5, 1)
bayesnec:::set_distribution(x0, silence_y_msgs = TRUE, silence_x_msgs = FALSE)
#> [1] "Beta"

d <- data.frame(x = rep(x0, each = 5), y = rep(c(8, 6, 3, 1), each = 5))
bdat <- model.frame(bnf(y ~ crf(x, model = "nec3param")), data = d)
cd <- bayesnec:::check_data(bdat, Gamma(link = "identity"), "nec3param")
range(cd$mod_dat$x)
#> [1] 0 1        # a shift would give 0.025 and 0.999
```

`set_distribution()` returns `"Gamma"`, `"Beta"`, `"gaussian"`, `"poisson"` or
`"binomial"`. `check_data()` compares `x_type` against `"Gamma"` once, which
matches, and against `"beta"` twice, which cannot.

Found while fixing #258, which touches the same three lines but does not change
which of them are reachable.

<details>

`R/check_data.R`, the three predictor sites:

```r
if (min(x) == 0 & x_type == "Gamma" & !x_transformed) {   # fires
if (min(x) == 0 & x_type == "beta" & !x_transformed) {    # cannot fire
if (max(x) == 1 & x_type == "beta" & !x_transformed) {    # cannot fire
```

The response corrections below them compare `fam_tag`, which comes from
`family$family` and is genuinely lower case for the `beta` family, so only the
predictor comparisons are affected.

Deciding this needs an answer to what the predictor shift is for. It changes the
values the fit sees and the values the priors are built from, and it is silent,
unlike the `Gamma` response shift which messages. If it exists to keep a
predictor away from zero for equations that divide by it or take its logarithm,
the condition should be the equation rather than the distribution of the
predictor. If it exists for the priors alone, it should not be written into the
data at all.

Whichever way it is resolved, a test asserting the behaviour of a 0-1 bounded
predictor is needed; `tests/testthat/` has none.

</details>

### LINKED ISSUE #269: Remove the predictor zero correction: it is not required by any family, prior or equation, and it silently changes the recorded predictor
## What

`check_data()` replaces a zero predictor with `min(x[x > 0]) / 10` before the
data reach `brm()`. The correction is not needed by any family, prior or model
equation, it is applied silently, and it is not undone in the estimates it
changes. It should be removed.

The one calculation that depends on the predictor being positive is the rate of
the `gamma` prior on `nec` and `ec50`, which is built from
`quantile(predictor, 0.5)`. That should be taken from the tested concentration
series rather than from the observation vector, so that it does not depend on
how many replicates each concentration received.

## Why it matters

A predictor is not a modelled variable. No likelihood is defined for it, so no
family constrains its support, and the correction was written by symmetry with
the response correction beside it rather than for a reason of its own.

Three consequences follow from keeping it.

- **It is silent.** The response correction emits a message; the predictor
  correction emits nothing. The predictor the fit uses differs from the
  predictor the user recorded, with no notification.
- **It is not corrected back.** #93 has been open since 2021 for exactly this,
  and its second comment records that the predictor is the half that matters:
  "this is only important for adjustments to x, as these impact the endpoint
  estimates when they are near zero."
- **It depends on the design rather than on the data.** `min(x[x > 0]) / 10`
  means the same control is fitted at a different value depending on the lowest
  non-zero concentration tested, so two laboratories testing the same substance
  fit different controls.

It is also applied inconsistently. #265 records that `check_data()` tests
`x_type == "beta"` while `set_distribution()` returns `"Beta"`, so a 0-1
bounded predictor containing zeros has never been corrected, in any release,
and nothing has been reported to fail there.

## Evidence

R 4.6.1, `brms` 2.23.0, cmdstanr 0.9.0 / cmdstan 2.39.0, `dev` at 130aede3.

**The prior does not require a positive predictor.** `define_prior()` for
`nec3param` under `Gamma(link = "identity")`, on `x = c(0, 5, 15, 45, 135)`:

```
with the correction     <lower=0.5,upper=135> gamma(5, 0.1333)
without the correction  <lower=0,  upper=135> gamma(5, 0.1333)
```

The prior is the same; only the truncation bound differs. `gamma(5, r)` has
zero density at zero, so `lower = 0` is a proper truncation.

**Stan does not require it.** The structural risk is the three equations that
raise the predictor to an estimated power -- `ecxsigm` (`x^exp(d)`),
`nechormepwr` and `nechorme4pwr` (`x^(1 / (1 + exp(slope)))`) -- where the
derivative with respect to the exponent is `pow(x, y) * log(x)`, undefined at
`x = 0`. A minimal Stan model containing `pow(x[n], exp(d))` with a zero in `x`
was checked with the cmdstan `diagnose` method: autodiff and finite differences
agree to 2e-8. This was predictable from the package's own code, since
`necsigm` computes `(step(x - nec) * (x - nec))^exp(d)`, whose base is exactly
zero for every observation below `nec`; `bayesnec` therefore already raises
zero to an estimated power on every `necsigm` fit.

**A fit does not require it.** `nec3param` and `ecxsigm`, `Gamma(identity)`,
`x = 0, 5, 15, 45, 135` with six replicates, 2 chains x 2000 iterations,
`seed = 17`, fitted with the correction and with it disabled. All four fits
converged with zero divergent transitions.

| | with correction | without correction |
|---|---|---|
| `nec3param`, `nec` | 11.73 [5.54, 19.20] | 11.73 [5.93, 21.30] |
| `nec3param`, EC10 | 16.93 [11.13, 24.20] | 16.89 [11.08, 26.09] |
| `ecxsigm`, EC10 | 6.42 [3.73, 11.41] | 6.08 [3.11, 11.08] |

**What the correction masks.** `define_prior()` builds the `gamma` rate as
`1 / (quantile(predictor, 0.5) / 2)`. Where more than half the observations sit
at zero -- a design with many control replicates -- the median is zero and the
prior string becomes `gamma(5, Inf)`. On
`x = c(rep(0, 12), 5, 15, 45, 135, 200, 300)`:

```
raw                     <lower=0,  upper=300> gamma(5, Inf)
with the correction     <lower=0.5,upper=300> gamma(5, 4)
```

The second has a prior mean for `nec` of 1.25 on a predictor running to 300. So
the correction converts a hard failure into a silently degenerate prior. The
defect is that the prior scale is derived from a replicate-weighted quantile of
the predictor rather than from the concentration series.

<details>

**`R/check_data.R` -- remove the three predictor corrections.** The
`min(x) == 0 & x_type == "Gamma"` branch, and the `min(x) == 0 & x_type ==
"beta"` and `max(x) == 1 & x_type == "beta"` branches that #265 records as
unreachable. This resolves #265 by deletion rather than by repair: the branches
are removed rather than being made to fire, so #265 should be closed as
superseded rather than fixed.

`x_transformed` and the `x_type` assignment become unused in `check_data()`
once the branches are gone. The `set_distribution(x, ...)` call should be
retained, because it is also the integer-predictor guard: `x` is retrieved with
`retrieve_var()`, an integer column passes `is.numeric()`, and the "bayesnec
does not currently support integer concentration data" error is raised from
inside `set_distribution()` when `silence_x_msgs = FALSE`. `define_prior()`
makes the same call, so deleting this one would relocate the error into prior
construction rather than remove it; it belongs in the documented data check.

**`R/define_prior.R` -- take the `gamma` rate from the concentration series.**
Replace `quantile(predictor, probs = 0.5)` in the `Gamma` entry of `x_prs` with
the median of the distinct predictor values. The rule is that the prior scale
should describe the concentrations that were tested, not how many replicates
each received. The precedent is inside the package: `survival_by_x()` already
primes the `hu` block of a hurdle or zero-inflated fit from
`sort(unique(predictor))`.

Measured effect:

| predictor | current rate | median of distinct levels |
|---|---|---|
| `0, 5, 15, 45, 135`, six replicates each | 0.1333 | 0.1333, unchanged |
| twelve controls plus 5, 15, 45, 135, 200, 300 | 4, prior mean `nec` 1.25 | 0.0444, prior mean `nec` 112.5 |

The second row is the rate as released, taken over the *corrected* observations,
whose median is 0.5. On the recorded predictor the same expression returns
`Inf`, so removing the correction without changing the rate would replace a
degenerate prior with a failure.

A balanced design is unaffected. An unbalanced one gets a different `nec` and
`ec50` prior, which is a change to be recorded in NEWS.

The `gaussian` entry of `x_prs` is left alone. It applies where the predictor
takes negative values, a median of zero is a legitimate centre for a normal so
there is no failure to fix, and its scale term `sd(predictor) * 10` is
deliberately diffuse. Changing it would alter the prior on every log-scale fit
for no stated defect.

**Tests.** `tests/testthat/test-fit_bayesnec.R:104` and `:150` assert
`min(x) == 0.1` after the correction and are inverted; the
`fit_bayesnec() hands brm() the repaired predictor` block is replaced by one
asserting the predictor reaches `brm()` as recorded, zeros included. The
`a zero on a transformed predictor is left where it is` test collapses, since
transformed and untransformed predictors are now treated the same. A test is
added for the many-controls prior, asserting the rate is finite.

**Documentation.** `vignettes/example1.Rmd:93` states that `bayesnec`
substitutes `min(x[x > 0]) / 10` for a zero on the recorded axis. That
paragraph was rewritten under #258 and needs rewriting again. `.Rmd.orig`
states the same text.

**Definition of done.** A predictor containing zeros reaches `brm()` unchanged;
`nec()`, `ecx()` and `nsec()` are on the recorded predictor scale with no
undisclosed offset; the prior rate is finite for any predictor with at least
one positive value; `R CMD check` and the full test suite pass.

**Stop and ask rather than guess** if removing the correction turns out to
break a family or equation not covered above, or if a decision is needed on
whether #93 can be closed by this change.

</details>

beckyfisher:
Closes #269. Closes #265.

## What

A zero in the predictor is no longer corrected. `check_data()` replaced it with
`min(x[x > 0]) / 10` before the data reached `brm()`; that correction is
removed, and the predictor now reaches the fit exactly as recorded.

The rate of the `gamma` prior on `nec` and `ec50` is now built from the
distinct predictor values rather than from the observation vector. That is the
one calculation that had depended on the predictor being positive, but the
change is not confined to predictors containing a zero: **any design with
unequal replication across concentrations gets a different `nec` and `ec50`
prior.**

## Why it matters

No likelihood is defined for a predictor, so no family constrains its support
and no correction was ever required. The correction was written by symmetry with the
response correction beside it: both appeared in the first version of
`check_data()` (`89a15d03`, 2020-05-25) under one comment, `# error catching
for 0 for gamma by adding very small value`, and the two branches are
character-for-character parallel. The response one is necessary; `Gamma` and
`Beta` exclude zero from their support. The predictor one treated `x_type` as
if it were a likelihood family for the predictor, which it is not — `x_type`
selects only the family of the prior on `nec` and `ec50`.

Three consequences follow from keeping it. It was applied without notice, where
the response correction emits a message, so the predictor the fit used differed
from the predictor the user recorded. It was never reversed in the estimates it
changed; #93 has been open for that since 2021, and its second comment records
that the predictor is the half that matters. And its size depended on the design
rather than on the data, so the same control was fitted at a different value
depending on the lowest non-zero concentration tested.

**Three changes of behaviour follow, and all three are recorded in NEWS.** A fit
whose predictor includes an exact zero gives different estimates; the change is
small on the designs measured below, but it is a change. A `disp(~log(x))`
sub-model on a predictor including a zero, which was evaluated against the
substituted value and ran, now reaches Stan as `-Inf` and the fit does not run;
the offset has to be added to the data instead. Nothing validates a `disp`
formula for finiteness before `brm()` sees it, which is raised separately as
#271. And a fit whose data was corrected under an earlier version can no longer
be combined with a new fit of the same data, because `c()` and `+` on `bnecfit`
objects compare the stored data frames exactly; `amend()` keeps such a set
consistent in its data but gives an added model the new prior beside the old
one, so where the comparison between models matters the set should be refitted.

**The prior change reaches further than the zero case.** On a strictly positive
predictor of 0.5, 1, 3, 10, 30 and 100 replicated 1, 3, 6, 8, 2 and 1 times,
the `nec` prior changes from `gamma(5, 0.2)` to `gamma(5, 0.3077)` — a change in
its mean from 25 to 16.25 on a predictor running to 100. Unequal replication is
ordinary in this field, so this is the larger class of affected fits. A balanced
design is unaffected.

## Evidence

R 4.6.1, `brms` 2.23.0, cmdstanr 0.9.0 / cmdstan 2.39.0, `dev` at 130aede3.

**The prior did not require it.** `define_prior()` for `nec3param` under
`Gamma(link = "identity")` on `x = c(0, 5, 15, 45, 135)`:

```
with the correction     <lower=0.5,upper=135> gamma(5, 0.1333)
without the correction  <lower=0,  upper=135> gamma(5, 0.1333)
```

The prior is the same; only the truncation differs. `gamma(5, r)` has zero
density at zero, so `lower = 0` is a proper truncation.

**Stan did not require it.** The structural risk is the three equations that
raise the predictor to an estimated power — `ecxsigm` (`x^exp(d)`),
`nechormepwr` and `nechorme4pwr` (`x^(1 / (1 + exp(slope)))`) — where the
derivative with respect to the exponent is `pow(x, y) * log(x)`, undefined at
`x = 0`. A minimal Stan model containing `pow(x[n], exp(d))` with a zero in `x`
was checked with the cmdstan `diagnose` method: autodiff and finite differences
agree to 2e-8. This was predictable from the package's own code, since
`necsigm` computes `(step(x - nec) * (x - nec))^exp(d)`, whose base is exactly
zero for every observation below `nec`.

**A fit did not require it.** `nec3param` and `ecxsigm`, `Gamma(identity)`,
`x = 0, 5, 15, 45, 135` with six replicates, 2 chains x 2000 iterations,
`seed = 17`, fitted with the correction and with it disabled. All four fits
converged with zero divergent transitions.

| | with correction | without correction |
|---|---|---|
| `nec3param`, `nec` | 11.73 [5.54, 19.20] | 11.73 [5.93, 21.30] |
| `nec3param`, EC10 | 16.93 [11.13, 24.20] | 16.89 [11.08, 26.09] |
| `ecxsigm`, EC10 | 6.42 [3.73, 11.41] | 6.08 [3.11, 11.08] |

**What the correction was masking.** `define_prior()` built the `gamma` rate as
`1 / (quantile(predictor, 0.5) / 2)`. Where more than half the observations sit
at a zero control the median is zero and the prior string became
`gamma(5, Inf)`. On `x = c(rep(0, 12), 5, 15, 45, 135, 200, 300)`:

```
raw                     <lower=0,  upper=300> gamma(5, Inf)
with the correction     <lower=0.5,upper=300> gamma(5, 4)
on this branch          <lower=0,  upper=300> gamma(5, 0.0444)
```

The second has a prior mean for `nec` of 1.25 on a predictor running to 300, so
the correction converted a hard failure into a silently degenerate prior.

<details>

**`R/check_data.R`.** All three predictor corrections are removed: the
reachable `min(x) == 0 & x_type == "Gamma"` branch and the two branches #265
records as unreachable, `min(x) == 0 & x_type == "beta"` and
`max(x) == 1 & x_type == "beta"`. #265 is therefore closed by removal rather
than by repair; no released version applied those two, so nothing changes for a
0-1 bounded predictor.

The `set_distribution(x, ...)` call is kept, in statement position, rather than
deleted with the branches that consumed its value. It is also the
integer-predictor guard: `retrieve_var()` preserves the column type,
`is.numeric()` is `TRUE` for an integer, and the "bayesnec does not currently
support integer concentration data" error is raised from inside
`set_distribution()` when `silence_x_msgs = FALSE`. `define_prior()` makes the
same call and would raise the same error, so deleting this one would relocate
the error into prior construction rather than remove it; it belongs in the
documented data check, before any prior is built.
`tests/testthat/test-fit_bayesnec.R` asserts it of `check_data()` specifically,
so a future tidy-up of the apparently unused call fails a test rather than
passing quietly.

`x_transformed` is removed with the branches. Nothing in `check_data()` now
treats a transformed predictor differently from a bare one, which is the
correct outcome: the special case existed only because a correction could not
reach the fit through a transformation.

**`R/fit_bayesnec.R`.** The `write_back_checks(data, bdat, "x_var", x)` call is
removed. `check_data()` no longer corrects the predictor, so the write-back
could only restate what the column already held — the same reasoning #258
applied when it removed the trials write-back. `write_back_checks()` itself is
unchanged and is now called for the response alone.

**`R/define_prior.R`.** The `Gamma` entry of `x_prs` takes its rate from
`median(unique(predictor))` in place of `quantile(predictor, probs = 0.5)`. The
rule is that a prior scale should describe the concentrations that were tested,
not how many replicates each of them received. The precedent is inside the
package: `survival_by_x()` already primes the `hu` block of a hurdle or
zero-inflated fit from `sort(unique(predictor))`, so on an unbalanced design the
`mu` and `hu` blocks of one fit received different rates for the same
concentrations — `gamma(5, 0.4)` against `gamma(5, 0.1333)` on eight controls
with four replicates each at 5, 15, 45 and 135. They no longer differ for that
reason. They can still differ for a substantive one: the `mu` block is primed
from the non-zero subset, so a concentration at which every response is zero is
absent from its series, and on the same design with complete mortality at 135
the `mu` block gets `gamma(5, 0.2)`. That is intended — the `mu` block is fitted
only to the survivors. The two readings
agree for a balanced design, so no fit of that shape gets a different prior. The
distinct-value median cannot be zero while the predictor reaches above one,
which is the condition under which this entry is selected, so no guard against
a non-finite rate is added — an unreachable one would be noise.

The `gaussian` entry is left alone, deliberately. It applies where the
predictor takes negative values, a median of zero is a legitimate centre for a
normal so there is no failure to fix, and its scale term `sd(predictor) * 10`
is deliberately diffuse. Changing it would alter the prior on every log-scale
fit for no stated defect.

**Tests.** `tests/testthat/test-fit_bayesnec.R`: the two assertions that the
predictor arrives as `0.1` are inverted to `expect_identical` against the
recorded column, at both the `check_data()` level and the `fit_bayesnec()`
level; the `a zero on a transformed predictor is left where it is` case becomes
`check_data() leaves the predictor alone, transformed or not`, since the two
are no longer distinguished; and two cases are added, one for a 0-1 bounded
predictor keeping both its bounds and one for the integer-predictor error.
`tests/testthat/test-define_prior.R` gains four: the rate when most
observations are controls and the rate on an unbalanced but strictly positive
predictor, both stated as prior strings rather than recomputed from the
expression under test; the prior being unchanged for a balanced design, which
does recompute the old expression, since agreement with it is the claim; and
`nec` truncated at `lb = 0` rather than `lb = 0.1`, asserted at the
`get_priors()` level, which is where the changed bound is visible to a user.

The full suite passes: 0 failures, the skips are the usual `On CRAN` set, and
the single warning is the pre-existing one in `test-get_priors.R:239`.

**Documentation.** `?bnec`, under **Additional technical notes**, described the
offset verbatim and was the only remaining statement of it in the shipped
package; that paragraph is rewritten and `man/bnec.Rd` regenerated.
`vignettes/example1.Rmd` stated that `bayesnec` substitutes
`min(x[x > 0]) / 10` for a zero on the recorded axis; it now states that a zero
predictor is not corrected on either scale, that an offset for a logarithm is
the user's to choose, and what happens when the logarithm is written in `disp`
instead. `vignettes/example3.Rmd` described the `nec` prior as
having "maximum density (mean, see above) at the median value of the
predictor". That was wrong before this change and would have stayed wrong: the
prior is `gamma(5, 2 / m)`, whose mean is `2.5 * m` and whose maximum density is
at `2 * m`. The sentence now states the shape and the rate and both derived
quantities, and says why the distinct values are used. That the prior peaks at
`2 * m` is itself a defect — on a linearly spaced series the mode falls on the
upper truncation bound — but correcting the prior is a change to every default
`nec` and `ec50` and is raised separately as #273. Both `.Rmd` and `.Rmd.orig` are
updated. The NEWS entry written under #258 described the predictor correction
as a behaviour change of that release; since 2.2.0 is unreleased, that
paragraph is rewritten rather than contradicted two bullets later.

</details>

**Rendered vignette output is now stale.** `example2`, `example3`, `example4`
and `example6` all fit a predictor including an exact zero, so their committed
numbers and figures no longer match what the code produces. `example6` is the
one whose prior also changes: `nassarius$dose` is unbalanced, so its rate
changes from 3.333 to 4. `example1` was checked and its fitted predictors
include no exact zero. A precompile is not a merge gate under
`notes/implementation/00_protocol.md`, and #190 already tracks the rebuild.

**#93.** Its second comment records that the predictor adjustment is the half
that matters, and this change removes it. What remains of #93 is the response
correction, which does not affect an estimate on the predictor scale and is not
an offset that could be reversed in one. Of the three response corrections only
the `Gamma` zero shift is reported; the `beta` zero shift and the `beta` and
`zero_inflated_beta` one shift are silent, so the objection this change makes to
the predictor correction still stands for those two. #93 is left open rather than
closed here, because narrowing it is the maintainer's decision; a comment on #93
records the position and the measurement.

**Found while reviewing, raised separately.** #271, no `disp()` sub-model is
checked for finiteness before `brm()` evaluates it. #272,
`set_distribution()` returns `NULL` for an integer vector with negative values.
#273, the `nec`/`ec50` gamma prior peaks at twice the median predictor, so on a
linearly spaced series its mode sits on the upper truncation bound. #274,
`update()` with `newdata` reports a boundary correction and then discards it,
which is the #258 failure mode on a route #258 did not cover.

https://claude.ai/code/session_0115RVMa5riTtaaz2aGTZJot


beckyfisher:
## Review summary

Six independent review rounds were run against this branch, each by a session
with no knowledge of the ones before it beyond the written record. The change
converged: rounds 1 and 2 each found four items to fix before merge, round 3
found one, rounds 4 and 5 found two and one, all of them corrections to text
rather than to code, and round 6 found none. Round 5 ran the full test suite on
the branch and on `dev` for the first time.

```
branch   1852 pass   0 fail   110 skip   1 warning
dev      1844 pass   0 fail   110 skip   1 warning
```

The skip lists are byte-identical and the single warning
(`test-get_priors.R:239`) occurs on both. `test_dir()` does not set `NOT_CRAN`,
so the five files nearest the change were rerun with it set: 289 pass, 0 fail,
0 skip.

### What the reviews found in the code

Two defects in the change itself, both in the first draft:

1. **A `disp(~log(x))` sub-model on a zero predictor was a regression the first
   draft did not record.** That formula was evaluated against the substituted
   value and ran; it now reaches Stan as `-Inf` and the fit does not run.
   Verified with `brm()` mocked. The behaviour is correct but it is a fit that
   ran before, so it is in NEWS, in `vignette("example1")` and in a test. The
   absence of any finiteness check on a `disp` formula is #271.
2. **The prior change reaches every unequally replicated design, not only those
   with a zero.** On a strictly positive predictor of 0.5, 1, 3, 10, 30 and 100
   replicated 1, 3, 6, 8, 2 and 1 times, `gamma(5, 0.2)` becomes
   `gamma(5, 0.3077)`. Unequal replication is ordinary, so this is the larger
   class of affected fits, and neither the PR body nor NEWS had said so.

### What the reviews found in the text

Six statements that measurement contradicted:

- **`?bnec` still documented the removed offset verbatim**, and was the last
  description of it in the shipped package.
- **The reason given for keeping the `set_distribution()` call was false.**
  `define_prior()` raises the same error, so deleting the call relocates it
  rather than removing it. The call stays for a different reason: the data check
  is where it belongs.
- **`vignette("example3")` stated the wrong location for the `nec` prior** —
  "maximum density at the median value of the predictor". `gamma(5, 2 / m)` has
  mean `2.5 m` and mode `2 m`. That error predates this change; correcting it
  exposed #273, since on a linearly spaced series `2 m` is the largest
  concentration tested and the truncated prior rises across its whole range.
- **NEWS claimed the many-controls design failed under earlier versions.** It
  did not. The predictor correction fired before the median was taken, so
  `get_priors()` returned `gamma(5, 4)` silently — a `nec` prior mean of 1.25 on
  a predictor running to 300, against 112.5 here. That ninety-fold change is the
  largest single effect this change has on an existing fit and NEWS had not
  stated it.
- **"The two blocks of a hurdle fit now agree" was over-general.** They agree
  wherever every concentration retains a non-zero response. With complete
  mortality at the top concentration — the ordinary reason to fit a hurdle
  family — the `mu` block gets `gamma(5, 0.2)` against `gamma(5, 0.1333)`,
  because it is primed from the non-zero subset. That is intended behaviour;
  the claim was wrong.
- **A comment posted on #93 said the remaining response correction is
  reported.** `check_data()` makes three and only the `Gamma` zero shift emits
  anything; the `beta` zero shift and the `beta`/`zero_inflated_beta` one shift
  are silent. Measured, and now recorded on #93.

### What the reviews confirmed

- No model equation divides by the predictor or takes its logarithm. The three
  that raise it to an estimated power were fitted with an exact zero present,
  with no `NaN` in any draw; `necsigm` already computes `(step(x - nec) *
  (x - nec))^exp(d)`, whose base is exactly zero below `nec`, on every fit.
- Every path that rebuilds a prior from a stored fit — `amend()`,
  `update.bnecfit()`, `has_family_changed()`, `bnec_group()`, `bnec_hurdle()`,
  `bnec_joint()` — either runs through `check_data()` or reads the stored data
  frame, so none assumed a positive predictor independently.
- `median(unique(predictor))` cannot be reached with an `NA`, cannot be zero for
  any predictor the `Gamma` entry applies to, and agrees with the observation
  median for a balanced design over a 2-9 level by 1-8 replicate sweep.
- Every added test fails when the change it pins is reverted, checked by
  reinstating the removed code in scratch worktrees.
- `man/` and `NAMESPACE` regenerate byte-identically; `checkRd()` on the edited
  `.Rd` reports no note `dev` does not.

### Raised out of scope

Four issues, none caused by this branch:

- **#271** — no `disp()` sub-model is checked for finiteness before `brm()`
  evaluates it, so `disp(~log(x))` fails inside Stan rather than in
  `check_data()`.
- **#272** — `set_distribution()` returns `NULL` for an integer vector with
  negative values, so an integer response fails reporting a `family` argument
  the user did not supply.
- **#273** — the default `nec`/`ec50` gamma prior peaks at twice the median
  predictor, so on a linearly spaced series its mode sits on the upper
  truncation bound and the truncated prior rises across the whole range it
  permits.
- **#274** — `update()` with `newdata` reports a boundary correction and then
  discards it, which is the #258 failure mode on a route #258 did not cover.
  Demonstrated on the packaged `manec_example`.

Recorded rather than raised: the two silent response corrections, on #93, which
is the issue they belong to; and the stale rendered vignette output, on #190,
which already tracks the rebuild.

### Left for the maintainer

**#93.** #269's specification named it a stop-and-ask. The predictor half its
own second comment identifies as the important one is removed by this change,
and no response substitution is an offset that could be reversed in an estimate.
A comment on #93 records the position, the measurement, and two candidate
resolutions. It is not closed by this pull request.

beckyfisher:
## The failing Windows job is pre-existing on `dev`

`windows-latest (release)` fails on this branch and it fails identically on
`dev`, at `130aede3`, the commit the branch is based on. It is not caused by
this change.

```
dev    130aede3   [ FAIL 17 | WARN 1 | SKIP 4 | PASS 2246 ]
#270   d7bf1c4a   [ FAIL 17 | WARN 1 | SKIP 4 | PASS 2254 ]
```

The same 17 tests fail, at the same lines, in both — `test-bnec_hurdle.R:173`,
`test-check_fit.R:187, 199, 208, 218, 227`, `test-dispersion.R:25, 45, 56`,
`test-get_priors.R:149, 187`, `test-inits_functions.R:446, 546, 743, 796, 821`
and `test-zero-inflated-counts.R:122` — every one of them with
`Error: invalid connection` raised from `do.call(brm, all_args)`. The 8 extra
passes on the branch are the tests it adds. None of those files is touched by
this change except for one comment in `test-check_fit.R` and one appended test
at the end of `test-get_priors.R`.

macOS and both Ubuntu jobs pass on `dev` and on this branch.

The cause is in the workflow rather than in the package: the **Use Cmdstan to
Fix** step force-installs `StanHeaders`, and when `https://mc-stan.org/r-packages/`
is unreachable the call falls back to CRAN and installs 2.39.1 over the 2.32.10
that `pak` had installed to match `rstan` 2.32.7. Raised as #275, with the
evidence and three candidate fixes.

`dev` at `bc5f7d00` passed the same job on 2026-09-02 at 04:55 UTC with
`[ FAIL 0 | WARN 1 | SKIP 3 | PASS 2274 ]` and no StanHeaders reinstallation, so
the breakage arrived between then and 23:52 UTC the same day, with no package
change that touches Stan or the backend.

Issue #275, opened 2026-09-03
R-CMD-check fails on windows-latest: the workflow force-installs StanHeaders from CRAN over the version rstan was matched to


beckyfisher:
## What

`R-CMD-check` fails on `windows-latest` with 17 test errors, all of them
`invalid connection` raised from `brm()`. The cause is the **Use Cmdstan to
Fix** step, which force-installs `StanHeaders` from whichever repository
answers. When `https://mc-stan.org/r-packages/` is unreachable the call falls
through to CRAN and installs StanHeaders 2.39.1 over the 2.32.10 that `pak` had
just installed to match `rstan` 2.32.7. `rstan` then cannot compile a model
against headers three minor versions ahead of it.

## Why it matters

The check matrix is red on `dev`, so no pull request can show a clean matrix and
the Windows job stops distinguishing a real regression from this one. The
failure is also intermittent by construction: it depends on whether the
mc-stan.org index answers at the moment the job runs, so the same commit passes
or fails on different days with no change to the package.

## Evidence

The failure is not caused by any package change. `dev` at `130aede3` and PR #270
at `d7bf1c4a` give byte-identical failures:

```
dev    130aede3   [ FAIL 17 | WARN 1 | SKIP 4 | PASS 2246 ]
#270   d7bf1c4a   [ FAIL 17 | WARN 1 | SKIP 4 | PASS 2254 ]
```

The same 17 tests, at the same lines, in both:
`test-bnec_hurdle.R:173`, `test-check_fit.R:187, 199, 208, 218, 227`,
`test-dispersion.R:25, 45, 56`, `test-get_priors.R:149, 187`,
`test-inits_functions.R:446, 546, 743, 796, 821`,
`test-zero-inflated-counts.R:122`. The 8 extra passes on the branch are the
tests it adds. `invalid connection` appears 42 times in each log. Every
backtrace ends at the same place:

```
7. └─bayesnec:::fit_bayesnec(...) at bayesnec/R/bnec.R:624:5
8.   └─base::tryCatch(do.call(brm, all_args), error = fit_failed)
```

macOS and both Ubuntu jobs pass on `dev` and on the branch.

**When it started.** `dev` at `bc5f7d00`, 2026-09-02 04:55 UTC:
`[ FAIL 0 | WARN 1 | SKIP 3 | PASS 2274 ]`, StanHeaders 2.32.10 and no
reinstallation. `dev` at `130aede3`, the same day at 23:52 UTC: the failure
[...]

Issue #274, opened 2026-09-03
update() with newdata reports a boundary correction and then discards it, so the fit fails naming the condition the package said it had repaired


beckyfisher:
## What

`update()` on a `bnecfit` with `newdata` reports a boundary correction to the
user and then discards it. `has_family_changed()` runs `check_data()` on the new
data, which emits the "your response contains zeros ... they have been shifted"
message, and then reads only `checked_df$family` from the result. The corrected
data frame is dropped, and the user's raw `newdata` is passed to
`brms::update()`.

This is the #258 failure mode on a route #258 did not cover. The write-back that
fixed it lives in `fit_bayesnec()`, which this path does not go through.

## Why it matters

The user is told the package repaired their data and then the fit fails naming
the condition it said it had repaired, which is the symptom #258 was closed for.
Supplying new data to an existing fit is the ordinary way to refit the same
model set to a second experiment or to a corrected dataset, so the route is not
obscure.

## Evidence

`dev` at 130aede3, R 4.6.1, on the packaged `manec_example` with three response
values set to zero and `family = Gamma(link = "identity")`:

```r
data(manec_example)
f <- pull_out(manec_example, model = "nec4param")
d <- f$fit$data
d$y[1:3] <- 0

bayesnec:::has_family_changed(list(f), d, Gamma(link = "identity"))
#> Your response contains zeros, which a Gamma distribution cannot represent.
#> They have been shifted to 0.00208 (one tenth of the smallest non-zero value).
#> ...
#> [1] TRUE

sum(d$y == 0)
#> [1] 3
```

The message is emitted; `d` is unchanged. `R/bnecfit-methods.R:169-171` calls
that function and keeps only its logical result, and `R/bnecfit-methods.R:196-205`
builds the `brms::update()` call with `newdata = quote(newdata)`, the object the
user supplied.

`write_back_checks()` is called from `fit_bayesnec()` and from nowhere else, so
no correction reaches `brm()` on this path.

<details>
<summary>Implementation detail</summary>

`R/helpers.R:939-959`, `has_family_changed()`. It exists to answer one question
-- whether the data implies a different family -- and returns a logical, so the
corrected data frame it computes on the way has nowhere to go. It also runs
`check_data()` once per model in the set, so on a `bayesmanecfit` the message is
emitted once per model.

Two things to settle when this is taken up.

**Where the correction is applied.** Either `update.bnecfit()` gains its own
write-back, using `write_back_checks()` on the model frame of each fit's own
[...]

Issue #272, opened 2026-09-03
set_distribution() returns NULL for an integer vector with negative values, so an integer response fails reporting a family argument the user did not supply


beckyfisher:
## What

`set_distribution()` returns `NULL` for an integer vector containing negative
values. The integer branch tests `if (min(x) >= 0)` and has no `else`, so the
function falls off the end. Automatic family selection reads that `NULL` and
the call fails reporting the `family` argument, which the user did not supply.

## Why it matters

An integer response with negative values is ordinary input: a difference, an
increment, a change in a count between two times. The user sees

```
Error: Argument "family" either is not an actual family, or is of incorrect
       class.
```

and has supplied no `family` argument at all, so the message names something
they cannot act on and says nothing about the response. The equivalent numeric
response is handled -- `set_distribution()` returns `"gaussian"` for a numeric
vector with negative values -- so the two differ only in the storage type of
the column.

## Evidence

`dev` at 130aede3, R 4.6.1.

```r
bayesnec:::set_distribution(c(-3L, 0L, 4L), support_integer = TRUE)
#> NULL

d <- data.frame(x = rep(c(0.5, 1, 10, 100), each = 5),
                y = as.integer(rep(c(5, 2, 0, -3), each = 5)))
get_priors(y ~ crf(x, model = "nec3param"), data = d)
#> Error: Argument "family" either is not an actual family, or is of incorrect
#>        class.
```

Pre-existing; found while reviewing #270 and unrelated to that change.

<details>
<summary>Implementation detail</summary>

`R/set_distribution.R:43-55`. The integer branch is

```r
} else if (inherits(x, "integer")) {
  if (!support_integer && !silence_x_msgs) {
    stop(...)
  } else {
    if (min(x) >= 0) {
      if (is.null(trials)) "poisson" else "binomial"
    }
  }
}
```

The inner `if` has no `else`, so a negative minimum returns `NULL`. The value is
read at `R/helpers.R:625` (`retrieve_valid_family()`) and `R/bnec_group.R:121`,
both of which pass it straight to `validate_family()`.

Two candidate resolutions, and the choice is a decision rather than a detail.
Returning `"gaussian"` treats the column as a numeric one, which is what its
values are; the count families cannot represent a negative value, and no other
family `bayesnec` supports is both discrete and unbounded below. Erroring
instead names the response and asks the user to choose, on the grounds that an
integer column with negative values may equally be a coding error. Whichever is
[...]

Issue #271, opened 2026-09-03
No disp() sub-model is checked for finiteness before brm() sees it, so disp(~log(x)) on a zero predictor fails inside Stan rather than in check_data()


beckyfisher:
## What

Nothing checks that a `disp()` sub-model produces finite values before `brm()`
evaluates it. `check_data()` tests the predictor and the response for
finiteness and names the column when either fails, but it inspects only the
population variables `crf()` declares. A `disp` formula is an arbitrary `brms`
formula evaluated against the user's data frame, and an infinite value in it
reaches Stan.

## Why it matters

The two paths give the same user error two different diagnostics. A logarithm
of a predictor containing a zero, written in `crf()`, stops in `check_data()`
with the column named:

```
Your predictor column contains values that are not finite.
```

The same logarithm written in `disp()` produces a `brms` warning about the data
in general, after which the fit does not run, with nothing naming the term
responsible. `disp(~log(x))` is a documented form -- see
`?bayesnecformula` and `vignette("example1")` -- so this is reachable from
ordinary use.

## Evidence

R 4.6.1, `brms` 2.23.0. On
`d <- data.frame(x = rep(c(0, 1, 10, 100), each = 5), y = rep(c(8, 6, 3, 1), each = 5))`,
the formula `y ~ crf(x, model = "nec3param") + disp(~log(x))` builds the `brms`
formula `shape ~ log(x)`, and:

```r
sdata <- brms::make_standata(brms::bf(y ~ a, a ~ 1, shape ~ log(x), nl = TRUE),
                             data = d, family = Gamma(link = "identity"))
#> Warning: Found infinite values in the data, which may cause issues for Stan.
range(sdata$X_shape[, "logx"])
#> [1] -Inf 4.60517
```

The fit does not then run.

This became reachable with #269, which removed the substitution of
`min(x[x > 0]) / 10` for a zero predictor: the `disp` formula was previously
evaluated against the substituted value, so `log(0)` never arose. The gap
itself is older and is not confined to zeros -- any `disp` expression that is
not finite on the recorded data reaches Stan the same way, `disp(~ 1 / (x - 1))`
on a predictor including 1 among them.

<details>
<summary>Implementation detail</summary>

`R/check_data.R` tests `is.finite(mean(x))` and `is.finite(mean(y))` on the
values `retrieve_var()` returns for `x_var` and `y_var`. The columns a `disp`
formula constructs are not among them; `parse_disp_term()` (`R/disp_model.R`)
[...]

PR #276, opened 2026-09-03
Add test files for check_data(), plot() and autoplot()

### LINKED ISSUE #277: Write the three missing test files: check_data(), plot() and autoplot() have none in any release
## What

Write the three test files that do not exist in any release: `test-check_data.R`,
`test-plot.R` and `test-autoplot.R`.

## Why it matters

The code producing the current defects is the code with no test file. Measured
on 2026-09-03 against `origin/master` (2.1.3.1) and `dev`:

- `check_data.R` has no test file in either version. It is the origin of #258,
  #265, #269, #271 and #274 — five defects in two weeks.
- `plot.R` and `autoplot.R` have no test file. They are #268, #160 and #161.
- `xform` has **zero assertions anywhere on the plotting path**. It is asserted
  only on the estimators, in `test-ecx.R`, `test-nec.R`, `test-nsec.R`,
  `test-ecnsec.R` and `test-average_estimates.R`. #268 is exactly that gap, and
  the issue records that it could only be read from source.

Where a test file was written, the discovery stopped: `test-validate_family.R`
0 → 99 assertions, `test-inits_functions.R` 0 → 114, `test-fit_bayesnec.R`
0 → 34, `test-define_prior.R` 33 → 103.

A defect found by a user report is one defect. A test file written over the path
that produced it is the population on that path.

## Evidence

Nine of nine recent defects are present in the CRAN release, checked line by
line rather than taken from the issue text: #256 (`validate_family.R:14-15`),
#258 (`fit_bayesnec.R:36`), #265 (`check_data.R:55`, `:63`), #266
(`inits_functions.R:101`), #267 (`bnecfit-methods.R:150`), #268 (two sites in
`plot.R`, two in `autoplot.R`), #269 (`check_data.R:48`, `:56`), #272
(`set_distribution.R:48`), #244 (absent from `inits_functions.R`). One issue in
the run was caused by the run's own work: #271.

Suite-wide, assertion density per line of `R/` has roughly doubled since the
release, from 0.052 to 0.101.

## Expected outcome

**This issue is expected to open two or three further issues**, and that is the
point of it rather than a failure of it. Three findings already have no issue
of their own and are listed in the implementation detail below.

It is also expected to close or supersede nothing on its own: #268 and #274 are
pinned as current behaviour with the assertion to invert named in a comment, so
each fix becomes a one-line inversion plus the code change rather than a fresh
reproduction.

<details>

**Scope.** Three files, no fitting. Every assertion runs off `manec_example` and
hand-built model frames, so the suite cost is seconds rather than minutes.

- [ ] `tests/testthat/test-check_data.R` — specifies `check_data()` as a
      function rather than as a step: its guards, the arithmetic and the
      messaging of each response correction, the shape of what it returns, and
      what each of its three callers does with the result. Does not duplicate
      the write-back (`test-fit_bayesnec.R`), the censoring exemptions
      (`test-cens.R`), the hurdle zeros (`test-hurdle_family.R`) or
      `check_normalisation()` (its own file).
- [ ] `tests/testthat/test-autoplot.R` — `ggbnec_data()` return shape, `xform`
      on the fitted curve and on the raw data, and the #268 reproduction.
- [ ] `tests/testthat/test-plot.R` — the base-graphics paths, read by plotting
      to a null `pdf()` device and reading `par("usr")`, which is enough to
      assert whether `xform` reached the predictor axis without comparing
      images.

**The #268 reproduction needs no fit.** Give a stored fit a formula that
transforms its response — `fit$bayesnecformula <- bnf(exp(y) ~ crf(x, model =
"nec4param"))` — and `find_transformations()` reports `"y"` while
`pop_var_is_transformed(bdat, "x_var")` is `FALSE`. `xform` is then silently
skipped on all four sites. `exp()` rather than `log()`: `manec_example`'s
response reaches -6.9.

**Three findings with no issue yet.** Raise as one issue when this merges.

- **Two branches of `check_data()` cannot be reached.** The non-numeric
  predictor branch (`:112-118`) is preceded by `retrieve_var(error = TRUE)` at
  `:108`, which raises first with a different message. The numeric group-level
  branch (`:203-209`) is preceded by `model.frame()`, which refuses first. Both
  are the shape of #265 — a condition written, never fired, unnoticed for years.
- **`NA` and `NaN` are dropped rather than refused.** The finiteness guard sees
  only what `model.frame()` passes it, and incomplete cases are removed first.
  `Inf` reaches the guard and is refused; `NA` and `NaN` are removed silently
  and the fit proceeds on fewer rows than the user supplied, with no message.
- **`ggbnec_data(x, nec = FALSE)` does nothing.** The argument is `add_nec`;
  `autoplot()` takes `nec` and forwards it as `add_nec`. `ggbnec_data()` is
  exported and documented separately, so the obvious transfer of the argument
  name is absorbed by `...` and the annotation is still computed and returned.

**Definition of done.** The three files exist, the full suite passes with no new
warnings, and every assertion that pins a defect rather than correct behaviour
says so in a comment naming the issue and the inversion.

**Stop and ask rather than guess** if a test cannot be written without first
deciding what the correct statistical behaviour is. Pin the current behaviour
with a comment and raise it instead; that boundary is `00_protocol.md`'s.

</details>

beckyfisher:
Closes #277.

Adds the three test files that have never existed in any release:
`test-check_data.R`, `test-plot.R` and `test-autoplot.R`. No package code is
changed.

## What

66 assertions over three new files. Nothing fits a model — every assertion runs
off `manec_example` and hand-built model frames, so the suite cost is seconds.

## Why it matters

The code producing the current defects is the code with no test file. Measured
on 2026-09-03 against `origin/master` (2.1.3.1) and `dev`:

- `check_data.R` has no test file in either version. It is the origin of #258,
  #265, #269, #271 and #274 — five defects in two weeks.
- `plot.R` and `autoplot.R` have no test file. They are #268, #160 and #161.
- **`xform` has zero assertions anywhere on the plotting path.** It is asserted
  only on the estimators. #268 is exactly that gap.

Where a test file was written, the discovery stopped: `test-validate_family.R`
0 → 99 assertions, `test-inits_functions.R` 0 → 114, `test-fit_bayesnec.R`
0 → 34, `test-define_prior.R` 33 → 103.

## Evidence

Nine of nine recent defects are present in the CRAN release, checked line by
line rather than taken from the issue text: #256 (`validate_family.R:14-15`),
#258 (`fit_bayesnec.R:36`), #265 (`check_data.R:55`, `:63`), #266
(`inits_functions.R:101`), #267 (`bnecfit-methods.R:150`), #268 (two sites in
`plot.R`, two in `autoplot.R`), #269 (`check_data.R:48`, `:56`), #272
(`set_distribution.R:48`), #244 (absent from `inits_functions.R`).

**#268 is reproduced here without fitting anything**, which the issue records as
not having been possible: giving a stored fit a formula that transforms its
response makes `find_transformations()` report `"y"` while the predictor is
untransformed, and `xform` is then silently skipped at all four sites. The base
paths are read by plotting to a null `pdf()` device and reading `par("usr")`.

## Three findings with no issue yet

- **Two branches of `check_data()` cannot be reached.** The non-numeric
  predictor branch (`:112-118`) is preceded by `retrieve_var(error = TRUE)` at
  `:108`, which raises first with a different message; the numeric group-level
  branch (`:203-209`) is preceded by `model.frame()`. Both are the shape of
  #265 — a condition written, never fired, unnoticed for years.
- **`NA` and `NaN` are dropped rather than refused.** The finiteness guard sees
  only what `model.frame()` passes it. `Inf` reaches it and is refused; `NA` and
  `NaN` are removed silently and the fit proceeds on fewer rows than supplied.
- **`ggbnec_data(x, nec = FALSE)` does nothing.** The argument is `add_nec`.
  `autoplot()` takes `nec` and forwards it, so the obvious transfer of the
  argument name is absorbed by `...`.

All three are pinned as current behaviour, not asserted to be correct.

## Two tests pin a defect rather than correct behaviour

Each says so in a comment naming the issue and the assertion to invert:

- **#268** — a transformed response suppresses `xform` on the predictor axis.
  Pinned on all four sites: two in `test-plot.R`, two in `test-autoplot.R`.
- **#274** — `has_family_changed()` reports a boundary correction and discards
  it, so `update()` with `newdata` fails on the condition it reported repaired.

Each fix becomes a one-line inversion plus the code change, rather than a fresh
reproduction.

## What is deliberately not duplicated

| already covered | where |
|---|---|
| the `check_data()` write-back into `brm()`'s data | `test-fit_bayesnec.R` |
| the censoring exemptions | `test-cens.R` |
| zeros preserved for a hurdle family | `test-hurdle_family.R` |
| `check_normalisation()` | `test-check_normalisation.R` |

## Verification

`NOT_CRAN=true`, R 4.6.1, `dev` at 9b8a4067.

```
test-check_data.R   PASS 37   FAIL 0
test-autoplot.R     PASS 18   FAIL 0
test-plot.R         PASS 11   FAIL 0
```

Full suite, `test_dir()` with `NOT_CRAN=true`:

```
FAIL 0 | WARN 2 | SKIP 0 | PASS 2390
```

Both warnings are pre-existing and neither is in a file this branch touches:
`test-disp_model.R:412` (a `loo` `p_waic` warning) and `test-get_priors.R:239`.
The three new files report zero warnings between them.

`DESCRIPTION` is bumped 2.1.3.25 -> 2.1.3.26 to keep the dev counter monotonic.
**No `NEWS.md` entry**: nothing here changes what a user of the package sees,
and a bullet saying tests were added would be noise in a release note. Flagged
because `00_protocol.md` item 4 asks for one by default.

**The Windows job is expected to be red.** #275: `R-CMD-check` force-installs
`StanHeaders` over the version `rstan` was matched to, and the failure is
identical on `dev`. It is not caused by this branch.

## Left undone

The three findings above have no issue of their own yet; #277 records them, and
they should be raised as one issue once this merges. `01_work_queue.md` records
them so they are not lost.

Note that `Closes #277` will not fire on merge, because this pull request
targets `dev` rather than the default branch. Close it by hand.

https://claude.ai/code/session_01QN6wLkaDNhzpwVr3ffFAKQ

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,7 +28,7 @@ Suggests:
     knitr,
     rmarkdown,
     R.utils,
-    testthat (>= 3.1.7)
+    testthat (>= 3.1.8)
 VignetteBuilder: knitr
 URL: https://open-aims.github.io/bayesnec/
 BugReports: https://github.com/open-aims/bayesnec/issues
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,7 +28,7 @@ Suggests:
     knitr,
     rmarkdown,
     R.utils,
-    testthat (>= 3.0.0)
+    testthat (>= 3.1.7)
 VignetteBuilder: knitr
 URL: https://open-aims.github.io/bayesnec/
 BugReports: https://github.com/open-aims/bayesnec/issues
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: bayesnec
 Title: A Bayesian No-Effect- Concentration (NEC) Algorithm
-Version: 2.1.3.25
+Version: 2.1.3.26
 Authors@R: c(person("Rebecca", "Fisher", email = "r.fisher@aims.gov.au", role = c("aut", "cre"),comment = c(ORCID = "0000-0001-5148-6731")), person("Diego R.","Barneche",role="aut",comment = c(ORCID = "0000-0002-4568-2362")), person("Gerard F.","Ricardo",role="aut",comment = c(ORCID = "0000-0002-2220-3971")), person("David R.","Fox",role="aut",comment = c(ORCID = "0000-0002-3178-7243")))
 Description: Implementation of No-Effect-Concentration estimation that uses 'brms' (see Burkner (2017)<doi:10.18637/jss.v080.i01>; Burkner (2018)<doi:10.32614/RJ-2018-017>; Carpenter 'et al.' (2017)<doi:10.18637/jss.v076.i01> to fit concentration(dose)-response data using Bayesian methods for the purpose of estimating 'ECx' values, but more particularly 'NEC' (see Fox (2010)<doi:10.1016/j.ecoenv.2009.09.012>), 'NSEC' (see Fisher and Fox (2023)<doi:10.1002/etc.5610>), and 'N(S)EC (see Fisher et al. 2023<doi:10.1002/ieam.4809>). A full description of this package can be found in Fisher 'et al.' (2024)<doi:10.18637/jss.v110.i05>. This package expands and supersedes an original version implemented in 'R2jags' (see Su and Yajima (2020)<https://CRAN.R-project.org/package=R2jags>; Fisher et al. (2020)<doi:10.5281/ZENODO.3966864>).
 Depends:
--- a/tests/testthat/setup.R
+++ b/tests/testthat/setup.R
@@ -50,7 +50,7 @@ transformed_response_fit <- function(fit, model) {
 }
 
 # The same fixture for a model set. Both plotting paths read the formula off
-# mod_fits[[1]] alone (R/plot.R:258, R/autoplot.R:346), so that is the only
+# mod_fits[[1]] alone (R/plot.R:258, R/autoplot.R:347), so that is the only
 # element that has to change.
 transformed_response_manec <- function(manec) {
   mod <- names(manec$mod_fits)[1]
--- a/tests/testthat/setup.R
+++ b/tests/testthat/setup.R
@@ -53,9 +53,12 @@ transformed_response_fit <- function(fit, model) {
 # whether a variable was transformed, and it is not safe for reading posterior
 # quantities off the fit. Do not reuse it for anything else.
 
-# The same fixture for a model set. Both plotting paths read the formula off
+# The same fixture for a model set. On the model-average branch, which is the
+# one the pinning tests use, both plotting paths read the formula off
 # mod_fits[[1]] alone (R/plot.R:258, R/autoplot.R:347), so that is the only
-# element that has to change.
+# element that has to change. Not so with all_models = TRUE: plot() then draws
+# each candidate through plot.bayesnecfit, which reads that candidate's own
+# formula at R/plot.R:118. Do not use this fixture on that branch.
 transformed_response_manec <- function(manec) {
   mod <- names(manec$mod_fits)[1]
   manec$mod_fits[[1]] <- transformed_response_fit(manec$mod_fits[[1]], mod)
--- a/tests/testthat/setup.R
+++ b/tests/testthat/setup.R
@@ -48,6 +48,10 @@ transformed_response_fit <- function(fit, model) {
   )
   fit
 }
+# NB: the returned object states a formula its stored fit was not fitted with.
+# That is safe for the plotting paths, which read the formula only to decide
+# whether a variable was transformed, and it is not safe for reading posterior
+# quantities off the fit. Do not reuse it for anything else.
 
 # The same fixture for a model set. Both plotting paths read the formula off
 # mod_fits[[1]] alone (R/plot.R:258, R/autoplot.R:347), so that is the only
--- a/tests/testthat/setup.R
+++ b/tests/testthat/setup.R
@@ -28,3 +28,39 @@ nec4param <- pull_out(manec_example, model = "nec4param") |>
 ecx4param <- pull_out(manec_example, model = "ecx4param") |>
   suppressMessages() |>
   suppressWarnings()
+
+# Give a stored fit a formula that transforms its RESPONSE and nothing else.
+# This is the fixture that reproduces #268 without fitting anything: no packaged
+# fit transforms its response, and find_transformations() answers for the
+# formula as a whole, so a transformation here is enough to make the plotting
+# paths treat the untransformed predictor as transformed. exp() rather than
+# log(): manec_example's response reaches -6.9 and log() of it is NaN, and only
+# the presence of a transformation matters.
+#
+# Shared by test-plot.R and test-autoplot.R, which pin the same defect on the
+# base-graphics and ggplot2 paths and would otherwise define it twice.
+transformed_response_fit <- function(fit, model) {
+  # The model name is substituted into the formula text rather than referenced:
+  # crf() evaluates its model argument where the formula is used, not where it
+  # is written, so a variable reference is out of scope by then.
+  fit$bayesnecformula <- bayesnecformula(
+    stats::as.formula(paste0("exp(y) ~ crf(x, model = \"", model, "\")"))
+  )
+  fit
+}
+
+# The same fixture for a model set. Both plotting paths read the formula off
+# mod_fits[[1]] alone (R/plot.R:258, R/autoplot.R:346), so that is the only
+# element that has to change.
+transformed_response_manec <- function(manec) {
+  mod <- names(manec$mod_fits)[1]
+  manec$mod_fits[[1]] <- transformed_response_fit(manec$mod_fits[[1]], mod)
+  manec
+}
+
+# The largest predictor value the ggplot2 path put in its plotting frame. Used
+# by test-autoplot.R, and by test-plot.R to compare the two paths' xform
+# decisions against each other.
+gg_x_max <- function(obj, ...) {
+  max(suppressMessages(ggbnec_data(obj, ...))$x_e, na.rm = TRUE)
+}
--- a/tests/testthat/test-autoplot.R
+++ b/tests/testthat/test-autoplot.R
@@ -41,6 +41,13 @@ test_that("ggbnec_data returns the same frame shape for a model set", {
   skip_on_cran()
   d <- suppressMessages(ggbnec_data(manec_example))
   expect_true(all(c("x_e", "y_e", "y_ci", "x_r", "y_r") %in% names(d)))
+  # add_nec on the model-set branch, R/autoplot.R:356. Asserted here rather
+  # than in a block of its own so the frame above is computed once: making that
+  # condition unconditional failed nothing until this line was added, while its
+  # bayesnecfit twin at :312 was already observed.
+  expect_true("nec_vals" %in% names(d))
+  without <- suppressMessages(ggbnec_data(manec_example, add_nec = FALSE))
+  expect_false("nec_vals" %in% names(without))
 })
 
 test_that("the nec annotation is present by default and suppressible", {
@@ -73,8 +80,16 @@ test_that("xform is applied to the predictor axis of a single fit", {
 
 test_that("xform is applied to the predictor axis of a model set", {
   skip_on_cran()
-  expect_equal(gg_x_max(manec_example, xform = function(x) x * 100),
-               gg_x_max(manec_example) * 100, tolerance = 1e-8)
+  plain <- suppressMessages(ggbnec_data(manec_example))
+  scaled <- suppressMessages(ggbnec_data(manec_example,
+                                         xform = function(x) x * 100))
+  expect_equal(max(scaled$x_e, na.rm = TRUE),
+               max(plain$x_e, na.rm = TRUE) * 100, tolerance = 1e-8)
+  # And the raw data column, R/autoplot.R:354. Dropping x_r from that mutate()
+  # failed nothing until this line was added, while its bayesnecfit twin at
+  # :310 was already observed.
+  expect_equal(max(scaled$x_r, na.rm = TRUE),
+               max(plain$x_r, na.rm = TRUE) * 100, tolerance = 1e-8)
 })
 
 test_that("xform reaches the raw data as well as the fitted curve", {
--- a/tests/testthat/test-autoplot.R
+++ b/tests/testthat/test-autoplot.R
@@ -1,0 +1,149 @@
+# autoplot() and ggbnec_data() had no test file of their own in any release.
+# The eleven plot-related lines scattered through test-bayesnec_methods.R,
+# test-bayesmanec_methods.R and test-expand_classes.R assert that a ggplot
+# comes back, and nothing else -- in particular xform, which exists so that a
+# fit on a transformed predictor can be drawn on the recorded scale, has never
+# been asserted anywhere on the plotting path. It is asserted only on the
+# estimators, in test-ecx.R, test-nec.R, test-nsec.R, test-ecnsec.R and
+# test-average_estimates.R.
+#
+# That gap is #268: the decision to apply xform is made with an all-or-nothing
+# guard on the formula as a whole, so a transformation on the RESPONSE
+# suppresses xform on the PREDICTOR axis. The issue records that it could only
+# be read from source, because no packaged fit transforms its response. It can
+# in fact be reproduced without fitting anything, by giving a stored fit a
+# formula that transforms its response, and that is done below.
+#
+# Nothing here fits a model. Every assertion runs off manec_example.
+
+ap_fit <- function() {
+  data(manec_example, package = "bayesnec")
+  suppressMessages(pull_out(manec_example, model = "nec4param"))
+}
+
+ap_manec <- function() {
+  data(manec_example, package = "bayesnec")
+  manec_example
+}
+
+# Give a stored fit a formula that transforms its response and nothing else.
+# exp() is used rather than log() because manec_example's response reaches
+# -6.9, and log() of it is NaN; the point is only that a transformation is
+# present on the response, not which one.
+ap_transform_response <- function(fit) {
+  fit$bayesnecformula <- bayesnecformula(exp(y) ~ crf(x, model = "nec4param"))
+  fit
+}
+
+ap_x_max <- function(obj, ...) {
+  max(suppressMessages(ggbnec_data(obj, ...))$x_e, na.rm = TRUE)
+}
+
+
+# ---- what ggbnec_data returns -----------------------------------------------
+
+test_that("ggbnec_data returns the columns autoplot draws from", {
+  d <- suppressMessages(ggbnec_data(ap_fit()))
+  expect_true(all(c("x_e", "y_e", "y_ci", "x_r", "y_r") %in% names(d)))
+  expect_gt(nrow(d), 0)
+})
+
+test_that("ggbnec_data returns the same frame shape for a model set", {
+  d <- suppressMessages(ggbnec_data(ap_manec()))
+  expect_true(all(c("x_e", "y_e", "y_ci", "x_r", "y_r") %in% names(d)))
+})
+
+test_that("the nec annotation is present by default and suppressible", {
+  with_nec <- suppressMessages(ggbnec_data(ap_fit()))
+  without <- suppressMessages(ggbnec_data(ap_fit(), add_nec = FALSE))
+  expect_true("nec_vals" %in% names(with_nec))
+  expect_false("nec_vals" %in% names(without))
+})
+
+test_that("ggbnec_data takes add_nec, and absorbs nec = FALSE into dots", {
+  # autoplot() takes nec = and forwards it as add_nec =. ggbnec_data() is
+  # exported and documented separately, and the obvious transfer of the
+  # autoplot argument name does nothing at all: nec = FALSE is swallowed by
+  # ... and the annotation is still computed and returned. Pinned as current
+  # behaviour rather than asserted to be correct.
+  swallowed <- suppressMessages(ggbnec_data(ap_fit(), nec = FALSE))
+  expect_true("nec_vals" %in% names(swallowed))
+})
+
+
+# ---- xform on the predictor axis --------------------------------------------
+
+test_that("xform is applied to the predictor axis of a single fit", {
+  f <- ap_fit()
+  expect_equal(ap_x_max(f, xform = function(x) x * 100),
+               ap_x_max(f) * 100, tolerance = 1e-8)
+})
+
+test_that("xform is applied to the predictor axis of a model set", {
+  m <- ap_manec()
+  expect_equal(ap_x_max(m, xform = function(x) x * 100),
+               ap_x_max(m) * 100, tolerance = 1e-8)
+})
+
+test_that("xform reaches the raw data as well as the fitted curve", {
+  f <- ap_fit()
+  plain <- suppressMessages(ggbnec_data(f))
+  scaled <- suppressMessages(ggbnec_data(f, xform = function(x) x * 100))
+  expect_equal(max(scaled$x_r, na.rm = TRUE),
+               max(plain$x_r, na.rm = TRUE) * 100, tolerance = 1e-8)
+})
+
+test_that("the transformation predicate answers per variable", {
+  f <- ap_transform_response(ap_fit())
+  bdat <- model.frame(f$bayesnecformula, data = f$fit$data,
+                      run_par_checks = TRUE)
+  # The predictor is not transformed in this formula and the per-variable
+  # predicate says so; find_transformations(), which the plotting paths read,
+  # answers for the formula as a whole and reports the response.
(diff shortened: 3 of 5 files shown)


beckyfisher:
Ten review rounds against the branch, each run on a separate checkout with no
prior context. All findings are addressed in the ten commits that follow
`aa5ebae2`. No package code under `R/` is changed by this branch;
`git diff origin/dev...HEAD -- R/` is empty.

Rounds 3 to 10 found more in the review fixes than in the original three files.
Seven assertions that could not fail were found, three of them introduced by
earlier rounds, and twelve sites in `R/` that no assertion observed at all.

The recurring shape is a `bayesmanecfit` site left uncovered after its
`bayesnecfit` twin was covered. Rounds 6, 7, 8, 9 and 10 each found an instance
after the previous round had corrected another — five in all. Round 10
enumerated every `bayesnecfit`/`bayesmanecfit` pair of code paths in
`R/plot.R` and `R/autoplot.R` and mutation-tested each; the three it found are
closed, and the enumeration is the reason for thinking the set is now complete
rather than merely that the next round found nothing. They are listed below rather than
quietly repaired, because the recurrence is itself the result worth recording.

## What

The description of #268 in the opening post is wrong, and the four assertions
that pinned it were pinning half of the defect. `xform` is not skipped when the
formula transforms the response. It is skipped on the **axis** and still
applied to the **annotation**.

There are twelve sites, not four. The four the opening post counts are the
guards; each has two annotations beside it that the guard does not cover.

| site | inside the `length(trans_vars) == 0` guard |
|---|---|
| `R/plot.R:125` — the axis, single fit | yes |
| `R/plot.R:131` — `nec`, single fit | no |
| `R/plot.R:130` — `ec10`, single fit | no |
| `R/plot.R:279` — the axis, model set | yes |
| `R/plot.R:283` — `nec`, model set | no |
| `R/plot.R:284` — `ec10`, model set | no |
| `R/autoplot.R:308` — the curve, single fit | yes |
| `R/autoplot.R:313` — `bind_nec()`, single fit | no |
| `R/autoplot.R:316` — `ecx()`, single fit | no |
| `R/autoplot.R:352` — the curve, model set | yes |
| `R/autoplot.R:357` — `bind_nec()`, model set | no |
| `R/autoplot.R:360` — `ecx()`, model set | no |

Each of the twelve is now pinned by an assertion that fails when that line
alone is brought inside its guard. Four of the eight annotation sites were
observed by nothing until rounds 6, 7 and 8.

## Why it matters

Where the caller supplies an `xform` — with the default `identity` the two
halves coincide and nothing is visible — the two halves of the figure are drawn
on different scales, so the NEC line is not on the figure at all. That is a
stronger symptom than the issue records, and it is what a user would report.

For this pull request the consequence is narrower than first stated, and the
first statement of it was wrong. The original four assertions read the axis
alone. A change to the axis guard alone does fail one of them, so that half was
pinned. A change to the **annotation** alone failed none of the four, so that
half was not pinned at all, and a fix that corrected the annotation to match
the untransformed axis would have made the figure internally consistent, left
it on the wrong scale, and passed. Assertions were added until every one of the twelve sites has one that fails
when that line alone is brought inside its guard, so the comment instructing
the next reader to invert it names a test that fails. That is site-by-site
sensitivity, not independence: a change to an axis guard also fails the
annotation pins beside it, because the annotation pins compare the annotation
against the axis.

## Evidence

Measured on `nec4param` with a response transformed by `exp()` and
`xform = function(x) x * 100`:

- `ggbnec_data()` returns `max(x_e) = 3.22052` with and without `xform`, while
  the largest `nec_vals` changes from 1.5278 to 152.78.
- On the base path `par("usr")[2]` is 3.348047 and `abline()` is called with
  `v = c(146.1036, 136.083, 152.7784)`.

The pinning set was then checked by mutation rather than by inspection. Each
row was run twice, on separate checkouts by separate reviewers, with the
mutation applied to a copy of `R/` and the repository left unedited:

| mutation to `R/` | assertions that fail |
|---|---|
| the `R/plot.R:125` axis guard alone | 4, including the cross-path one |
| the `nec` annotations moved inside the guard, axis untouched | exactly the 4 that pin them |
| the `ec10`/`ecx` annotations moved inside the guard, `nec` untouched | exactly the 4 that pin them |
| the model-set `ec10`/`ecx` alone | exactly the 2 that pin them |

Both mutations were reverted and `R/` confirmed unchanged.

## CRAN gating of the plotting tests

`skip_on_cran()` was added to the plotting tests, and to the one
`test-check_data.R` block that reads a stored fit. Six `test_that` blocks
elsewhere in the suite call `plot()`, `autoplot()` or `ggbnec_data()` and five
are gated; the ungated one is `test-bayesnec_methods.R:16`. Restricting that to
the fit plotting path — the two `test-check_fit.R` blocks call `plot.checkfit`,
a different method — it is three gated of four. Three of the five gated blocks
write `if (Sys.getenv("NOT_CRAN") == "") { skip_on_cran() }` rather than the
bare call used here. The two agree for the unset case this paragraph concerns,
since `skip_on_cran()` already tests `NOT_CRAN`; they differ only if `NOT_CRAN`
is set to something other than a true value, when the bare call skips and the
wrapper does not.

The consequence, which the opening post does not state and which is recorded
here because it is not obvious from the diff: under a genuine CRAN check the
branch adds no plotting coverage. Measured with `test_dir(load_package = "none")` and
`NOT_CRAN` unset, `test-autoplot.R` skips all 16 of its blocks and `test-plot.R` all 14 of its
own; `test-check_data.R` skips 1 block of 23 and runs 35 of its 38
assertions. The whole of the #268
pinning is inert there.

It is not inert in CI. An `r-lib/actions` step earlier in the job exports
`NOT_CRAN: true` job-wide — it is in the environment of the
`setup-r-dependencies` step, before `check-r-package` starts — so every
`skip_on_cran()` in the suite is inactive on every push. Run 33741479642
reports `SKIP 3 | PASS 2382`, and those three skips are `skip_on_ci()`, not
`skip_on_cran()`: `test-component-args.R:67`, `test-disp_model.R:397` and
`:423`. None of the `skip_on_cran()` call sites fired.

Two qualifications on that evidence. It was measured on `aa5ebae2`; the gating,
the `setup.R` fixtures and the mocked assertions are in the ten commits that
follow, whose first CI run is the one this push triggers. And the property the gating
depends on comes from `setup-r@v2-branch` and `setup-r-dependencies@v2`, both
moving refs, so it is not pinned by this workflow.

The alternative considered was leaving the tests ungated. What that would add
to a CRAN check is the two fully gated files plus one `test-check_data.R`
block, about 76 s, not the 84 s the three files take, since
`test-check_data.R` runs on CRAN either way. The gating was kept, for
consistency with the rest of the suite and because CI exercises these tests on
every push.

## Findings with no issue yet

The opening post lists three. Rounds 7 to 10 add four more, all the same
shape as the two dead `check_data()` branches already listed
— a condition written that cannot fire, or that does not test what it reads as
testing. Both are pinned as current behaviour, not asserted to be correct.

- **The `lxform` branch cannot be reached.** `R/plot.R:152-157` and `:304-309`
  compose an axis call for the case where `lxform` is not a function.
  `R/plot.R:80-82` has already stopped on that input and `lxform` is not
  reassigned between them, so the condition is always FALSE and the `else` at
  `:164` always runs. `axis(side = 1)` and `axis(side = 1, at = signif(xticks,
  2))` are dead code, and `xticks` is always applied through the other branch.
- **`position_legend` refuses the value its documentation names, and its guard
  reads backwards.** `R/plot.R:42-43` documents a numeric vector "as per a call
  to legend"; the guard accepts only the eight keyword strings, so
  `position_legend = c(0.5, 0.5)` is refused. The arguments to `match()` are
  also the wrong way round, so `c("topright", "bogus")` passes the guard and
  fails inside `legend()` with `'arg' must be of length 1`, naming neither the
  argument nor the valid values. Two call sites, `:89-91` and `:232-235`, with
  differently worded messages; both are asserted.
- **`ggbnec_data()` validates `xform` on one class and not the other.**
  `R/autoplot.R:293-294` stops with "xform must be a function." for a
  `bayesnecfit`; the `bayesmanecfit` method has no such guard, so
  `ggbnec_data(manec_example, xform = "notafunction")` fails inside
  `dplyr::mutate()` with `could not find function "xform"`, naming neither the
  argument nor the requirement. Neither half is asserted here.
- **A return value assigned and never read.** `R/check_data.R:211` assigns
  `custom_name <- check_custom_name(family)` and nothing reads it;
  `check_custom_name()` is pure, so the call has no effect. The same dead
  assignment is at `R/plot.R:93`, `:263` and `R/autoplot.R:178`.

These should be raised as one issue with the three already recorded, making
seven.

## Corrections to the opening post

Six claims in the post are now wrong and the body should be edited rather than
left standing:

- "`xform` is then silently skipped at all four sites" — there are eight
  sites; `xform` is skipped at the four guarded ones and applied at the four
  beside them. The table above replaces it.
- "66 assertions over three new files" — 102.
- `test-autoplot.R PASS 18` — 23.
- `test-plot.R PASS 11` — 41.
- `test-check_data.R PASS 37` — 38.
- "#268 ... Pinned on all four sites: two in `test-plot.R`, two in
  `test-autoplot.R`" — there are eight sites and each is now pinned: the axis,
  the `nec` annotation and the `ecx` annotation, for a single fit and for a
  model set.
- "the suite cost is seconds" — the three files run in 84 s.

## Changes made

Eight kinds of change, in full below the fold:

1. eight assertions added, pinning the eight annotation sites; the four axis
   sites were already pinned at `aa5ebae2`. Four of the eight were observed by
   nothing until rounds 6, 7 and 8: the model-set `nec` at `R/plot.R:283` and
   `R/autoplot.R:357`, and the model-set `ec10`/`ecx` at `R/plot.R:284` and
   `R/autoplot.R:360`. `add_ecx` had no assertion anywhere in the suite;
2. **six assertions that could not fail**, replaced by ones that can — a
   tautology in the #274 test, a `bot` prior selection that is `character(0)`
   for `nec3param`, an `expect_silent()` set that any argument value satisfied,
   a count of `abline()` calls that does not identify which was drawn, an
   `nlpar` comparison insensitive to the correction, and a cross-path test on
   a fixture where the guard never fires;
3. four duplicated assertions dropped, and the fixtures `setup.R` already
   supplies used instead of rebuilt;
4. `skip_on_cran()` added, with the consequence stated above;
5. six line references and two quoted measurements corrected;
6. `DESCRIPTION` requires `testthat (>= 3.1.8)` rather than `3.0.0`, because
   3.1.7 cannot mock an imported binding;
7. prose corrections in the test comments;
8. two further unreachable or self-contradicting guards in `R/plot.R` pinned,
   listed under "Findings with no issue yet" below.

<details>

### The added assertions

`test-plot.R`, "the nec annotation is drawn off the end of the axis". It
intercepts `abline()` with `local_mocked_bindings(.package = "bayesnec")`
rather than recomputing the expected value. Recomputing it would assert a
number the test worked out for itself, which would still pass if
`R/plot.R:130-131` changed — the mutation it exists to catch. Mocking this way
is already the suite's convention: `test-fit_bayesnec.R:31` and
`test-inits_functions.R:385` both mock with `.package = "bayesnec"`.
(`test-hurdle_family.R:177` and `:191` mock without it, which is a different
mechanism.)
`abline` is imported at `NAMESPACE:216` rather than defined in the namespace,
which is why the `DESCRIPTION` floor below has to be 3.1.8.

`test-plot.R` mocks `legend()` and `abline()` the same way for `all_models`,
`add_nec` and `add_ec10`. Asserting only that those calls are silent passes
identically for every value of each argument, so it would not detect the
argument being ignored; the mocked calls say which branch of
`R/plot.R:179-192` was taken and which candidates `R/plot.R:253` named.

`test-autoplot.R`, "the nec annotation is transformed while the axis is not".
It reads the `nec_vals` and `x_e` columns `ggbnec_data()` returns, so it asserts
package output rather than a recomputation.

### A cross-path test that could not fire

"plot and autoplot agree on the predictor axis" ran on `nec4param`, whose
formula transforms nothing. `find_transformations()` returns `character(0)`
there, so both guards apply `xform` whatever either guard is changed to, and
the comparison was vacuous — a fix to `R/plot.R:125` alone left it passing.

It now runs on the transformed-response fixture, where the guard fires, and
compares the ratio each path applies rather than a padded base-graphics axis
limit against a `conditional_effects` grid maximum. That also removes a
`base_max * 0.9` bound, which encoded base graphics' 4% default axis expansion
and would have failed on any change to `xaxs` or `xlim` for reasons unrelated
to `xform`.

### Seven assertions that could not fail

Each was found by a different round, and each is replaced by one whose failure
was then demonstrated by mutating `R/` and observing the failure. Three of the
seven were introduced by earlier rounds of this same review.

| assertion | why it could not fail | mutation that now fails it |
|---|---|---|
| `expect_identical(sum(d$y == 0), 3L)` in the #274 test | restates the fixture set three lines above; `has_family_changed()` takes `d` by value | a return that is no longer a bare logical |
| `all(is.finite(...pr$prior[pr$nlpar == "bot"]))` in the `get_priors` test | `nec3param` has no `bot`, so the selection is `character(0)` and `all(is.finite(numeric(0)))` is `TRUE` | `R/get_priors.R:164` reading the uncorrected response |
| three `expect_silent()` calls on `all_models`, `add_nec`, `add_ec10` | passes identically for every value of each argument | `R/plot.R:236` disabling the `all_models` loop; `:180` drawing `ec10` |
| a count of `abline()` calls | one call is drawn whether the branch is the nec or the ec10 | `R/plot.R:180` drawing `ec10` in the nec branch |
| `expect_identical(pr$nlpar, pr_shifted$nlpar)` | `nlpar` is decided by the model and family, not by the shifted values | — removed |
| `all(vapply(calls, ...))` in the `lxform` pin | `all()` of an empty list is `TRUE`, so it passed if `axis()` were never called | deleting `R/plot.R:166`, which leaves the axis undrawn |

The `get_priors` one is the most consequential of the seven, because
`test-check_data.R` is the file the gating leaves running on CRAN. Measured:
with `R/get_priors.R:164` reading the uncorrected response, the `top` prior
changes from `gamma(2, 0.3077)` to `gamma(2, 0.25)` and all 37 assertions in
the file still passed. It is now asserted by equality against the same data
with the shift applied by hand, which is the only form that discriminates, and
that form fails under the mutation.

`add_nec` and `add_ec10` are now asserted on the values drawn, not the number
of them: the nec branch must draw `nec4param$ne`, the ec10 branch must draw
something else, and the combined branch must draw both in that order.

### A tautology in the #274 test

`expect_identical(sum(d$y == 0), 3L)` followed `d$y[1:3] <- 0` three lines
above. `has_family_changed()` takes `d` by value and returns a logical
(`R/helpers.R:957-958`), so the assertion could not fail for any state of the
package.

Replaced by the two facts that can fail: the message is emitted from this
route, and the whole of what comes back is a length-one logical. The second is
what makes the correction unreachable by the caller, which cannot obtain the
repaired data frame from a logical, so it inverts if #274 is fixed by changing
the return.

### An assertion weakened by its own helper

`pl_x_limits()` wrapped the call in `suppressMessages()`, so the five
`expect_silent()` calls made through it could not fail on a message. They could
still have failed on a warning or on printed output, so this was a partial
weakening rather than a complete one, and it was latent: these methods emit
nothing today. The suppression is removed.

### Duplicated coverage removed

| assertion dropped | where it already exists |
|---|---|
| `plot()` on a plain `bayesnecfit` and `bayesmanecfit` | `test-bayesnec_methods.R:16`, `test-bayesmanec_methods.R:19` |
| the integer predictor guard | `test-fit_bayesnec.R:170` |
| `pop_var_is_transformed()` | `test-fit_bayesnec.R:184` |
| `max(x_e)` unchanged under `xform`, in the annotation test | the sibling test above it in `test-autoplot.R` |

The first three are relative to `aa5ebae2`, which the opening post describes;
the fourth was introduced by `8d8e9a65` and removed by `363441f4`. Each is
recorded in the header of the file that drops it — a table in
`test-check_data.R:13-19`, prose in `test-plot.R:49-53`. What is kept
from the last is the fixture property the pinning tests depend on and which is
asserted nowhere else: `find_transformations()` reports `"y"` for this formula.
`all_models`, `add_nec` and `add_ec10` have no coverage elsewhere and are kept.

### Fixtures

`setup.R:25` already builds `nec4param`. Both plot files rebuilt it on every
call — `pull_out()` measured between 0.91 s and 1.14 s, 16 invocations — and
called `data()` inside a helper. `data()` defaults to `envir = .GlobalEnv`, so
that wrote to the global environment rather than the function frame on every
call, while `setup.R:24` had already loaded the object. Both files now use
`nec4param` and `manec_example` directly.

`transformed_response_fit()`, `transformed_response_manec()` and `gg_x_max()`
are defined in `setup.R`, which is where this repository keeps shared test
fixtures. `gg_x_max()` has to live there rather than in `test-autoplot.R`
because `test-plot.R` needs it for the cross-path comparison and testthat gives
each test file its own environment.

### Line references corrected

The comments cited `R/plot.R:119`, `:179`, `:259` and `R/autoplot.R:306`,
`:350`. Those are the `find_transformations()` calls and, in the first case,
the `if` above the `abline()`. The guard is at `R/plot.R:125` and `:279` and
`R/autoplot.R:308` and `:352`; the `abline()` is at `R/plot.R:180`. A sixth,
`R/autoplot.R:346`, is the `prep_raw_data()` call; the `model.frame()` feeding
the guard is at `:347`.

### Assertion counts per file

| file | `aa5ebae2` | now | change |
|---|---|---|---|
| `test-check_data.R` | 37 | 38 | −1 integer predictor duplicate, +1 #274 replacement, +1 discarded `custom_name` |
| `test-plot.R` | 11 | 41 | −2 duplicate silence block, −3 `expect_silent`, +18 the argument assertions on both classes, +6 annotation pins, +6 the two newly pinned guards, +3 `lxform` on both classes, −1 cross-path, +1 count guard |
| `test-autoplot.R` | 18 | 23 | −2 `pop_var_is_transformed` duplicates, +1 fixture property, +4 annotation pins, −1 duplicate `max(x_e)` |

### An incidental correction to DESCRIPTION

`testthat (>= 3.0.0)` is raised to `3.1.8`. `local_mocked_bindings(.package = )`
was added in 3.1.7, but 3.1.7 rebinds only in the package namespace, and
`abline`, `legend` and `fixef` are imported rather than defined, so they exist
only in the imports environment: measured on the loaded package, all three are
absent from `asNamespace("bayesnec")` and present in its parent. testthat's own
NEWS records the fix under 3.1.8 — "`with_mocked_bindings()` and
`local_mocked_bindings()` now also bind in the imports namespace".

The declared floor was already wrong on `dev`, where `test-inits_functions.R:385`
mocks `fixef` and is not gated by `skip_on_cran()`; this branch adds three more
sites in `test-plot.R`. On a machine with testthat 3.1.7 the declared minimum
would be satisfied and the suite would still fail with `no binding for "fixef"`.
Verified from the 3.1.7 and 3.1.8 sources and NEWS, not by installing 3.1.7.

### Runtime

The three files run in 84 s, against 74 s for the original, both measured with
`test_local()` on R 4.6.1 excluding about 5 s of `load_all()` (88.7 s and
79.1 s inclusive). The figure has risen through the review, from 35 s after
round 1: identifying each annotation rather than counting it, and doing so on
both classes, adds six `ecx()` calls, and `ecx()` on a `bayesmanecfit` draws a
weighted posterior. Single runs on a contended machine; treat as approximate. Note that `test_local()` calls
`local_assume_not_on_cran()`, so it sets `NOT_CRAN` itself and the figure is for
the ungated path; the CRAN path was measured separately with `test_dir()`, as
above.

</details>

## Verification

R 4.6.1. Three files, `test_local()`:

```
test-check_data.R   PASS 38   FAIL 0
test-autoplot.R     PASS 23   FAIL 0
test-plot.R         PASS 41   FAIL 0
```

Full suite, `test_local()`:

```
FAIL 0 | WARN 2 | SKIP 0 | PASS 2426
```

Measured on `27d504de` with `test_local()` under `NOT_CRAN=true`, R 4.6.1. It
agrees with the derivation from the opening post's own numbers,
2390 − 66 + 102 = 2426.

The two warnings are pre-existing and neither is in a file this branch touches:
`test-disp_model.R:412` (a `loo` `p_waic` warning) and `test-get_priors.R:239`.

The Windows job is expected to fail. #275: `R-CMD-check` force-installs
`StanHeaders` over the version `rstan` was matched to, and the failure is
identical on `dev`.

beckyfisher:
The findings with no issue of their own are now #278, which records all seven:
the three from the opening post and the four found during review.

They are grouped in one issue rather than split because they are one kind of
defect — a condition written that cannot fire, or that tests something other
than what it appears to — and #265 was exactly that.

Six of the seven are pinned as current behaviour by the test files this pull
request adds, with the assertion to invert named beside each, so each fix is an
inversion rather than a fresh reproduction. The seventh, the missing `xform`
guard on `ggbnec_data.bayesmanecfit()`, is not pinned here; it was found in the
last review round and neither half of it is asserted.

#278 also records the two items that need a decision rather than a repair: what
should happen to `NA` and `NaN` rows, which are currently dropped silently, and
whether `position_legend` should accept the numeric pair its documentation
names.

PR #279, opened 2026-09-05
Refuse missing values; accept both position_legend forms; drop four dead guards (#278)

### LINKED ISSUE #278: Three argument behaviours that surprise, and four unreachable guards, in check_data(), plot() and ggbnec_data()
## What

Three arguments in `check_data()`, `plot()` and `ggbnec_data()` behave in a way
a user would not expect from the documentation, and four guards are unreachable
dead code. The two groups are kept in one issue because they were found
together while reviewing #276, but they are not the same kind of thing and the
second group is tidy-up.

## Why it matters

**The three behavioural items.** One silently discards rows, so a fit proceeds
on fewer observations than were supplied. One refuses the argument form its own
documentation names. One validates an argument for a `bayesnecfit` and not for
a `bayesmanecfit`, so the same user error gives a clear message or an obscure
one depending on which object was passed.

**The four dead-code items have no behavioural consequence** and are recorded
for one reason only: someone maintaining these functions can otherwise spend
time on a branch that cannot run, or change it believing it has an effect. In
every case the input is handled correctly and the error that does fire is
clear. If that is not worth a change, closing this half is a reasonable
outcome.

## Evidence

Measured on `27d504de`, R 4.6.1, by calling each function directly. Line
numbers are at that commit.

### Behavioural

| # | site | expected | observed |
|---|---|---|---|
| 1 | `check_data.R:121-126` | a non-finite response is refused | `Inf` is refused, but `NA` and `NaN` are dropped by `model.frame()` before the guard sees them. 20 rows supplied, 19 fitted, no message |
| 2 | `plot.R:42-43` documents `position_legend` as "a numeric vector … as per a call to legend" | `c(0.5, 0.5)` accepted | refused at `:89` and `:232`. Separately the `match()` arguments are reversed, so `c("topright", "bogus")` passes the guard and fails inside `legend()` with `'arg' must be of length 1`, naming neither the argument nor the valid values |
| 3 | `autoplot.R:337-364` | `xform` validated as it is at `:293-294` | no guard, so `ggbnec_data(manec_example, xform = "no")` fails with `could not find function "xform"` where the `bayesnecfit` method says `xform must be a function.` |

### Dead code, no behavioural consequence

| # | site | why it cannot run |
|---|---|---|
| 4 | `check_data.R:112-118` | `retrieve_var(error = TRUE)` at `:108` raises first, with `The input variable "x" is not numeric.` |
| 5 | `check_data.R:203-209` | `model.frame()` raises first, with `Group-level variables cannot be numeric.` |
| 6 | `plot.R:152-157`, `:304-309` | written for `lxform` not being a function; `:80-82` and `:222-224` have already stopped on that input |
| 7 | `check_data.R:211` | assigns `check_custom_name(family)` and never reads it; the function is pure. Same at `plot.R:93`, `:263`, `autoplot.R:178` |

For 4 and 5 the message that does fire is shorter than the unreachable one but
says the same thing, so nothing is lost to the user.

<details>

### Where each is pinned

All but item 3 are asserted as current behaviour by #276, with the assertion to
invert named in a comment beside it.

| # | pinned by |
|---|---|
| 1 | `test-check_data.R`, "an NA or NaN row is dropped by model.frame, not caught by the check" |
| 2 | `test-plot.R`, "position_legend refuses the numeric vector its documentation names" |
| 3 | not pinned; neither half is asserted |
| 4 | `test-check_data.R`, "a character predictor errors from retrieve_var, not check_data" |
| 5 | `test-check_data.R`, "a numeric group-level column errors from model.frame, not check_data" |
| 6 | `test-plot.R`, "the lxform branch in plot() cannot be reached, on either class" |
| 7 | `test-check_data.R`, "the check_custom_name result is discarded" |

### Proposed resolution

**Item 1 needs a decision before anything is implemented, and is the only one
of the seven with a scientific consequence.** `model.frame()` drops incomplete
cases by its own default, so the finiteness guard never sees them. Either
dropping is intended, in which case it should be announced the way the boundary
corrections are, or it is not, in which case the guard should run before
`model.frame()`. A fit silently using fewer rows than the user supplied is the
part of this issue worth acting on first.

**Item 2** is two changes. Correct the `match()` direction at both call sites —
`!all(position_legend %in% legend_positions)` is the intended test. Then decide
between the guard and the documentation, since `legend()` does accept a numeric
`x, y` pair: accepting both forms means admitting a length-2 numeric while
still rejecting an unknown keyword. That half is a choice, not a repair.

**Item 3** is one line: add the guard to `ggbnec_data.bayesmanecfit()`, matching
`autoplot.R:293-294`.

**Items 4 to 7** are deletions. For item 6, note that removing the dead branch
does not change the axis: the live `else` labels ticks with
`signif(lxform(x_ticks), 2)`, which with the default `lxform = identity` gives
seven data-derived positions rounded to two significant figures — measured on
`manec_example`, 0.032, 0.56, 1.1, 1.6, 2.2, 2.7, 3.2 rather than R's default
pretty values. Whether that labelling is wanted is a separate cosmetic
question, not part of this issue.

### Not to be done here

Do not fold #268 into this. It is a different defect — a scale mismatch within
one figure — with twelve sites of its own, all pinned by #276.

### Definition of done

Each item fixed or explicitly declined with the reason recorded here. For each
fix, the pinning assertion in #276's test files is inverted in the same commit,
so the test states the new behaviour rather than the old.

### What to stop and ask about

Item 1, and the documentation-versus-guard half of item 2. Both are choices
about intended behaviour. The dead-code group needs no decision beyond whether
it is worth the diff.

</details>

Found while reviewing #276. Verified on `27d504de` rather than taken from the
review text.

beckyfisher:
Closes #278.

## What

Three arguments in `check_data()`, `plot()` and `ggbnec_data()` behaved in a way
the documentation did not describe, and four guards were unreachable. All seven
items are addressed, following the two decisions RF made on the choices #278
flagged.

## Why it matters

`bnec()` no longer fits on fewer rows than were supplied. `stats::model.frame()`
removes incomplete cases before `check_data()` is given the model frame, so `Inf`
was refused while `NA` and `NaN` were removed silently — twenty rows supplied,
nineteen fitted, nothing said. That is now an error naming the count and the rows.

`position_legend` accepts the numeric form its own documentation has always named,
and refuses a value `legend()` cannot use with a message that names the argument.
`ggbnec_data()` reports the same `xform` error for both fit classes.

**This narrows a behaviour introduced earlier in 2.2.0.** The response write-back
was changed to match rows by name so that a data frame with an `NA` in a population
variable no longer failed with "replacement has *n* rows, data has *m*". That input
is now refused before the write-back is reached. The row matching is still required
and still asserted, by the case that gives the data frame row names other than
`1:n`. The NEWS entry was rewritten to say so.

## Evidence

Item 1 is the only one with a scientific consequence, and #278 records the
measurement: twenty rows supplied, nineteen fitted, no message. Items 2, 3, 6 and 7
were measured by calling each function directly at `27d504de`; items 4 and 5 by
observing which error fires. All but item 3 were pinned by #276, and every pinning
assertion is inverted in this branch, so each test now states the new behaviour.

The full `testthat` suite was run on R 4.6.1 with `NOT_CRAN=true`, so nothing
was skipped: 57 files, no failures and no errors. The two warnings reported are
both present on `dev`. The eight files most exposed to the change —
`test-check_data.R`, `test-plot.R`, `test-autoplot.R`, `test-fit_bayesnec.R`,
`test-cens.R`, `test-get_priors.R`, `test-hurdle_family.R` and
`test-check_normalisation.R` — account for 144 of those tests.

<details>

### Item 1 — the finiteness guard and missing values

Decided in the request: the guard refuses `NA` and `NaN` as it refuses `Inf`.

Three implementations were considered.

1. `na.action = na.pass` in `model.frame.bayesnecformula()`, so the rows survive
   and the existing guard names the column. **Rejected.** `check_data()` is not the
   first thing to see `bdat` in `bnec()` — `retrieve_valid_family()` calls
   `set_distribution()` on the response, and `check_normalisation()` and
   `check_inline_boundary()` both run first — so the error would come from one of
   those, about the wrong thing. It would also change what the other eight callers
   of the exported `model.frame` method receive.
2. Refuse inside `model.frame.bayesnecformula()`. **Rejected** for the same reason —
   eight callers, most of them on data a fit has already validated — and because
   the check belongs in the documented data check.
3. Refuse in `check_data()`, reading the rows off the `na.action` attribute
   `na.omit()` leaves on the model frame. **Adopted.** Unchanged for every other
   caller, and it covers all three routes into `check_data()` — `bnec()`,
   `fit_bayesnec()` and `get_priors()`.

The message reports the count and the row numbers. The column cannot be named from
the attribute once the rows are gone, so it is not.

The guard itself changed from `!is.finite(mean(x))` to `!all(is.finite(x))`. The
two agree on every input the first can see; the elementwise form also refuses a
missing value arriving through `options(na.action = "na.pass")`, where the attribute
is absent, and names which column holds it.

### Item 2 — `position_legend`

Two defects in one guard, at two call sites with two different messages. The
`match()` arguments were reversed, so `c("topright", "bogus")` passed and failed
inside `legend()` with `'arg' must be of length 1`. And the numeric form the
documentation named was refused.

Replaced with one `check_position_legend()`, accepting either a single keyword or a
length-two numeric and returning `list(x =, y =)`. The list is necessary: `legend()`
reads a length-two numeric `x` with `y = NULL` as the two corners of a rectangle,
not as a point, so passing the pair through as `x` would have placed the legend
somewhere the user did not ask for while appearing to work. The pair is split across
`x` and `y` at all six call sites.

Partial keyword matching was never available — the old guard required an exact
match before `legend()` saw the value — so `%in%` removes nothing that was available.

### Item 3 — `xform` for a model set

One guard added to `ggbnec_data.bayesmanecfit()`, matching
`ggbnec_data.bayesnecfit()`.

### Items 4 to 7 — deletions

Removed: the predictor and group-level type checks in `check_data()`, which
`retrieve_var()` and `stats::model.frame()` refuse first with the same substance;
the `lxform` branch in each `plot()` method; and the four discarded
`check_custom_name()` assignments #278 names.

The `check_custom_name` test now mocks the function to raise, which is the only way
the absence of a discarded call can be observed — asserting on the return value
cannot distinguish a pure call from no call, which is why the assignment survived.

### Found but not changed

Four further discarded `check_custom_name()` assignments exist outside the four
#278 names: `R/fit_bayesnec.R:38` (overwritten at `:57`), `R/amend.R:251`,
`R/define_prior.R:128` and `R/helpers.R:442`. Same defect, no behavioural
consequence, and outside what was agreed for this branch. A follow-up issue is
warranted if that group is considered worth changing.

### Files

| file | change |
|---|---|
| `R/check_data.R` | missing-value guard; elementwise finiteness; two dead branches and one discarded call removed |
| `R/plot.R` | `check_position_legend()`; both guards replaced; both `lxform` branches and both discarded calls removed; `position_legend` documentation |
| `R/autoplot.R` | `xform` guard on the model-set method; one discarded call removed |
| `R/bnec.R` | `@param data` states the complete-case requirement |
| `NEWS.md` | four entries; the write-back entry rewritten |
| `tests/testthat/` | every #276 pin inverted; three new cases |

</details>

https://claude.ai/code/session_01LqT4YPVMUGKu3m5QQZFXfs

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -169,10 +169,11 @@
   between models matters, refit the whole set.
 
 - The response write-back now matches rows by name rather than assigning
-  wholesale, so a data frame carrying an `NA` in any population variable no
-  longer fails with "replacement has *n* rows, data has *m*". The model frame
-  drops incomplete cases, so it is shorter than the data frame `brm()` is given
-  wherever one is present.
+  wholesale, so it is correct for a data frame whose row names are not `1:n`
+  and cannot fail with "replacement has *n* rows, data has *m*". A missing
+  value in a population variable, which is the other way the model frame comes
+  back shorter than the data frame `brm()` is given, is refused outright as of
+  the entry below rather than accommodated by the write-back.
 
 - `trials()` carrying arithmetic — `y | trials(n * 2) ~ crf(x, ...)` — no longer
   fits every observation against the wrong number of trials. The trials column
@@ -181,6 +182,42 @@
   then evaluated `trials(n * 2)` against *that*: a recorded 10 became 40. The
   write-back is removed; `check_data()` never corrects the trials variable, so
   it had nothing to carry.
+
+- `check_data()` refuses a row with a missing value in a population variable
+  rather than letting `stats::model.frame()` remove it. The model frame drops
+  incomplete cases before the finiteness guard is reached, so `Inf` was refused
+  while `NA` and `NaN` were removed silently: twenty rows supplied, nineteen
+  fitted, nothing said. The error reports how many rows held a missing value
+  and which they were. Where those rows are genuinely to be discarded, apply
+  `na.omit()` to the data frame before calling `bnec()`, so that the sample the
+  estimates are derived from is the user's decision and is visible in the
+  script. The guard also reads `is.finite()` elementwise rather than on the
+  mean, so a missing value that reaches it through
+  `options(na.action = "na.pass")` is refused as well, naming the column
+  (#278).
+
+- `position_legend` in `plot()` accepts the numeric form its documentation has
+  always named. Either of the two forms `graphics::legend()` takes is now
+  valid: one of the eight position keywords, or a numeric vector of length two
+  giving the x and y coordinates. The guard it replaces also tested
+  `match(legend_positions, position_legend)`, matching the valid keywords into
+  the user's value rather than the reverse, so `c("topright", "bogus")` passed
+  it and failed inside `legend()` with `'arg' must be of length 1`, which names
+  neither the argument nor the valid values. Both `plot()` methods now use one
+  guard with one message (#278).
+
+- `ggbnec_data()` validates `xform` for a `bayesmanecfit` as it already did for
+  a `bayesnecfit`. `ggbnec_data(x, xform = "no")` failed with
+  `could not find function "xform"` for a model set where a single fit reported
+  `xform must be a function.` (#278).
+
+- Four unreachable guards removed, with no change in behaviour. The predictor
+  and group-level type checks in `check_data()` composed a longer message than
+  `retrieve_var()` and `stats::model.frame()`, which refuse the same input
+  first and say the same thing; the `lxform` branch in each `plot()` method is
+  preceded by a guard that has already stopped on that input. Four
+  `check_custom_name()` calls whose result nothing read are removed alongside
+  them (#278).
 
 # bayesnec 2.1.4
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -171,9 +171,9 @@
 - The response write-back now matches rows by name rather than assigning
   wholesale, so it is correct for a data frame whose row names are not `1:n`
   and cannot fail with "replacement has *n* rows, data has *m*". A missing
-  value in a population variable, which is the other way the model frame comes
-  back shorter than the data frame `brm()` is given, is refused outright as of
-  the entry below rather than accommodated by the write-back.
+  value in any variable the formula names, which is the other way the model
+  frame comes back shorter than the data frame `brm()` is given, is refused
+  outright as of the entry below rather than accommodated by the write-back.
 
 - `trials()` carrying arithmetic — `y | trials(n * 2) ~ crf(x, ...)` — no longer
   fits every observation against the wrong number of trials. The trials column
@@ -183,23 +183,30 @@
   write-back is removed; `check_data()` never corrects the trials variable, so
   it had nothing to carry.
 
-- `check_data()` refuses a row with a missing value in a population variable
-  rather than letting `stats::model.frame()` remove it. The model frame drops
-  incomplete cases before the finiteness guard is reached, so `Inf` was refused
-  while `NA` and `NaN` were removed silently: twenty rows supplied, nineteen
-  fitted, nothing said. The error reports how many rows held a missing value
-  and which they were. Where those rows are genuinely to be discarded, apply
-  `na.omit()` to the data frame before calling `bnec()`, so that the sample the
-  estimates are derived from is the user's decision and is visible in the
-  script. The guard also reads `is.finite()` elementwise rather than on the
-  mean, so a missing value that reaches it through
-  `options(na.action = "na.pass")` is refused as well, naming the column
-  (#278).
+- A row with a missing value in any variable the formula names is refused
+  rather than left for `stats::model.frame()` to remove. The model frame drops
+  incomplete cases before the finiteness guard in `check_data()` is reached, so
+  `Inf` was refused while `NA` and `NaN` were removed silently: twenty rows
+  supplied, nineteen fitted, nothing said. The error reports how many rows held
+  a missing value and which they were, by row name. Where those rows are
+  genuinely to be discarded, apply `na.omit()` to the data frame before calling
+  `bnec()`, so that the sample the estimates are derived from is the user's
+  decision and is visible in the script.
+
+  The check runs in `bnec()` before any model is fitted, and in `bnec_group()`
+  before the first level is, so a model set states the refusal once and a
+  grouped fit states it before it samples anything. It is retained in
+  `check_data()` for the two routes that do not come through `bnec()`:
+  `get_priors()` and `amend()`. The finiteness guard also reads `is.finite()`
+  elementwise rather than on the mean, so a missing value that reaches it
+  through `options(na.action = "na.pass")` is refused as well, naming the
+  column (#278).
 
 - `position_legend` in `plot()` accepts the numeric form its documentation has
-  always named. Either of the two forms `graphics::legend()` takes is now
-  valid: one of the eight position keywords, or a numeric vector of length two
-  giving the x and y coordinates. The guard it replaces also tested
+  always named. Both the keyword form and the coordinate form
+  `graphics::legend()` takes are now valid: one of the eight position keywords
+  the previous guard listed, or a numeric vector of length two giving the x and
+  y coordinates. The guard it replaces also tested
   `match(legend_positions, position_legend)`, matching the valid keywords into
   the user's value rather than the reverse, so `c("topright", "bogus")` passed
   it and failed inside `legend()` with `'arg' must be of length 1`, which names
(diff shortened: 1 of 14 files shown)


beckyfisher:
Five review cycles were run on this branch. Four changes follow, and the PR body
above predates them.

## What

The missing-value refusal is now raised before anything is fitted, not from
`check_data()` alone. The guard is `check_complete_cases()` and it is called
from `bnec()`, immediately after the model frame is built; from `bnec_group()`,
before the level loop; and from `check_data()`, for the two routes that do not
come through `bnec()`, which are `get_priors()` and `amend()`.

The rows are reported by row name rather than by position.

`?bnec`'s Details section on missing values is rewritten.

Three documented claims are narrowed to what the code does.

## Why it matters

`check_data()` runs once per model inside `fit_bayesnec()`, and `bnec()` wraps
that call in `try()` whenever `length(model) > 1`. A refusal raised from there
was printed once per model, every model was recorded as failed, and the call
ended on "None of the models fit successfully, try using bnec with a single
model ... or check ?show_params", which names neither the missing values nor the
remedy. The default `model` argument is a set, so that was the common path.

`bnec_group()` fits each level in turn, so a missing value in level *k* was
reached only when level *k* was fitted. Before this branch the row was dropped
silently and the call completed; without a check before the loop the branch
would have replaced that with a failure after the earlier levels had sampled.

`?bnec` still carried a Details section headed "NAs are thrown away", stating
that brms excludes NAs prior to fitting and that "we advise that you exclude any
NAs in your data prior to fitting". Neither is true of `bnec()` as of this
branch, and it is the section a user reads about NA handling.

## Evidence

Measured on R 4.6.1 with `NOT_CRAN=true`.

`bnec(y ~ crf(x, model = c("nec4param", "ecx4param")))` on 20 rows with one `NA`
printed the refusal twice and then the all-models-failed message. It now raises
once, from `bnec()`.

`bnec_group()` on 24 rows in two levels of twelve, with one `NA` in level "b":
level "a" compiled and sampled to completion before level "b" raised. It now
raises before the first level is announced.

The same `bnec_group()` run reported the missing row as row 8, its position
within the level subset, for a row the user recorded at 20. `attr(data,
"na.action")` names each element with the row name and holds the position as its
value; the names are read instead. On a frame whose rows are named `s101`–`s120`
the message named `s103` rather than 3.

`bnec_hurdle()` was checked and needs no change: `R/bnec_hurdle.R:156` refuses an
`NA` response with a message specific to a survival endpoint before anything else
runs.

The nine test files most exposed to the change pass with no failures and no
errors: `test-check_data.R` 42 assertions, `test-fit_bayesnec.R` 36,
`test-plot.R` 49, `test-autoplot.R` 25, `test-bnec.R` 8, `test-bnec_group.R` 69,
`test-get_priors.R` 61, `test-bnec_hurdle.R` 56 and `test-amend.R` 21. The full
suite was run on the same code with `NOT_CRAN=true`, so nothing was skipped:
2,443 assertions, no failures and no errors. The two warnings reported, in
`test-disp_model.R` and `test-get_priors.R`, are both present on `dev`.

<details>

### Placement inside `bnec()`

Immediately after `model.frame()` rather than beside `check_normalisation()`, so
that `retrieve_valid_family()` does not choose a family from a sample smaller
than the one supplied. Asserted by the absence of the family-selection message,
which is the only observable thing that runs between the two candidate
positions on that input.

The reasoning is the one `R/bnec.R:588-605` already records for
`check_normalisation()` and `check_inline_boundary()`, and it applies more
strongly here because the default `model` argument is a set.

`fit_bayesnec()` is not exported, so the "direct `fit_bayesnec()` call" named in
the first draft of the comment and the NEWS entry is not a user-reachable route.
The comment and the entry name `get_priors()` and `amend()` instead.

### The three narrowed claims

- The `Inf` guard covers the predictor and the response, not every variable the
  formula names. `@param data` and the Details section say so.
- `position_legend` was documented as accepting "either of the two forms
  `legend()` accepts". The keyword set is `legend()`'s nine less `"center"`.
  `?plot` states the exclusion and a test pins the refusal on both fit classes,
  so widening the set later is a decision rather than an oversight.
- NEWS said "a missing value in a population variable" in two entries; any
  variable the formula names is covered, an `ogl()` grouping variable included.
  The removal entry attributed the group-level refusal to `stats::model.frame()`;
  it is `check_formula()`, at `R/bayesnecformula.R:392`, `:429` and `:458`.

### Tests added

Eight assertions across four files, for behaviour nothing else reached: the
model-set route through `bnec()`; the absence of the family-selection message,
which is what distinguishes the two candidate placements; `bnec_group()`
refusing before the first level is announced; the row-name reporting; and
`"center"` refused on both classes.

</details>

Issue #268, opened 2026-09-02
plot() and autoplot() skip xform on the predictor axis when the response is transformed


beckyfisher:
## What

`plot()` and `autoplot()` decide whether to apply `xform` to the predictor axis
with the same all-or-nothing guard that #258 removed from `fit_bayesnec()`. A
transformation on the **response** therefore suppresses `xform` on the
**predictor** axis, silently, and the axis is drawn on the fitted scale while
the caller asked for it on the recorded one.

## Why it matters

`xform` exists so that a fit on a transformed predictor can be plotted on the
scale the concentrations were recorded on. Where the response also carries a
transformation — `log(y) ~ crf(log(x), ...)` — the argument is accepted and
ignored, so the axis is labelled in log units while the reader has asked for
concentrations. Nothing reports it. The estimates `nec()` and `ecx()` return are
unaffected; this is the plot alone.

## Evidence

`dev` at bc5f7d00, R 4.6.1. The guard is the length of
`find_transformations()`, which answers for the formula as a whole:

```r
d <- data.frame(x = rep(c(1, 10, 100), each = 4), y = rep(c(9, 5, 1), each = 4))
bdat <- model.frame(bnf(log(y) ~ crf(x, model = "nec3param")), data = d)
bayesnec:::find_transformations(bdat)
#> [1] "y"
length(bayesnec:::find_transformations(bdat)) == 0
#> [1] FALSE
```

The predictor is untransformed in that formula, and the per-variable predicate
added for #258 says so:

```r
bayesnec:::pop_var_is_transformed(bdat, "x_var")
#> [1] FALSE
bayesnec:::pop_var_is_transformed(bdat, "y_var")
#> [1] TRUE
```

Four call sites read the coarse answer and skip `xform` on it:
`R/plot.R:119`, `R/plot.R:259`, `R/autoplot.R:306`, `R/autoplot.R:350`. Each is
of the form

```r
trans_vars <- find_transformations(bdat)
# if no transformations are applied via formula (including on trials),
# use xform on axis
if (length(trans_vars) == 0) {
```

This is read from the source and from the predicate, not from a rendered plot:
producing one needs a fitted object whose formula transforms the response, which
the packaged example fits do not have.

Found while fixing #258, which replaced the same guard in `fit_bayesnec()` but
deliberately did not touch the plotting paths.

<details>
<summary>Implementation detail</summary>

`pop_var_is_transformed(data, var)` (`R/helpers.R`, added for #258) answers the
[...]

PR #281, opened 2026-09-06
Measure every ECx from the control, and fix the crf() back-transform (#195, #196, #39, #206, #268, #160)

### LINKED ISSUE #195: ecx()/ecnsec()/nsec(): inert hormesis_def, divergent type implementations, and a wrong documented default
## Summary

This replaces the original text of this issue, which spent most of its length on
the `x_range` sensitivity of `type = "relative"`. That sensitivity is the
intended meaning of a *relative* estimate — the user chooses `x_range` by expert
judgement — and it is not a defect. It has been removed.

What is left is a short list of things that are actually wrong in the code: one
argument that does nothing, one documented default that is not the effective
default, and three places where the same documented sentence is implemented
three different ways. Plus one genuine gap, flagged as an enhancement rather
than a bug.

The factual basis, established from source, is that every `type` in `ecx()` is
referenced against the **posterior predicted curve over the x grid**, per draw —
not against fitted parameters and not against observed data:

| `type` | `ecx_y` | line |
|---|---|---|
| `"relative"` | `max(y_d) - (max(y_d) - min(y_d)) * ecx_val/100` | `R/ecx.R:318-319` |
| `"absolute"` | `max(y_d) * (1 - ecx_val/100)` | `R/ecx.R:330-331` |
| `"direct"` | `ecx_val` | `R/ecx.R:342` |

where `y_d` is one row of `posterior_epred()` — a single draw's predicted curve
over the grid (`R/ecx.R:195`; orientation confirmed at `R/posterior_epred.R:27`,
*"cols are x, rows are iterations"*). This matters below only because it is what
makes items 1, 2 and 5 demonstrable.

---

## 1. `hormesis_def` has no effect in `ecx()` — its only consumer is commented out

**Bug.** `?ecx` documents:

> For `hormesis_def`, if "max", then ECx values are calculated as a decline from
> the maximum estimates (i.e. the peak at NEC); if "control", then ECx values
> are calculated relative to the control, which is assumed to be the lowest
> observed concentration.

`hormesis_def` is validated (`R/ecx.R:131-134`) and passed down to member fits
(`R/ecx.R:290`), but the block that would act on it is commented out:

```r
# R/ecx.R:189-193
  # if (grepl("horme", object$model)) {
  #   n <- seq_len(nrow(p_samples))
  #   p_samples <- do_wrapper(n, modify_posterior, object, x_vec,
  #                           p_samples, hormesis_def, fct = "rbind")
  # }
```

`modify_posterior()` (`R/helpers.R:228-239`) is the only consumer, and with the
call disabled it is now dead code for `ecx()`. It would have set the rising limb
to `NA` so that `range(y)` and `max(y)` excluded the hormetic increase:

```r
# R/helpers.R:228-239
modify_posterior <- function(n, object, x_vec, p_samples, hormesis_def) {
  posterior_sample <- p_samples[n, ]
  if (hormesis_def == "max") {
    target <- x_vec[which.max(posterior_sample)]
    change <- x_vec < target
  } else if (hormesis_def == "control") {
    target <- posterior_sample[1]
    change <- posterior_sample >= target
  }
  posterior_sample[change] <- NA
  posterior_sample
}
```

Consequences:

- **`hormesis_def` is silently ignored.** Both settings return bit-identical
  posteriors — see the demonstration below.
- **`ecx()` unconditionally behaves as `hormesis_def = "max"`**, because
  `ecx_x_relative()` and `ecx_x_absolute()` both take `max(y_d)` over the whole
  grid, which for a hormesis curve *is* the peak. The documented default is
  `"control"`.
- This also makes the `(top)` gloss in the `?ecx` Details wrong for hormesis
  models specifically: the upper anchor is the peak, which exceeds `top`.
- The `if (length(which(!is.na(y))) == 0)` guards and the `na.rm = TRUE`
  arguments in all three `ecx_x_*()` helpers are now vestigial — nothing on the
  live path can introduce an `NA`.

`nsec()` still honours `hormesis_def = "max"` (`R/nsec.R:166-168`), and so does
`ecnsec()` (`R/ecnsec.R:117-121`), so `ecx()` is the odd one out.

**Demonstrated** on a `nechorme` fit (`dev/ecx_bug_verification.R`), comparing
the full posteriors rather than just the medians:

```
ecx EC10 relative  control median 1.54505   max median 1.54505   identical: TRUE
ecx EC10 absolute  control median 1.54825   max median 1.54825   identical: TRUE
ecx EC50 relative  control median 1.86100   max median 1.86100   identical: TRUE
ecx EC50 absolute  control median 1.88972   max median 1.88972   identical: TRUE
```

`identical()` on the two posterior vectors is `TRUE` in every case — not merely
close. The argument is read, validated, passed down, and never used.

For contrast, on the **same fit** `ecnsec()` does respond to it, and strongly:

```
ecnsec relative  control 1.3439   max 1.9985   differ: TRUE
ecnsec absolute  control 1.2744   max 1.8969   differ: TRUE
```

So `hormesis_def` moves `ecnsec()` by roughly 49% and moves `ecx()` by exactly
zero.

**Options:** re-enable the block and check `modify_posterior()` against the
current `p_samples` layout; or remove `hormesis_def` from `ecx()` and document
that hormesis models are always anchored at the peak. Either is defensible;
leaving a documented argument that does nothing is not.

---

## 2. `ecx()` and `ecnsec()` implement the same documented sentence differently

**Inconsistency.** `R/ecnsec.R:33-38` carries the `type` description *verbatim*
from `R/ecx.R:34-39` — the same sentence, rendered identically at
[`?ecx`](https://open-aims.github.io/bayesnec/reference/ecx.html) and `?ecnsec`:

> For `type` "relative" is calculated as the percentage decrease from the
> maximum predicted value of the response (top) to the minimum predicted value
> of the response. Type "absolute" (the default) is calculated as the percentage
> decrease from the maximum value of the response (top) to 0.

The two functions implement it with different quantities:

| anchor | `ecx()` | `ecnsec()` |
|---|---|---|
| upper | `max(y_d)` over the grid (`R/ecx.R:318`, `:330`) | `p_samples[, 1]`, the **first grid point** (`R/ecnsec.R:120`) |
| lower, `"relative"` | `min(y_d)` over the grid (`R/ecx.R:318`) | `p_samples[, ncol]`, the **last grid point** (`R/ecnsec.R:124`) |
| lower, `"absolute"` | `0` (`R/ecx.R:330`) | `0` (`R/ecnsec.R:126`) |

```r
# R/ecnsec.R:117-127
  if (hormesis_def == "max") {
    control_posterior <- apply(p_samples, 1, max)
  } else {
    control_posterior <- p_samples[, 1]
  }

  if(type=="relative"){
    min_posterior <- p_samples[, ncol(p_samples)]
  } else {
    min_posterior <- 0
  }
```

For a monotonically decreasing curve the grid extremes and the grid endpoints
are the same points and the two agree. For any non-monotone curve — every
`horme` model — they are different points and the two disagree. Note also that
`ecnsec()`'s `hormesis_def = "max"` branch reproduces exactly what `ecx()` does
unconditionally, which is the same root cause as item 1.

**Demonstrated** on the same `nechorme` fit, reporting the four anchor
quantities directly (medians across draws):

```
ecx()    upper anchor = max over grid   : 0.9765
ecnsec() upper anchor = first grid point: 0.97035    (hormesis_def = 'control')
ecx()    lower anchor = min over grid   : 0.04933
ecnsec() lower anchor = last grid point : 0.04933    (type = 'relative')
```

The lower anchors coincide, because this curve is monotonically decreasing by
the end of the grid. The upper anchors do not: `ecx()` reads the hormetic peak,
`ecnsec()` reads the control.

The raw gap is small here (0.9765 vs 0.97035) because the fitted hormetic rise
was modest — I am not claiming otherwise. But it enters the percent-effect
calculation as a difference of differences, so it propagates: it is exactly the
gap that produces the 1.3439-vs-1.9985 split in item 1. A small difference in
the anchor is not a small difference in the answer.

---

## 3. Three different formulae for the percent effect at the NSEC

**Inconsistency.** The same conceptual quantity is computed three ways:

```r
# R/nsec.R:156-159  -- attached to every nsec() result as attr "ecnsec_relativeP"
  ecnsecP <- apply(p_samples, MARGIN = 1, FUN = function(r){
    #(max(r) - diff(range(r)))/reference * 100
    (1-diff(c(min(r), reference))/(diff(range(r)))) * 100
  })
```

```r
# R/ecnsec.R:129-131
  dif_valsC <- control_posterior-min_posterior
  ecnsecP <-  (control_posterior -  reference)/  dif_valsC * 100
```

The `nsec()` attribute uses `min(r)` and `diff(range(r))` — grid extremes, the
`ecx()` convention — while `ecnsec()` uses grid endpoints. On the same
`nec3param` fit:

```
nsec(fit3) median                            : 1.54389

nsec() attr "ecnsec_relativeP"               : 1.5023  (0.2850, 2.6251)
ecnsec(type = "relative", resolution = 1000) : 1.5743  (0.3589, 2.6970)
ecnsec(type = "absolute", resolution = 1000) : 1.4863  (0.3362, 2.5441)
```

Three answers to "what percent effect does this NSEC correspond to". The spread
is small here (1.49–1.57%) because the fit is monotone and its curve nearly
reaches zero, so the three references nearly coincide — I am not claiming a
large numerical error on this dataset. The problem is structural:

- **The `nsec()` attribute is hardcoded to the relative form.** `nsec()` has no
  `type` argument, so the attribute is always relative, while `ecnsec()` defaults
  to `type = "absolute"`. A user reading `attr(nsec(fit), "ecnsec_relativeP")`
  and a user calling `ecnsec(fit, nsec = ...)` with default arguments are
  reading two different estimands, and nothing says so. (The attribute name does
  contain `relative`, which is the only clue.)
- The same block is duplicated four times — `R/nsec.R:156-159`, `:351-354`,
  `:378-381` (`nsec.brmsfit`) and `:508-510`, `:538-540` (`nsec.drc`) — so any
  correction has to be made in five places.

One further point I could not resolve and am flagging rather than guessing:

```r
# R/ecnsec.R:112-115
  pred_val_nsec <- posterior_epred(object,
                                   newdata = newdat_nsec$newdata,
                                   re_formula = NA)
  reference <- median(pred_val_nsec[, 2])
```

`reference` is collapsed to a scalar median across draws, while
`control_posterior` on the next lines is per draw. Every other reference
quantity in `ecx()`/`nsec()` is either per draw throughout or a scalar
throughout. This may well be deliberate — `nsec` arrives as a single supplied
number, so there is an argument for not giving it a posterior — but the effect
is that uncertainty in the response at the NSEC is discarded while uncertainty
in the control is retained. Worth a decision either way.

---

## 4. `ecnsec()` documents `resolution = 1000` but the effective default is 10

**Bug, documentation/API.** The generic declares `resolution = 1000`
(`R/ecnsec.R:56`) and that is what `man/ecnsec.Rd:11` shows in `\usage`. Both
methods declare `resolution = 10`:

- `R/ecnsec.R:79` — `ecnsec.bnecfit`
- `R/bayesnechurdlefit-methods.R:910` — `ecnsec.bayesnechurdlefit`

Dispatch uses the method's default, so a user who does not pass `resolution`
gets 10, not the documented 1000:

```
generic default          : 1000
bnecfit method default   : 10
hurdle method default    : 10
man/ecnsec.Rd \usage says: 1000
```

**Scope, stated honestly:** on a monotone fit this has no numerical effect,
because `ecnsec()` reads `p_samples[, 1]` and `p_samples[, ncol]`, which are the
`x_range` endpoints at any resolution:

```
ecnsec(fit3, nsec = 1.54389)                    : 1.4863  (0.3362, 2.5441)
ecnsec(fit3, nsec = 1.54389, resolution = 1000) : 1.4863  (0.3362, 2.5441)
```

It does matter for `hormesis_def = "max"`, which takes `apply(p_samples, 1, max)`
over the grid — a 10-point grid can miss the peak. Either way the documented
default and the effective default should agree.

---

## 5. The vignette and the JSS article describe `"relative"` incorrectly

**Documentation/code divergence, the #170 pattern.** `?ecx` itself is correct on
`"relative"`. The vignette is not. `vignettes/example2b.Rmd:102`, repeated
verbatim at `article/article.Rmd:798`:

> For `type = "relative"` *EC~x~* is calculated as the percentage decrease from
> the maximum predicted value of the response ($\tau = \text{top}$) to the
> minimum predicted value of the response (**i.e., *relative* to the observed
> data**).

No observed `y` value enters `ecx()` anywhere — the minimum is a predicted-curve
minimum over `x_range`. This sentence caused a real disagreement between two
people about what `"relative"` means, which is what prompted this investigation.
The `$\tau = \text{top}$` gloss in the same sentence is wrong in the other
direction, suggesting a fitted parameter.

Two smaller ones in the same family:

- **`?ecx` on `"absolute"`** (`R/ecx.R:36-38`, `man/ecx.Rd:69-74`, inherited by
  `man/ecnsec.Rd:67-74` and `man/compare_posterior.Rd:64-67`) says "the maximum
  value of the response (top)", dropping the word *predicted* that the relative
  sentence directly above it uses twice. Read literally that is the observed
  maximum. The code uses `max()` of the predicted curve.
- **`vignettes/example1.Rmd:489`** — "ECx is invariant to the choice of constant
  divisor, because it is a relative decline from the fitted `top`". The
  conclusion is right; the reason is the same `top`-for-`max(pred)` substitution.

In every case the code is correct and the prose is what needs changing.

---

## 6. Gap (not a bug): no `type` measures the decline from control to `bot`

Raised separately, and recorded here because it is the one substantive
*functional* observation left after removing the non-issues.

Given the Details description, the available estimands are top→0
(`"absolute"`), max→min over the grid (`"relative"`), and a raw y value
(`"direct"`). None of them is *the decline from the control to the fitted
lower asymptote* — the quantity of interest when a four-parameter model has a
genuine floor and the maximum achievable effect is less than 100%.

`"relative"` approximates it, and converges on it as `x_range` widens far enough
for the curve to flatten. Two data points from the fits already run:

| fit | fitted `bot` | grid minimum at default `x_range` |
|---|---|---|
| `nec4param` on `nec_data` (curve still falling at max x) | 0.011881 | 0.057595 |
| `nec4param` on simulated data with a floor at 0.40 (curve flat well before max x) | 0.40469 | 0.40523 |

So the approximation is good when the tested range covers the plateau and poor
when it does not, exactly as expected — and the user has no indication which
case they are in.

A `type` computed from the fitted `top` and `bot` directly would give the same
estimand without depending on `x_range`, and would be well defined only for
models that have a `bot` (an error or fallback would be needed for the
three-parameter forms). Whether that is worth adding is a design question, not a
defect. Noted here so it is not lost.

---

## Explicitly *not* part of this issue

For the record, since the first version of this issue claimed otherwise:

- **`"relative"` depending on `x_range` is intended behaviour**, not a bug. It
  follows directly from the documented definition, and `x_range` is the user's
  expert-judgement lever.
- **`"relative"` returning an estimate for an effect size beyond the maximum
  achievable effect** is a consequence of that same intended definition, not a
  coding error. It is a reporting-transparency matter, and item 6 is the
  constructive version of it.
- **Model averaging computing the reference range per member** is a real
  mechanism (`R/ecx.R:286-297`) but appears self-limiting: where members' lower
  limits diverge enough to matter, the weighting collapses onto one of them
  (weights of `5.3e-12` vs `1.0` in the case tested). No action proposed.

## Related

- **#170** — same documentation/code divergence pattern; item 5 here is another
  instance, and as there, the code is the correct side.
- **#39** and **#166** — grid-search precision and boundary behaviour. Different
  mechanism from anything above: those concern how precisely a given target is
  located on the grid, not how the target is defined or which argument is read.
  Item 4's resolution mismatch is adjacent to #39's territory.
- Daniels, Hommen & Ratte (2026), "Four-parameter nonlinear regression and
  maximum achievable effect in ecotoxicology", *Integrated Environmental
  Assessment and Management* **22**:734–740,
  [doi:10.1093/inteam/vjaf085](https://doi.org/10.1093/inteam/vjaf085) — the
  external motivation for item 6, and for stating the relative/absolute
  distinction precisely in the documentation (item 5).

Verification code: `dev/ecx_bug_verification.R`, output in
`dev/ecx_bug_verification_results.txt`.

beckyfisher:
## What

Every ECx, NSEC and ECNSEC is now measured from the **control** — the predicted
mean at the lowest concentration in the supplied predictor, per posterior draw —
and `type` becomes a four-value vocabulary naming what the percentage is
measured towards. Alongside it, the substitution that put an estimate back on
the fitted scale is corrected, and the same correction is applied to the three
further sites that carried a copy of it.

Closes #195, #196, #39, #206, #268, #160, #161.

## Why it matters

**The reference was the maximum of the predicted curve, not the control.** For a
monotonic decreasing curve those are the same point. For a hormetic curve the
maximum is the peak at the *NEC*, so an ECx was reported as a decline from the
peak — the behaviour `hormesis_def = "max"` describes, applied unconditionally
while the argument that was supposed to select it had been commented out.
`nsec()` meanwhile was anchored on the control, so the two estimators disagreed
about the same curve. Anchoring both on the control also makes a hormetic ECx
well defined rather than ambiguous: the target lies below the control and the
rising limb lies above it, so the curve crosses the target exactly once.

**An inline transformation with arithmetic gave a wrong number, silently.** A
model fitted with `crf(log(x + 1))` had its estimate inverted as `log(x)`,
because the substitution replaced the parsed call's first argument slot rather
than the predictor variable inside it. Nothing errored and the value returned
was plausible. `vignette("example1")` fits `crf(log(raw_x + 1))` and reports an
ECx from it.

**A curve that never reached the target returned the lowest concentration
tested.** Both estimators took the grid point whose prediction was nearest the
target. Where the curve never declines that far, the nearest point is the first
one — so the control concentration was returned as the ECx, with nothing said.

## Evidence

**#196, measured.** `nec4param` fitted to `nec_data` on `crf(log(raw_x + 1))`,
2 chains, 800 post-warmup draws, seed 333, against the same model fitted through
a pre-computed `log(x + 1)` column. `fixef` agreement between the two fits is
exactly 0, so the fits are identical and only the estimate differs:

| EC10 | median | 2.5% | 97.5% |
|---|---|---|---|
| 2.1.3 code | 1.372 | 1.333 | 1.409 |
| this branch | 1.598 | 1.567 | 1.628 |
| pre-computed column | 1.598 | 1.567 | 1.628 |

A 14.1 per cent error on the fitted scale. Back on the recorded concentration
scale via `xform = function(x) exp(x) - 1`, the reported EC10 is **2.94 against
3.94** — the offset dropped from the inverse, arriving as a 25 per cent
under-estimate of the concentration.

**The four types, on `manec_example`'s `nec4param`.** Each measures from the
same control towards a different floor, so the estimates order as the floors do.
Control 2.168; floors are `bot` −8.42 (relative), lowest predicted response
−4.13 (range), 0 (absolute):

| type | EC10 |
|---|---|
| absolute | 1.502 |
| range | 1.581 |
| relative | 1.673 |

**The NA rule fires on the packaged example.** `nsec(nec4param)` reports that 1
of 100 draws never falls below the control's 1 per cent quantile. Those draws
previously returned `max(x_vec)` and were averaged into the summary as though
they were estimates.

**#206** is measured on the issue: the six zero-bounded equations fit cleanly
under `gaussian` and model weights reject them where the shape is wrong.

**#161** is a separate reproduction, recorded on the issue: the reported call
shape now completes and every post-processing step returns. The issue records
the failure as a screenshot with no transcribed error text, so that is a
reproduction of the call rather than of the message; the comment says so and
says which two mechanisms fixed here would have produced a post-processing
failure on that input.

**#268 and #160** are pinned by the six tests PR #276 wrote against current
behaviour, each carrying an `INVERT THIS TEST WHEN #268 IS FIXED` comment naming
the assertion to change. All six are inverted here, and `test-plot.R` and
`test-autoplot.R` pass with 0 skips.

**Verification of the branch as a whole.** Full test suite under
`NOT_CRAN=true`: 57 files, 2556 assertions, 0 failures, 0 errors, 0 skips.
`roxygen2::roxygenise()` produces no diff, so `man/` and `NAMESPACE` match `R/`.
`R CMD check` passes on macOS, Windows, and Ubuntu release and devel.

<details>

### The rulings this implements

Recorded as D15 and D16 in `notes/implementation/03_decisions.md`, taken by RF
on 2026-09-06. They align `bayesnec` with toxval T8, T9 and T10 so that the
migration does not change these results a second time. Per D11 the estimator
fixes are made here rather than deferred to the migration.

### New shared helpers, all in `R/helpers.R`

- `sub_x_transformation(value, formula)` — substitutes on the predictor
  variable, found with `all.vars()`, rather than on `x_call[[2]]`. Replaces the
  block at `R/ecx.R:198-201` and its three copies. A `crf()` term naming more
  than one variable is refused rather than substituted into whichever comes
  first; `check_formula.bayesnecformula()` already rejects such a formula
  earlier, so this is a backstop.
- `crossing_x(y, target, x_vec)` — first crossing, interpolated between the
  bracketing grid points, `NA` where there is none. Replaces `nsec_fct()`, which
  is removed, and the `min_abs()` search in the `ecx_x_*` functions.
- `control_posterior(object, newdata, epred_fun, x_at)` — the control read at
  `min(observed x)` rather than at the grid's first column, so `x_range` does
  not move the reference. Takes the epred call as a closure so a two-block
  fit's `dpar` handling reaches the control as well as the curve. `x_at` lets a
  `bayesnechurdlefit` pin both components to one concentration.
- `hurdle_control_x(object)` — that concentration, taken from the survival
  component to match where `hurdle_component_preds()` takes its range.
- `family_has_lower_bound(family)` — decides whether `relative` has a finite
  denominator for a `bot`-free equation.
- `to_axis_scale(values, bdat, formula, x_grid_raw, xform)` — the #160 fix.
- `plot_ecx(object, family, dots)` — the ECx a plotting method annotates,
  written once so `plot()` and `autoplot()` cannot drift apart.
- `warn_censored_draws(values, estimate)` and `without_censored_warning(expr)` —
  the censoring report, raised as a classed `bayesnec_censored` condition so a
  method reporting on a vector it assembled can muffle the reports of the calls
  that built the parts.
- `check_removed_args(dots)` — refuses `hormesis_def` by name.

`modify_posterior()` is removed; it was the dead consumer of `hormesis_def`.
`manec_asymptote()` in `R/ecx.R` assembles the model-averaged theoretical
asymptote, stacked to match `posterior_epred.bayesmanecfit()` exactly — same
equation order, same `pull_draw_index()` — so element *i* of the asymptote is
the same (equation, draw) pair as row *i* of the curve it is the denominator
for.

### Files touched

`R/ecx.R`, `R/nsec.R`, `R/ecnsec.R`, `R/nec.R`, `R/check_models.R`,
`R/plot.R`, `R/autoplot.R`, `R/helpers.R`, `R/expand_classes.R`,
`R/bayesnechurdlefit-class.R`, `R/bayesnechurdlefit-methods.R`,
`R/compare_estimates.R`, `R/compare_posterior.R`, `R/average_estimates.R`.

### Consequences followed through to every caller

The control reference and the `NA` rule reach more than the three estimator
functions, and the second and third commits on this branch carry them through:

- **Every summary of a posterior that can now hold `NA` takes `na.rm`, and
  reports the count.** `nec()` on both classes, the four `bayesnechurdlefit`
  summaries, and the pairwise probability in `compare_estimates()`, which is
  computed with `mean()` and so returned `NA` for a whole comparison rather than
  erroring. `nsec.brmsfit` is the one exception, deliberately — see below.
- **The `bayesnechurdlefit` class implements D15 ruling 2 too.**
  `hurdle_component_preds()` returns a control read at the lowest observed
  concentration and pinned to the same concentration on both sides, computed
  there so the control and the curve are truncated to the same draws by the same
  operation. Measured on a real `bnec_hurdle()` fit, `x_range = c(0, 10)` now
  changes the ECx by 8e-6 and the NSEC by 1.1e-5.
- **`type` is validated against one definition wherever it is taken**, so
  `compare_estimates()`, `compare_posterior()` and `average_estimates()` accept
  `"range"` — the name the rename warning gives callers for the behaviour they
  had — and each warns once per call rather than once per fit.
- **`plot()` and `autoplot()` annotate the same quantity for a gaussian fit.**
  `plot()` asked for `type = "relative"` under its 2.1.3 meaning, the
  control-to-minimum span; `autoplot()` took the `ecx()` default. Both now ask
  for `"range"`, which leaves the drawn annotation unchanged from 2.1.3. Under
  the renamed `"relative"` that line would have annotated a third quantity,
  warned about a rename the caller had not asked for, and errored for a
  `bot`-free equation, which #206 has just made fittable under gaussian.
- **`ecnsec()` answers a model set under every percentage type**, rather than
  failing inside `posterior` on a `NULL` `fit`.
- **`to_axis_scale()` keeps the estimate's attributes**, so
  `autoplot(x, ecx = TRUE)` works for an inline-transformed predictor with
  `xform` at its default — the shape `vignette("example1")` uses. It previously
  ended in "replacement has length zero".

### Two things worth a reviewer's attention

**The #196 defect reached more than `ecx()`.** Three further sites carried the
same `x_call[[2]]` substitution: `expand_nec()`, which writes the stored *NEC*
or *NSEC* at fit time, and both hurdle methods. So a fit using an inline
transformation with arithmetic stored a wrong `nec` whether or not `ecx()` was
ever called.

**`relative` produces a lot of `NA` on `manec_example`, and that is correct.**
`nec4param`'s `bot` there is −8.42 with a 95 per cent interval of −13.58 to
−5.68, against an observed response range of roughly 0 to 2.2. It is an
extrapolated asymptote rather than an estimated feature of the data, so a
relative EC50 target of −3.12 lies below anything the curve reaches inside the
tested range and 23 of 100 draws return `NA`. `absolute` does not inherit that,
because 0 is fixed. This is the caution toxval T9 records, now visible rather
than hidden behind a nearest-grid-point answer.

### Edge cases covered by new tests

`sub_x_transformation` on `log(x)`, `log(x + 1)`, `log(x/2)`, `sqrt(x)`, a
pre-computed column, an `NA`-containing vector, and a two-variable term;
`crossing_x` on an exact crossing, an interpolated one, no crossing, an all-`NA`
draw, an `NA` target, and a hormetic curve where the crossing must be past the
peak; `ecx_asymptote` refusing a `bot`-free equation under `gaussian` and
returning 0 under `Gamma` (the draws lookup is mocked, because both packaged
fits have a `bot`); `ecx_val` of 0, negative and 150; every removed-argument
path; the `ecnsec`/`ecx` round trip, which is the assertion that the two answer
the same question of the same curve; a censored posterior reaching `nec()` and
the hurdle summaries; the hurdle control read off the grid rather than the data;
`to_axis_scale` carrying the `ecx_val` attribute; and `compare_estimates()`
returning a probability rather than `NA` when a draw is censored.

### What is deliberately not here

`type = "relative"` on a hurdle fit is refused rather than implemented: the
combined endpoint is `mu * (1 - hu)` and its asymptote is a product of two
equations rather than a fitted parameter. `absolute` and `range` are defined
there and work.

`nsec.brmsfit` has the same missing `na.rm` as the summaries listed above and is
left as it is. It is registered and exported but `@noRd`, so it has no rendered
help page, no test coverage and no vignette use, and `toxval` is implementing
that surface. `nsec.drc` does not share the defect — it has no `quantile()`
summary.

`vignettes/example1.Rmd` has its prose corrected in both the `.Rmd.orig` and the
rendered `.Rmd`, but the rendered numeric outputs of the `crf(log(raw_x + 1))`
fits are still pre-2.2.0 and need a re-render, which requires refitting. The
EC10 those chunks print is the 1.372 measured as wrong above. Tracked on #190,
which collects everything waiting on a `precompile.R` run; this branch adds one
item to that list and does not discharge it.

</details>

https://claude.ai/code/session_01SwfMNden64L4UuFCcscyLr


beckyfisher (commented):
## What was verified

Branch checked out and exercised under R 4.6.1 (Debian, WSL2). Every touched test
file passes with `NOT_CRAN=true`, 0 failures and 0 skips: `test-helpers.R` 57,
`test-ecx.R` 55, `test-nsec.R` 30, `test-ecnsec.R` 17, `test-plot.R` 49,
`test-autoplot.R` 25, `test-check_models.R` 32, `test-mu_support.R` 188,
`test-get_priors.R` 63; and, unchanged by this PR but downstream of it,
`test-compare_estimates.R` 5, `test-average_estimates.R` 7,
`test-compare_posterior.R` 6, `test-nec.R` 10,
`test-bayesnechurdlefit-methods.R` 60. `roxygen2::roxygenise()` produces no diff,
so `man/` matches `R/`.

Two claims in the PR body reproduce. The three types order as reported on
`manec_example`'s `nec4param` --- absolute 1.502, range 1.581, relative 1.673.
And `ecx()` on a `bayesnecfit` is now invariant to `x_range`: 1.501824 for the
default grid and for one extended two units below the data, which is D15 ruling 2
working.

The central change is sound. The findings below are at its edges --- sites the
change reaches but does not follow through to.

## Findings that raise an error on a valid call

**1. `nec()` errors when the no-effect posterior contains `NA`.**
`estimates_summary()` gained `na.rm = TRUE` (`R/helpers.R:245`); `R/nec.R:116`
and `R/nec.R:163` did not. `nsec_off_curve()` now writes `NA` into
`ne_posterior` for any draw whose curve does not fall to the control's `sig_val`
quantile (`R/expand_classes.R:64-75`), and `w_nec_calc()` carries those draws
into `w_ne_posterior`. Measured, injecting two `NA` into
`manec_example$w_ne_posterior`:

```
Error in quantile.default(unlist(nec_out), probs = prob_vals) :
  missing values and NaN's not allowed if 'na.rm' is FALSE
```

The packaged example predates the change so it does not fire there. Any set
refitted on this branch that includes a smooth equation whose curve does not
reach the reference will.

**2. Four further `quantile()` calls without `na.rm`, all newly reachable.**
`R/bayesnechurdlefit-class.R:159` (`nec`) and `:224` (`ecx`);
`R/bayesnechurdlefit-methods.R:266` (`nsec`) and `:932` (`ecnsec`). Confirmed
directly: `ecx_from_posterior()` on a two-draw matrix where one curve never
declines 50 per cent returns `c(NA, 5.556)`, and `quantile()` on that errors as
above. None of the four hurdle methods emits the NA-count warning either, so on a
hurdle fit the censoring is neither reported nor survivable.

**3. `compare_estimates()` rejects `type = "range"`.** `R/compare_estimates.R:41`
still tests against the old three-value set. Measured:

```
compare_estimates(list(a = manec_example, b = manec_example),
                  comparison = "ecx", type = "range")
#> Error: type must be one of "relative", "absolute" (the default) or"direct".
```

`"range"` is the name the rename warning gives users for the behaviour they had,
and `compare_estimates()` and `compare_posterior()` are the documented way to
compare an ECx across fits, so the migration path is unavailable there.

**4. `ecnsec(<bayesmanecfit>, type = "relative")` errors opaquely.**
`ecx_asymptote()` reads `object$fit`, which a `bayesmanecfit` does not have.
Measured:

```
ecnsec(manec_example, nsec = 1.5, type = "relative")
#> Error : Don't know how to transform an object of class 'NULL' to any
#>   supported draws format.
```

`"absolute"` and `"range"` work on the same object.

## Findings that change a reported number

**5. `plot()` asks for `type = "relative"` on every gaussian fit.**
`R/plot.R:116` and `:263`. Three consequences. The rename warning fires on
`plot(nec4param, add_ec10 = TRUE)` for a caller who never wrote `type`, measured.
The annotated value changes meaning: EC10 1.673 under the new `"relative"`
against 1.581 under `"range"`, which is what that line computed up to 2.1.3. And
with #206 removing the gaussian exclusion, `nec3param` under `gaussian` is now
fittable, while `ecx_asymptote()` refuses `"relative"` for a bot-free equation
under an unbounded family --- the PR's own test asserts that refusal --- so
`plot(fit, add_ec10 = TRUE)` on such a fit errors.

Separately, `ggbnec_data()`'s `add_ecx` path uses the default `type`, so `plot()`
and `autoplot()` annotate different quantities for the same gaussian fit. That
predates this PR, but the gap widens now that the two types differ by the `bot`
estimate rather than by the curve minimum.

**6. The hurdle methods do not implement D15 ruling 2.** `control <-
p_samples[, 1]` (`R/bayesnechurdlefit-class.R:216`), `reference <-
quantile(p_samples[, 1], sig_val)` (`R/bayesnechurdlefit-methods.R:258`) and
`p[1]` (`:926`) all read the control off the grid's first column, so `x_range`
still moves the reference on a hurdle fit. `control_posterior()` carries an
explicit `is_bayesnechurdlefit()` branch that nothing calls.

**7. `ecnsec.bayesnechurdlefit` ignores `type`.**
`R/bayesnechurdlefit-methods.R:926` computes `100 * (1 - p[k] / p[1])`, the
absolute form, whatever `type` is supplied, and calls neither
`validate_ecx_type()` nor `check_removed_args()`. So `type = "range"` silently
returns the absolute answer, an invalid `type` is accepted, and `hormesis_def` is
absorbed by `...` rather than refused. It is the only one of the eight affected
methods missing `check_removed_args()`.

## Documentation left behind

**8. `vignette("example1")` documents #196 as current behaviour.**
`vignettes/example1.Rmd:104` and the corresponding `.Rmd.orig`:

> Where the argument contains arithmetic of its own, such as `log(x + 1)`, `nec`
> is still reported on the transformed scale but `ecx` and `nsec` are not: they
> are back-transformed as though the offset were absent, and so are slightly in
> error. Pre-computing the transformation into its own column and passing it as a
> bare variable avoids this entirely.

This PR fixes both halves of that sentence --- the estimators and the
`expand_nec()` site that writes the stored *NEC* --- so the caution and its
recommended workaround are wrong as of this branch. The PR body names this
vignette as affected. Neither the source nor the rendered file is in the change
set, and the rendered outputs of the `crf(log(raw_x + 1))` fits are pre-2.2.0.

`vignettes/example2b.Rmd(.orig)` line 90/102 describes `hormesis_def` as an
argument of `ecx()`. Prose only, so the build is unaffected.

**9. `compare_posterior.R` `@details` has an orphan fragment.** Removing the line
`For \code{hormesis_def}, if "max", then ECx or NSEC values -- i.e.,`
(`R/compare_posterior.R:31`) left the three lines that continued it, so the block
reads "...estimated at a percent decline from the control. / depending on
argument `comparison` -- are calculated as a decline from the maximum estimates
(i.e. the peak at NEC); if "control", then...". The surrounding paragraph also
still gives the 2.1.3 definitions of `"relative"` and `"absolute"`. This ships in
`man/compare_posterior.Rd`.

<details>
<summary>Smaller points, none blocking</summary>

- **The censoring bound in the warning is on a different scale from the
  estimate.** `R/ecx.R:214-221` and `R/nsec.R:171-177` report
  `sub_x_transformation(max(x_vec), ...)` before `xform` is applied to the
  estimates. On `vignette("example1")`'s `xform = function(x) exp(x) - 1` the
  warning names a bound on the fitted scale beside estimates in concentrations.

- **`sub_x_transformation()` assumes the predictor is the first variable.**
  `all.vars(x_call)[1]` gives `x` for `crf(log(x + 1))` and `offset` for
  `crf(log(offset + x))`. Matching against
  `attr(model.frame(...), "bnec_pop")[["x_var"]]` would remove the ordering
  assumption.

- **`to_axis_scale()` clamps rather than declining to answer.**
  `approx(..., rule = 2)` draws an out-of-grid estimate at the axis limit,
  asserting a position it does not have, and #160 and #161 are about annotations
  landing where they should not. If the clamp is deliberate the reason belongs in
  the comment.

- **`crossing_x()` requires about 23 times the time of the search it replaces**:
  0.79 s against 0.034 s for 4,000 draws over a 1,000-point grid (R 4.6.1, WSL2,
  one core). Negligible against sampling, but `expand_nec()` calls
  `nsec_off_curve()` twice on the same posterior for a smooth hurdle block
  (`R/expand_classes.R:79` and `:108`).

- **The new fit-time `message()` in `nsec_off_curve()`** fires once per smooth
  equation during `bnec()`. On the default 23-model set that is a large amount of
  output for a condition the caller cannot act on until the fit returns.

</details>

Findings 1 to 4 each raise an error on a valid call and are the ones to fix
before merge. 5 and 8 are the ones a user would notice first.

https://claude.ai/code/session_01Sowtfnh9yrUf1a6eU8UfM8

beckyfisher:
Every finding in the review above is fixed in 027b9f9d, pushed to this branch.

## What changed

**Errors on valid calls.** `nec()` and the four `bayesnechurdlefit` summaries
called `quantile()` without `na.rm`, on posteriors this branch now writes `NA`
into; all six take it, and each reports the count through a classed
`bayesnec_censored` warning, so that a method summarising its components can
muffle theirs without muffling anything else they raise.
`compare_estimates()` and `average_estimates()` validate through
`validate_ecx_type()` rather than a second copy of the vocabulary, so `"range"`
is accepted and the rename is warned once for the call rather than once per fit.
`ecnsec()` on a model set with `type = "relative"` assembles a model-averaged
asymptote instead of failing on a `NULL` `fit`.

**Quantities.** `plot()`'s gaussian branch asks for `type = "range"`, not
`"relative"`. `"range"` is exactly what `"relative"` named up to 2.1.3, so the
drawn annotation is unchanged; under the renamed `"relative"` it would have
annotated the control-to-`bot` span instead --- EC10 1.673 against 1.581 on
`manec_example`'s `nec4param` --- warned about a rename the caller had not asked
for, and errored for a `bot`-free equation, which #206 has just made fittable
under gaussian. `autoplot()` is aligned to the same choice through a shared
helper; the two methods had annotated different quantities for the same fit.
The three hurdle estimators read the control from `hurdle_component_preds()`,
which now returns it read at the lowest observed concentration and pinned to the
same concentration on both sides, so D15 ruling 2 holds on that class too.
`ecnsec.bayesnechurdlefit` honours `type` rather than computing the absolute
form whatever was passed.

**Documentation.** The `vignette("example1")` paragraph that recorded #196 as
current behaviour, and recommended a pre-computed column as the workaround, now
records that the two forms agree and what the defect was.
`vignette("example2b")`'s `hormesis_def` paragraph is rewritten around the four
types. `compare_posterior`'s `@details` no longer carries the 2.1.3 definitions
or the orphan fragment the argument's removal left behind.

**The smaller points.** The censoring bound passes through `xform`, so it is on
the scale the estimate is read on. A `crf()` term naming more than one variable
is refused rather than inverted on whichever variable comes first.
`expand_nec()` memoises the `nsec_off_curve()` call a smooth hurdle block makes
twice, and names the equation in its message. `to_axis_scale()`'s `rule = 2`
clamp is kept, with the reason recorded: `rule = 1` would drop an out-of-grid
annotation with nothing said, which is the failure #160 and #161 report.

## Verification

905 assertions over 22 test files, 0 failures, 0 errors and 0 skips, under
`NOT_CRAN=true` on R 4.6.1 (Debian, WSL2). That covers the fitting paths which
run `expand_nec()` and the hurdle control, not only the estimator surface:
`test-expand_classes.R` 91, `test-bnec_group.R` 69,
`test-bayesnechurdlefit-methods.R` 68, `test-bnec_hurdle.R` 56,
`test-bayesmanec_methods.R` 51, `test-amend.R` 21. 45 of those assertions are
new. `roxygen2::roxygenise()` produces no diff.

## Two judgement calls worth a second opinion

**`plot()`'s gaussian branch could have taken `"absolute"` instead.** `"range"`
was chosen because it leaves the drawn annotation identical to 2.1.3, which is
the conservative reading of a line whose intent is recorded nowhere.
`"absolute"` would be the more consistent default and a silent change to every
gaussian figure.

**`ecnsec()` on a model set could have refused `"relative"` rather than
answering it.** It answers, because `ecx(<set>, type = "relative")` already works
by dispatching per equation, and a refusal would leave the two disagreeing about
what a model set supports. `manec_asymptote()` deliberately mirrors
`posterior_epred.bayesmanecfit()` --- same equation order, same
`pull_draw_index()` --- because element *i* of the asymptote must belong to the
same (equation, draw) pair as row *i* of the curve it is the denominator for.

## Still outstanding

`vignettes/example1.Rmd`'s prose is corrected in both the `.Rmd.orig` and the
rendered `.Rmd`, but its **rendered numeric outputs** from the
`crf(log(raw_x + 1))` fits are pre-2.2.0 and need a re-render, which means
refitting. The EC10 those chunks report is the 1.372 the PR body measures as
wrong, not the 1.598 this branch returns.

The three branches stacked on this one --- `batch-2-checks-and-reporting`,
`batch-4-group-scale` and `batch-3-nec-prior` --- are still on the previous base
and need restacking.

One correction to the review above: it proposed resolving the predictor from the
model frame's `bnec_pop` attribute. That would not have helped, because
`simplify_formula()` derives `x_var` with `all.vars(x_call)` and no subsetting,
so the attribute is the same guess. The ambiguous term is refused instead.

https://claude.ai/code/session_01Sowtfnh9yrUf1a6eU8UfM8

beckyfisher:
The PR body has been updated to describe the branch as it now stands, three
commits on from the version the review above was written against. Nothing in the
original What, Why it matters, or Evidence sections changed — the measurements
there still hold. What moved:

- **Evidence** gained a verification paragraph: full suite 2556 assertions, 0
  failures and 0 skips under `NOT_CRAN=true`; `roxygen2::roxygenise()` clean;
  `R CMD check` passing on macOS, Windows and Ubuntu release and devel.
- **New shared helpers** now lists `hurdle_control_x()`, `plot_ecx()`,
  `warn_censored_draws()` / `without_censored_warning()` and
  `manec_asymptote()`, and records the `x_at` argument added to
  `control_posterior()`.
- A new **Consequences followed through to every caller** section replaces
  nothing; it records the six places the control reference and the `NA` rule
  reach beyond the three estimator functions.
- **Files touched** adds `R/nec.R`.
- **What is deliberately not here** now records that `nsec.brmsfit` keeps the
  missing `na.rm` deliberately, and that `vignettes/example1.Rmd` still needs a
  re-render for its printed numbers.

The review comment above and my reply to it are unedited, and both remain
accurate about the code as it was at `027b9f9d`.

Issue #266, opened 2026-09-02
make_good_inits() spends up to ten thousand attempts before falling back, 561 seconds for one model on a twenty-row dataset


beckyfisher:
## What

Where `make_good_inits()` cannot find initial values inside the range of the
response, it exhausts `n_trials = 1e4` attempts before giving up. On a
twenty-row dataset that took 561 seconds for a single model, after which the
fit proceeded on Stan's own random initialisation, which is what it would have
used had the search stopped at the first attempt.

## Why it matters

The cost is paid per model. A `bnec()` call over the default `decline` set
would spend hours in the search before any sampling begins, and there is no
message while it runs: the last thing the user sees is "Finding initial values
which allow the response to be fitted using a nec3param model and a Gamma
distribution", and then nothing for as long as the search lasts. A user
watching that has no way to tell a long search from a hang.

The search is worth having; the objection is only to the cap. Ten thousand
attempts is not a bound that anyone would choose knowing what each attempt
costs, and the outcome after exhausting it is the same fallback that is
available immediately.

## Evidence

`dev` at 5f6ffe72, R 4.6.1, one core. `brm()` mocked, so the measurement is the
work `fit_bayesnec()` does before sampling and nothing else.

```r
d <- data.frame(x = rep(c(1, 5, 20, 100), each = 5))
d$y <- c(rep(c(8, 6, 3), each = 5), rep(0, 5))
bayesnec:::fit_bayesnec(formula = bnf(y ~ crf(x, model = "nec3param")),
                        data = d, model = "nec3param",
                        brm_args = list(family = Gamma(link = "identity")))
#> elapsed 561.2 s
#> Finding initial values which allow the response to be fitted using a
#>   nec3param model and a Gamma distribution.
#> bayesnec failed to find initial values within the range of the response.
#>   Using Stan's default initialisation process.
```

`Rprof` over a 60-second window of the same call attributes the time to
`get_init_predictions` (51% of total time) and `check_init_predictions` (27%),
called from the retry loop in `make_good_inits()`:

```
                         self.time self.pct total.time total.pct
"FUN"                         6.76    11.26      56.82     94.61
"check_init_predictions"      6.46    10.76      16.32     27.17
"get_init_predictions"        6.32    10.52      30.72     51.15
```

The same data with `x = c(0.1, 1, 10, 100)` instead completes in 3 to 12
seconds, so whether the search succeeds is sensitive to the predictor values in
a way that is not visible to the user.

[...]

Issue #262, opened 2026-08-27
dispersion(): report the posterior probability of over-dispersion, and state that beta_binomial does not address under-dispersion


beckyfisher:
## Summary

`dispersion()` returns a posterior for the ratio of observed to simulated
Pearson residual sums of squares, whose null value is 1. `summary = TRUE`
reduces that posterior to a median and an equal-tailed 95% interval, which
leaves the user to make the decision from a point estimate or from one tail
quantile. Neither uses the posterior.

The natural summary is the posterior probability of a departure:
`mean(disp > 1)`. It uses the whole posterior, it is directly interpretable, and
it is symmetric --- the same number read the other way answers the
under-dispersion question, which nothing currently addresses.

## The present position

The two rules available to a user are both poor:

- **Threshold the point estimate.** This discards the uncertainty the statistic
  was computed to express. `vignette("example1")` avoids it, describing values
  above 1 as over-dispersion and below 1 as under-dispersion with no cutoff.
- **Require the lower bound to exceed 1.** Coherent, but blunt. Each draw
  compares one observed residual sum against a single simulated replicate, so
  the interval's width at a typical design is dominated by replicate-to-
  replicate simulation noise. On the `nassarius` contaminant A subset, 34 tanks
  across 11 doses, the null interval computed from the fitted model spans
  roughly [0.33, 2.99] against an observed [0.57, 4.67]. A `Q2.5 > 1` rule fires
  only on gross over-dispersion, and a user applying it has no way to tell a
  screen that found nothing from a screen that could not have found anything.

A probability makes the second case visible: a value near 0.5 says the screen is
uninformative, where `Q2.5 = 0.57` does not obviously say so.

## Proposed

**1. Report `P(dispersion > 1)` from `summary = TRUE`**, alongside the existing
median and interval. One additional named element; no change to the default
return for `summary = FALSE`.

**2. Say what it does not answer.** A probability near 0 indicates
under-dispersion, and **`beta_binomial` does not address that case.** A
beta-binomial is a binomial with `p` mixed over a beta distribution, with
variance `n p (1 - p) [1 + (n - 1) rho]` and `rho >= 0`, so it is bounded below
by the binomial variance and can only ever add spread. The same holds for
`negbinomial` against `poisson`. Under-dispersion in counts points at the fit,
the design, or a constraint in the data-generating process, not at the family,
and the documentation should say so rather than leaving the user to infer that
the remedy is symmetric. It is not.

[...]

PR #282, opened 2026-09-06
Check what bnec() is given, and record what it changed (#274, #271, #272, #266, #93, #262, #261, #218)

### LINKED ISSUE #274: update() with newdata reports a boundary correction and then discards it, so the fit fails naming the condition the package said it had repaired

beckyfisher:
## What

Eight defects and reporting gaps in what `bnec()` checks before it fits, and in
what it tells you afterwards. Grouped into one pull request because they share
the pre-fit path and the objects it produces; the sections below are per issue.

Closes #274, #271, #272, #266, #93, #262, #261, #218.

## Why it matters

**Three of them make a fit fail, or silently alter the data.** `update()` with
`newdata` told the user their response had been repaired and then handed brms
the unrepaired frame, so the fit failed naming the condition just reported
fixed. A `disp(~log(x))` on a predictor containing a zero reached Stan and
stopped there, with nothing naming the term. An integer response with negative
values failed reporting a `family` argument the user had not supplied.

**Two are about time and attention.** The initial-value search ran up to ten
thousand attempts — 561 seconds for one model on a twenty-row dataset, per
model — with no output, and the outcome after exhausting it is the fallback
available at the first attempt.

**Three are about what can be recovered afterwards.** Which equations `bnec()`
declined to fit, and what it substituted in the response, were told to the
console once and then discarded. Neither survives a knitted document or a
`suppressMessages()` call, and both are what a methods section has to state.

## Evidence

**#274** is reproduced on the issue: `has_family_changed()` emits *"Your
response contains zeros ... They have been shifted"* and returns a bare logical,
with the user's `d` unchanged and `newdata = quote(newdata)` passed to
`brms::update()`.

**#271**, measured on `d <- data.frame(x = rep(c(0, 1, 10, 100), each = 5), ...)`
with `disp(~log(x))`: `make_standata()` warns *"Found infinite values in the
data"* and `range(sdata$X_shape[, "logx"])` is `-Inf 4.60517`. Confirmed on this
branch that the same call now stops in `bnec()` with the term named, before any
model is compiled.

**#272**: `set_distribution(c(-3L, 0L, 4L), support_integer = TRUE)` returned
`NULL`; it returns `"gaussian"` here, which is what `set_distribution(c(-3, 0, 4))`
already returned for the same values stored as numeric.

**#266** is measured on the issue at 561 s for a single model with `brm()`
mocked, so the measurement is the pre-sampling work alone.

**#93**: two of the three response corrections emitted nothing at all. Measured
here on a Beta response with three zeros and two ones, both are now reported
with the row count and both appear in `bnec_record(fit)$substitutions`.

**#218**: `m[sample(seq_len(n_samples), replace = FALSE)]` permutes only the
first `n_samples` elements, so the tail of a longer posterior was never drawn
from. Harmless at equal draw counts, systematic otherwise.

<details>

### Rulings

D16 in `notes/implementation/03_decisions.md` settles the two that were RF's:
#93 messages once per call and records the substitution on the fit, and #206's
sibling decisions. The rest are defects with one correct answer.

### The structural change, and why it is shared

Three of these issues are the same shape: `check_data()` runs **once per model**
inside `fit_bayesnec()`, so anything it reports is repeated for every member of a
model set, and anything it computes is thrown away unless the caller keeps it.
The existing placement rule — a refusal that is a property of the data and the
formula together is raised in `bnec()` before the model loop, with
`check_data()` keeping it as a backstop — is applied to the response
substitutions as well:

- `check_data()` still performs the corrections and now **returns** them, as a
  `substitutions` data frame.
- `report_substitutions()` emits them, called once from `bnec()` and once from
  `check_update_data()`.
- `attach_bnec_record()` stores them on the fitted object, following
  `attach_failed_models()`.

`check_disp_finite()` is raised from `bnec()` and `bnec_group()` rather than
from `check_data()` for the usual reason and for a second one: `check_data()` is
given the model frame, and a `disp()` term's variables are deliberately kept out
of it, so it could not perform the check even if it were the right place.

### `has_family_changed()` becomes `check_update_data()`

It now returns `list(changed_family = , data = , substitutions = )`, and
`update.bnecfit()` uses the corrected frame as `newdata`. Both routes into the
check are covered: `update(family = )` with no `newdata` reads the fit's stored
data, so the corrected frame is passed there too. The frame is substituted
whenever a correction was actually made, and `newdata` stays `NULL` where none
was, which is what tells brms to reuse the stored data rather than treat it as
new. Internal and unexported, so the rename costs no user code.

### New exported function

`bnec_record(x)` returns `requested`, `attempted`, `excluded` (a data frame of
equation and reason) and `substitutions`. `check_models()` accumulates the
exclusions through a local `note_drop()` at each of its six drop sites and
attaches them to its return; `bnec()` strips that attribute once it has read
it, so the model name it passes on stays a plain string.

`attempted` rather than `fitted`: an equation that was attempted and failed to
sample appears in it, so `requested` is partitioned exactly by `attempted` and
`excluded$model`. The record is kept through `update()` and rebuilt by
`amend()` for the set that call produced, so the message stating that the
substitution is recorded on the fitted object is true wherever it is emitted.
Both re-attach only where the input already had a record, so a `NULL` keeps its
documented meaning.

### #266's bound is on time, not attempts

`max_seconds = 10` alongside `n_trials = 1e3`, whichever comes first, checked
once per attempt. Time is the quantity the caller is paying and an attempt is
not a fixed amount of it. The fallback message now names the model, the number
of attempts and the elapsed seconds, so a long search can be told from a hang.

### #262 changes a column count

`dispersion(summary = TRUE)` returns four elements rather than three, and
`summary()$mod_weights` gains `dispersion_P_over_1`. Anything indexing that
table by position rather than by name would see the change; nothing in the
package does.

### #218 is documentation plus the length fix only

The pairing itself is unchanged, deliberately. The issue's proposed resolution
is documentation, a constraint recorded for #33, and the unequal-length fix;
replacing the O(n) approximation with the exact `mean(outer(A, B, "-") > 0)` is
16M operations at 4000 draws and is not attempted here.

### Tests

New or inverted: the #274 pin in `test-check_data.R` (inverted — the return is
no longer a bare logical, and the zeros are shifted in the frame returned); the
#93 silence pin (inverted — all three corrections are now reported, and none of
them from `check_data()`); `check_disp_finite()` on `log(x)` with a zero, on
`1/(x - 1)`, on a finite term, on a variance function, on no `disp()` term at
all, and on the same term with data that has no zero; `set_distribution()` on
negative integers, non-negative integers, with and without trials, and the
numeric equivalent; the exclusion record partitioning the requested set exactly,
and the empty-record edge case; and the initial-value budget, asserted on
elapsed time rather than on attempts.

</details>

https://claude.ai/code/session_01SwfMNden64L4UuFCcscyLr


beckyfisher:
Review of the branch as at 8b45295. Nine defects, verified against the branch rather than the diff alone. The structural argument holds: returning the substitution record from `check_data()` rather than only messaging it is the right application of the placement rule already used by `check_normalisation()` and `check_inline_boundary()`. The #218 sampling fixes, the #272 `set_distribution()` branch, the #262 four-element dispersion summary and the `check_models()` exclusion accounting are correct as written.

Two of the nine mean the pull request does not close the issue it names.

## Blocking

**1. #274 is not closed on the `newdata = NULL` route.** `R/bnecfit-methods.R:174-180`. The corrected frame is assigned back to `newdata` only inside `if (!is.null(newdata))`, but `check_update_data()` also runs on the `newdata = NULL, family = <new>` branch, where it reads `object[[1]]$fit$data`. On that route `report_substitutions()` states that the response has been shifted and recorded, and `brms::update()` is then called with `newdata = NULL`, so brms refits on the stored, unshifted data. Reproduction: a gaussian fit whose response contains an exact zero, updated with `family = Beta(link = "identity"), force_fit = TRUE`. The message states the boundary was repaired; Stan then fails on that boundary. This is #274's failure mode on the second of the two routes, and the new message makes the false claim explicit where the previous code was only silent. `update.bnecfit()` does not re-enter `fit_bayesnec()`, so no later write-back recovers it.

**2. The substitution report is discarded for any model set.** `R/helpers.R:946-965`. `data` is written back inside the loop, so iteration 2 rebuilds `bdat` from the corrected frame and `substitution_record()` returns `NULL`. `report_substitutions()` is then called after the loop on `checked_df`, which is the last iteration's result. For a `bayesmanecfit` — the object `recover_prebayesnecfit()` produces and `update.bnecfit()` passes — nothing is reported, while the same call on a single-model `bayesnecfit` reports normally. The comment above the call states the correct intent, but the mechanism depends on every iteration producing the same record, which the write-back guarantees it does not. The new test at `tests/testthat/test-check_data.R:373` passes a length-1 list, so it does not cover this.

## Should be fixed before merge

**3. `get_priors()` loses a message it previously emitted.** `R/check_data.R:284-292`, `R/get_priors.R:163`. The comment states the nudges are reported from "`bnec()`, `bnec_group()`, `get_priors()` and `update()`", but `report_substitutions()` has two call sites: `R/bnec.R:640` and `R/helpers.R:965`. `get_priors()` calls `check_data()` per model and reports nothing. Before this change the Gamma-zeros message was emitted from `check_data()` on that route, so priors are now derived from a shifted response with no output.

**4. The exclusion record contaminates `$model` on every single-model fit.** `R/bnec.R:626-627`, `R/check_models.R:251`. `attr(model, "excluded")` is read and never stripped. The multi-model branch is safe because `model[m]` drops it; the single-model branch passes `model` unstripped to `fit_bayesnec()`, which stores it as `out$model` (`R/fit_bayesnec.R:154`), and to `expand_nec(model = model)`, which forwards it as brms' `model_name`. Every single-model `bayesnecfit` then has a `$model` that prints a data frame attribute and returns `FALSE` from `identical(fit$model, "nec3param")`, and the record is serialised twice. `check_models()`'s own comment at lines 74-79 anticipates this.

**5. `check_disp_finite()` discards the two refusals it should surface.** `R/check_data.R:159-162`. `try(parse_disp_term(formula), silent = TRUE)` returns invisibly on error, discarding "Your formula has more than one `disp()` term" and "Could not parse the `disp()` term" from `R/disp_model.R:70-84`. A formula with two `disp()` terms passes the pre-loop check and raises inside `add_brm_defaults()` once per model within `try()`, which is the repeated-refusal pattern this pull request removes elsewhere.

## Minor

**6.** `R/check_data.R:183`. `!all(is.finite(values))` treats `NA` as non-finite, and `check_complete_cases()` runs on the model frame, which by design excludes `disp()` columns. A missing value in a dispersion-only column produces an error naming "log() of a zero and division by a zero" as the usual causes, which is the wrong diagnosis.

**7.** `R/bnec_record.R:66`. `attach_bnec_record()` is called only from `bnec()`, so `bnec_record()` returns `NULL` after `update()` or `amend()`. The documented meaning of `NULL` at line 29 is "an object fitted by a version that did not record it". This interacts with finding 1: the new `update()` message states the substitution "is recorded on the fitted object" when on that route it is not.

**8.** `R/bnec_record.R:64`. `fitted` holds the equations attempted, so a model that failed to sample appears in both `fitted` and `failed_models()`.

**9.** `R/inits_functions.R:341`. The fallback message reports `n_t - 1` attempts, but one draw is made before the loop at line 304, so the count understates the work by one.

<details>
<summary>Verification detail</summary>

Reviewed against `git diff origin/batch-1-ecx-definition...origin/batch-2-checks-and-reporting` with the surrounding files read in full: `R/check_data.R`, `R/bnec.R`, `R/bnec_group.R`, `R/helpers.R`, `R/check_models.R`, `R/fit_bayesnec.R`, `R/get_priors.R`, `R/expand_classes.R`, `R/disp_model.R`, `R/dispersion.R`. All eight CI jobs pass on 8b45295, so none of these are caught by the current test suite.

Findings 1, 2 and 4 were each traced to the object they reach rather than inferred from the diff: finding 1 to the `newdata = NULL` argument brms receives, finding 2 to `checked_df` being the last iteration's binding, finding 4 to `out$model` in `fit_bayesnec()` at line 154.

Two decisions are needed rather than deducible, and are noted here rather than fixed:

- Finding 3 can be resolved by adding a `report_substitutions()` call to `get_priors()` before its `lapply()`, computed once from `substitution_record()` as `bnec()` does, or by correcting the comment to name the two entry points that do report. The first matches D16 as stated; the second is the smaller change.
- Finding 8 is a naming question on a newly exported return value: rename `fitted` to `attempted`, or keep the name and intersect with the models that sampled.

</details>

beckyfisher:
All nine findings above are fixed in f0a77f8d.

The two blocking ones: `update(family = )` with no `newdata` now passes the corrected frame, and keeps `newdata` at `NULL` where nothing was corrected; the substitution record is captured on the first loop iteration, so a model set reports it. The other seven are as described above, with two decisions taken rather than assumed — the record is kept through `update()` and rebuilt by `amend()`, and `fitted` is renamed `attempted`.

The body has been edited, since two of its paragraphs no longer described the change. The `check_update_data()` paragraph said the frame is substituted only when `newdata` was supplied, which was the defect; it now states that both routes are covered. The `bnec_record()` paragraph named `fitted`. Nothing else in the body moved.

Measured on this branch: 19 test files, 1330 assertions, no failures, no errors, no skips, under `NOT_CRAN=true` on R 4.6.1. That covers every file the change touches plus `test-bnec.R`, `test-amend.R`, `test-cens.R`, `test-hurdle_family.R`, `test-fit_bayesnec.R` and `test-expand_classes.R`. It does not cover `bnec_group()`, `bnec_hurdle()` or `bnec_joint()` end to end.

<details>
<summary>Two things found while fixing, and one not fixed</summary>

**A test helper leaked global state.** `capture_brms_update()` traced `update.brmsfit` in brms's namespace, but an S3 method is registered in the *generic's* namespace table, so it is reached through `stats`. `untrace(where = asNamespace("brms"))` restored only the brms binding, and `getS3method("update", "brmsfit")` stayed the tracer for the rest of the session. It never showed because `test-bnec_record.R` sorts ahead of `test-bnecfit-methods.R`, so no later file called a real `update()`. The helper now writes a stub into that table and restores it on exit, which cannot leave anything behind.

**The record lost an element.** `bnec_rec$substitutions <- checked$substitutions` deletes the name when the value is `NULL`, so an update that substituted nothing returned a three-element record where `bnec()` and `amend()` return four. Single-bracket assignment of a one-element list instead, pinned by `expect_named()` on both the update and the amend leg.

**Not fixed, and pre-existing.** `expand_manec()` returns a bare one-element list when all but one model succeeds. `bnec()` and `amend_model_set()` both branch on `length(mod_fits) > 1` and route that through `expand_nec()`; `update.bnecfit()` does not, and classes it `c("bayesmanecfit", "bnecfit")` unconditionally. So `update()` on a `bayesmanecfit` where all but one refit fails returns an object claiming a class whose structure it does not have. Untouched by this branch — the only edit to that line was assigning the result to a variable — so it belongs in its own issue rather than here.

</details>

PR #286, opened 2026-09-06
Peak the nec and ec50 gamma prior at the median predictor, not twice it (#273)

### LINKED ISSUE #273: The default nec and ec50 gamma prior peaks at twice the median predictor, so on a linearly spaced series its mode sits on the upper truncation bound
## What

The default `gamma` prior on `nec` and `ec50` does not peak where the
documentation says it does, and on a linearly spaced concentration series its
maximum density falls on or beyond the upper truncation bound. The prior is
then monotonically increasing across the whole range it permits, which favours
a `nec` at the top of the tested concentrations.

## Why it matters

`vignette("example3")` and the JSS article both describe this prior as centred
on the middle of the predictor. What is built is `gamma(5, 2 / m)` where *m* is
the median of the distinct predictor values, so the mean is `2.5 * m` and the
maximum density is at `2 * m`. For a series spaced evenly from zero, `2 * m` is
the maximum of the series, so the truncation at `max(x)` removes the entire
declining half of the density.

A prior that rises monotonically toward the upper bound is not weakly
informative about a *NEC*. Where the data are uninformative — few
concentrations, a shallow response, a small sample — the posterior inherits
that slope, and the estimate is pulled toward the highest concentration tested.

## Evidence

R 4.6.1, `dev` at 130aede3. `R/define_prior.R` builds
`gamma(5, 1 / (m / 2))`, and `prior_string(..., lb = min(predictor),
ub = max(predictor))` truncates it. For shape 5 and rate *r* the mean is `5 / r`
and the mode is `4 / r`, so with `r = 2 / m` these are `2.5 * m` and `2 * m`.

```
levels                 median   rate     mode     mean    max(x)   truncated median
0, 5, 15, 45, 135        15.0   0.1333    30.0     37.5    135.0    35.0
0, 1, 10, 100             5.5   0.3636    11.0     13.8    100.0    12.8
0, 25, 50, 75, 100       50.0   0.0400   100.0    125.0    100.0    75.1
```

The third row is the case that matters. The mode sits exactly on the upper
bound, the truncated prior increases across the whole of 0 to 100, and its
median is 75.1 on a range of 0 to 100 — three quarters of the way up a series
the user chose to bracket the response.

A logarithmically spaced series, which is the more common design in this field,
is much less affected: the first two rows put the mode well inside the range.
So the defect is worst for the linearly spaced designs typical of a dilution
series prepared by equal steps.

Found while reviewing #270, which corrects the `vignette("example3")` sentence
that stated the location wrongly. The prior itself is unchanged by that PR and
by every release to date.

<details>

`R/define_prior.R`, the `Gamma` entry of `x_prs`, and the `pr_nec` and `pr_ec50`
`prior_string()` calls beside it. The `Beta` entry, `beta(2, 2)`, is symmetric
on (0, 1) and does peak at the centre; the `gaussian` entry is centred on the
median. The `Gamma` entry is the only one of the three that does not do what the
documentation describes, which suggests the shape and the rate were chosen
independently of each other.

Two candidate corrections, and the choice needs a measurement rather than an
argument:

- **Keep shape 5 and set the rate to `4 / m`.** The maximum density is then at
  *m* and the mean at `1.25 * m`. This is the smallest change and the one the
  documentation already claims.
- **Keep the rate at `2 / m` and set the shape to 2.** The mean is then *m*
  exactly, which is the relation the response-scaled priors use — those are
  `gamma(2, 1 / (q / 2))` for a quantile *q*, and the vignette's `mean = s * (1 / r)`
  is applied correctly there. This makes the predictor-scaled and
  response-scaled priors consistent with one another. It also flattens the
  prior considerably, shape 2 having a much heavier right tail than shape 5.

Either is a change to every default `nec` and `ec50` prior, so it needs to be
measured before it is chosen: refit a set of reference datasets under both and
report the change in the `nec` posterior, in particular for the linearly spaced
designs where the current prior is worst and for a dataset with few
concentrations, where the prior contributes most.

Note the interaction with #269, which changed *m* from the median of the
observations to the median of the distinct values. That change does not affect
this one: both readings of *m* are subject to the same shape and rate.

**Definition of done.** The `nec` and `ec50` prior has its maximum density
inside the tested predictor range for every spacing, `vignette("example3")`
states the relation that the code implements, and the change in the `nec`
posterior on the reference datasets is recorded.

</details>

beckyfisher:
## What

The default `nec` and `ec50` gamma prior places its maximum density at the
median of the distinct predictor values, not at twice it. The rate changes from
`2/m` to `4/m`.

Closes #273.

## Why it matters

**The three predictor-scaled priors should place their maximum density at a
central measure of the predictor, and two of the three do.** `x_prs` is selected
on the predictor's distribution, so it has three entries:

| predictor | prior | maximum density at |
|---|---|---|
| bounded on (0, 1) | `beta(2, 2)` | the centre of the interval |
| unbounded | `normal(median(x), 10 sd(x))` | the median |
| non-negative | `gamma(5, 2/m)` | **`2m`** |

The mode of `gamma(shape, rate)` is `(shape - 1) / rate`, so at rate `2/m` the
gamma entry peaks at twice the median while the other two peak at the centre of
what they describe. `gamma(5, 4/m)` peaks at *m*, which makes the three
consistent.

**It is also what the documentation states.** `?bnec` and
`vignette("example3")` describe a prior whose maximum density is at *m*. The
gamma entry was the only one of the three that did not do what it was documented
to do, which is what #273 reports.

**And the current placement has a consequence that is a property of the
arithmetic, not of any dataset.** `nec` and `ec50` are truncated to the observed
predictor range. For a series spaced evenly from zero the median is close to
half the maximum, so `2m` is close to the largest concentration tested: after
truncation the density is monotonically increasing across the entire range the
prior permits. A weakly informative default should not place its greatest mass
at the boundary of its own support, and for a protective estimate favouring the
highest concentration tested is the wrong direction.

## Evidence

Arithmetic, against the upper truncation bound. `m` is the median of the
distinct predictor values:

| design | rate | mode | mean | truncation bound | mode inside |
|---|---|---|---|---|---|
| doses 0, 25, 50, 75, 100 | `2/m` | **100** | 125 | 100 | **no** |
| doses 0, 25, 50, 75, 100 | `4/m` | 50 | 62.5 | 100 | yes |
| doses 0.1, 1, 10, 100, 1000 | `2/m` | 20 | 25 | 1000 | yes |
| doses 0.1, 1, 10, 100, 1000 | `4/m` | 10 | 12.5 | 1000 | yes |

On the linearly spaced design the mode falls exactly on the bound. `(5 - 1) / (2 / m)`
equals `max(x)` there identically, which the new test asserts.

**This changes every default fit** whose predictor is non-negative, and it
applies to `ec50` as well as to `nec`, since both read the same `x_prs` entry.

## The alternative considered

`gamma(2, 2/m)`, whose *mean* rather than mode is *m*. Rejected: its maximum
density is at `m/2`, so it matches neither the documented behaviour nor the
convention the other two entries follow, both of which are specified by where
the density peaks rather than by where its mean falls.

<details>

`R/define_prior.R`, the `Gamma` entry of `x_prs`. `pr_nec` and `pr_ec50` both
read that entry, so both change together, which is intended: they are the two
predictor-scaled parameters and there is no reason for them to differ.

**The `Gamma` label names the predictor, not the response family.** `x_prs` is
indexed on `x_type`, which is `set_distribution(predictor, ...)` at
`R/define_prior.R:156`. A non-negative predictor takes this entry whatever the
response family is; `Beta`, `Gamma`, `gaussian` and `poisson` responses on the
same predictor all receive an identical `nec` prior. The labels in `x_prs` reuse
family names for predictor types, which is worth knowing before reading that
line.

### Documentation

`vignettes/example3.Rmd.orig` described the old rate and its consequence
accurately — #270 corrected that sentence — so it becomes wrong with this change
and is rewritten. The rendered `.Rmd` is not regenerated here; per #190 a stale
rendered vignette is not a merge blocker, and the full precompile runs once after
the vignette pull requests settle.

### Tests

Three existing assertions in `test-define_prior.R` state the rate as a value
rather than recomputing it, which is deliberate — they say what the prior should
be, not how `define_prior()` arrives at it — so all three are updated:
`gamma(5, 0.0444…)` becomes `gamma(5, 0.0888…)` and `gamma(5, 0.3077…)` becomes
`gamma(5, 0.6154…)`. Those three are #269's tests about the distinct-value
median and are unrelated to this change except in the constant they quote.

The new test asserts the property rather than the constant: that the mode is at
*m*, that it is strictly inside the truncation bounds, and — for contrast — that
`(5 - 1) / (2 / m)` equals `max(x)` exactly on the linear design, which is the
defect stated as an assertion.

### Interaction with #269

#269 changed *m* from the median of the observations to the median of the
distinct values. Both readings are subject to the same shape and rate, so that
change and this one are independent.

</details>

https://claude.ai/code/session_01QVqihYoQkGUsZBhfcZcgju


beckyfisher:
## The recommendation this PR implements is weaker than its body states

RF asked why the measurement behind this change fitted a `Beta` response when
the prior is the `Gamma` entry of `x_prs`. The branch under test was the right
one — `x_prs` is indexed on the *predictor's* distribution, not the response
family — but the objection behind the question was correct: a `Beta` response at
a residual sd of 0.04 is a strongly informative likelihood, which is the
condition under which the prior matters least, and I held the response family
and the noise level at their most favourable without saying so.

Repeated on a `Gamma` response at a coefficient of variation of about 0.35. Full
table on #273. Averaged over three true *NEC* positions:

| prior | mean abs. bias | mean interval width |
|---|---|---|
| current | **5.67** | 27.75 |
| shape 5, rate 4/*m* (this PR) | 9.34 | **20.79** |
| shape 2, rate 2/*m* | 8.50 | 26.24 |

The trade-off is the same in direction as the `Beta` runs and roughly twice the
size. **On both response families the current prior is the most accurate of the
three**, and the correction buys about a 25% narrower interval for roughly sixty
per cent more bias. I expected the weaker likelihood to strengthen the case for
correcting; it does the opposite.

**What is left in favour of this change** is not the fits. It is that on a
linearly spaced series the current prior's mode sits on the upper truncation
bound, so the truncated density rises monotonically across everything it
permits and favours a high *NEC* — anti-conservative for a protective estimate —
and that `?bnec` and `vignette("example3")` describe a maximum density at *m*
that the code does not deliver.

**A third option belongs on the table and was not offered**: leave the prior and
correct the documentation instead. That resolves the documentation mismatch and
not the geometry.

This PR is left as it stands, implementing `gamma(5, 4/m)`. Either alternative is
a one-line change in the `Gamma` entry of `x_prs` plus the three constants in
`test-define_prior.R` and the paragraph in `example3`. **Do not merge this one on
the strength of the measurement alone** — the measurement no longer supports it.

Issue #39, opened 2020-11-16
Improve speed and precision of exc and nsec functions using uniroot.all


beckyfisher:
ecx and nsec values are currently estimated using an x vector based on x_range and precision, with x estimated as the x value of this vector closes to the target y value. This would be more elegantly handled using uniroot - however noting that uniroot.all would be required in the case of an hormesis model, with the second solution (if there are more than one) most likely corresponding to the desired estimate.

beckyfisher:
## Left open deliberately — PR #281 claimed this and only half-delivered it

#281's body lists this issue as closed. That was an overstatement on my part and
the issue is being left open, with what actually changed recorded here so the
remainder is clear.

## What was done

**The precision complaint is resolved.** The estimate is no longer "the x value
of the vector closest to the target y". `crossing_x()` finds the crossing with
`modelbased::zero_crossings()` and then **interpolates linearly between the two
grid points that bracket it**, so the answer is no longer constrained to land on
a grid point. It also returns `NA` where the curve does not reach the target
anywhere in the predictor range, rather than the nearest point — which for a
curve that never declines that far was the *lowest* concentration in the series,
returned as the ECx with nothing said.

**The hormesis difficulty is dissolved rather than solved.** This issue proposes
`uniroot.all` because a hormetic curve can cross the target more than once, with
"the second solution most likely corresponding to the desired estimate". Under
the ruling in #195 every ECx is now measured from the **control** rather than
from the maximum of the predicted curve. The target therefore lies below the
control, the rising limb lies above it, and **the curve crosses exactly once** —
so there is no longer a second root to select, and no rule needed for choosing
between roots.

## What was not done

**The speed half.** `ecx()` and `nsec()` still evaluate `posterior_epred()` over
a grid of `resolution` points (1000 by default) per draw and search that grid.
Nothing here replaces the grid with root-finding on the fitted function, which is
what the issue asks for and where the speed would come from.

So the remaining work is narrower than the issue as written: **not** "use
`uniroot.all` to handle multiple roots", which is no longer needed, but "replace
the grid evaluation with a root-find", which is a performance change with no
expected effect on the answer beyond removing the residual grid dependence of
`resolution`.

## Worth knowing before taking it

Any such change should be measured against the current behaviour rather than
assumed faster: the grid is evaluated once per draw and reused, so a root-find
per draw is not obviously cheaper, and `posterior_epred()` on a 1000-point grid
[...]

PR #300, opened 2026-09-09
Keep dispersion() where an observation is reproduced exactly (#298)

### LINKED ISSUE #298: dispersion() discards the whole statistic when one observation is reproduced exactly
## What

`dispersion()` returns an empty vector, and reports a bad model fit, whenever a
single observation is reproduced *exactly*. One group at the boundary discards
the statistic for every posterior draw.

## Why it matters

The statistic is lost precisely for the equations that describe the data best,
and the message sends the user to look for a problem that is not there.

`dispersion()` builds a Pearson residual as `(obs_y - prd_y) / sqrt(prd_var_y)`,
with `prd_var_y = mu (1 - mu) n` for a binomial. Where the response is zero at
the highest concentration and the fitted mean **underflows to exactly zero**,
the denominator is zero and the numerator is zero too, because the model has
reproduced that group exactly. The residual is `0/0`. `rowSums()` then makes the
whole draw `NaN`, and `any(is.na(disp))` discards all draws:

```r
disp <- rowSums(prd_sr) / rowSums(sim_sr)
if (any(is.na(disp))) {
  message("Your model predictions have generated no residuals; this is",
          " most likely cause by a bad model fit. Ignoring dispersion",
          " calculation.")
  numeric()
}
```

Underflow requires a fast decay, so this strikes the equations that fit a steep
concentration-response best and spares the ones that fit it poorly. The
diagnostic is therefore least available where the model average puts most of its
weight.

The message is also wrong on the facts, and reads as a grammatical error:
"most likely cause by" for "caused by".

## Evidence

Knitting `vignettes/example9.Rmd.orig` against `dev` at `0976caca`, R 4.6.1,
brms 2.23.0, 2026-09-08. The `nassarius` contaminant A survival set, `binomial`,
`decline`, nine equations fitted. Three returned `NA`:

| equation | weight (screened set) | dispersion |
|---|---|---|
| `ecxsigm` | **0.767** | `NA` |
| `ecxwb1p3` | 0.133 | `NA` |
| `ecxwb2p3` | 0.017 | `NA` |
| `nec3param` | 0.042 | 1.60 |
| `ecxexp` | 0.002 | 1.75 |

`ecxsigm` holds 77% of the model weight and has a Bayesian R-squared of 0.91.
Its dispersion is unavailable; `ecxexp`, at R-squared 0.69 and 0.2% of the
weight, reports one. The message fired three times in the one call.

The corresponding configuration is visible in that vignette's `pp_check` figure:
at the highest dose all three tanks are at zero survival and the model
reproduces them with no simulated spread.

<details>

### Two cases, and they need opposite treatment

`prd_var_y == 0` reaches the same `NaN` by two routes, and only one of them is
benign.

1. **`prd_var_y == 0` and `obs_y == prd_y`.** The model is degenerate at that
   observation and the data agree with it exactly. The Pearson term is `0/0`,
   the simulated term is `0/0` as well, and the observation carries no
   information about dispersion in either direction. It should be **excluded
   from both sums**, and every draw retained.
2. **`prd_var_y == 0` and `obs_y != prd_y`.** The model assigns zero variance to
   a value it did not predict. The Pearson term is infinite, and that is a real
   and severe misfit. It must **not** be dropped silently. `Inf / Inf` currently
   also yields `NaN`, so this case is swallowed by the same branch and reported
   with the same misleading message.

### Proposed behaviour

Compute both sums with the degenerate-and-agreeing observations removed, keep
every draw, and report how many observations were excluded --- the pattern #39
established for a censored `ecx()` posterior, where affected draws are excluded
and the count is named rather than the whole result being discarded.

Where case 2 occurs, stop or warn naming the observations, because that is a
result rather than a numerical artefact.

Where *every* observation is degenerate, there is genuinely nothing to compute
and an empty vector is the right answer --- with a message that says so rather
than diagnosing a bad fit.

### Also

- The message needs rewriting; "most likely cause by a bad model fit" is both
  ungrammatical and, in this case, false.
- `expand_classes.R` writes `dispersion_Estimate`, `dispersion_Q2.5`,
  `dispersion_Q97.5` and `dispersion_P_over_1` as `NA` for such a fit. Once the
  statistic is computable those columns populate, so `summary()`'s weights table
  gains rows it currently leaves blank.
- The same arithmetic runs for `poisson`, where `variance(mu) = mu` underflows
  the same way.

### Definition of done

`dispersion()` returns a usable statistic for a fit in which some but not all
observations are degenerate; the two cases above are distinguished in the code
with the reasoning recorded; tests cover a fit with an exactly-reproduced group
and a fit with a zero-variance disagreement; a NEWS entry; and the paragraph in
`vignette("example9")` that currently explains the `NA` is narrowed to whatever
behaviour survives.

Whether case 2 should be an error or a warning. It is a modelling verdict, not
an implementation detail.

</details>

beckyfisher:
## What

`dispersion()` returned an empty vector, and reported a bad model fit, whenever
a single observation had a fitted variance of exactly zero. It now excludes such
an observation from the statistic and keeps every posterior draw.

## Why it matters

The Pearson denominator underflows to zero only for a curve that decays fast,
so the statistic was lost precisely for the equations that describe a steep
concentration-response best, and the accompanying message sent the user to look
for a problem that is not there. `expand_nec()` writes the four `dispersion_*`
columns of the weights table from this vector, so those columns were blank for
the same equations.

## Evidence

A fifteen-row binomial fixture with the two highest doses at zero survival,
`nec3param`, R 4.6.1, brms 2.23.0, 2026-09-09: the fitted mean underflows to
exactly zero at six of the fifteen observations. Before this change the fit
returned an empty vector and wrote `NA` into the weights table; it now reports a
dispersion of 0.97 with a 95% interval of 0.30 to 2.48, and names the six
observations it excluded. This fixture is added as a test. The reported case,
from `vignette("example9")`, is the `nassarius` contaminant A survival set,
where `ecxsigm` at 0.767 of the model weight and a Bayesian R-squared of 0.91
reported `NA` while `ecxexp` at 0.002 of the weight reported a value.

`tests/testthat/test-dispersion.R` passes with 36 tests, 0 failures, 0 warnings,
`NOT_CRAN=true`.

<details>

### The two cases

The per-draw loop in `dispersion()` now collects only the fitted variance, and
the residual arithmetic moves to a new internal `pearson_dispersion()`, which
handles the whole draws-by-observations matrix at once. That is where the two
routes to a zero denominator are separated.

1. **Zero variance, response equal to the fitted value.** Both the observed and
   the simulated term are `0/0` and the observation says nothing about
   dispersion in either direction, so it is excluded from both sums and every
   draw is retained. The exclusion is elementwise rather than a fixed set --- an
   observation whose fitted mean underflows in a steep draw gives an ordinary
   residual in a shallower one --- and the message names the observations and
   the draws affected. Measured on the fixture, 400 draws: 229 exclude nothing,
   60 exclude three observations and 111 exclude six. A draw in which no
   observation contributes a residual returns `NA`, and that count is named too,
   following `nsec_off_curve()`.
2. **Zero variance, response not equal to the fitted value.** The model has
   assigned zero variance to a value it did not predict. The observed Pearson
   term is infinite and is kept, so the statistic is reported as `Inf`, and a
   warning names the observations. The infinity is the underflow --- in exact
   arithmetic the residual there is large and finite --- but the misfit it
   reports is real. For a converged fit the case is close to unreachable, since
   the likelihood rejects a draw assigning zero probability to an observed
   value, so the warning says to check that the response and the fitted mean are
   on the same scale.

Case 2 warns rather than stops, which is the question the issue reserved.
`dispersion()` is called once per equation from `expand_nec()` and that call is
not wrapped, so an error would abandon construction of the whole
`bayesmanecfit` for one candidate whose shape is wrong. This is the placement
argument already recorded in `R/bnec.R` for `check_data()`.

A third case not raised in the issue is handled the same way: a variance that is
negative or not finite reaches the same `na.rm` and would otherwise disappear in
silence, so it is excluded explicitly and reported. It is set missing before the
square root, so base R raises no unattributed "NaNs produced" warning of its
own.

### Also

- The empty vector and its message are now reserved for the case where every
  observation is degenerate, and the message no longer diagnoses a bad model
  fit. The "most likely cause by" wording is gone with it.
- Each report names the equation, because `bnec()` calls this once per model.
- A draw in which every observation is excluded returns `NA`, which
  `estimates_summary()` already drops as it does a censored ECx draw.

### Tests

Ten unit tests exercise `pearson_dispersion()` directly on small deterministic
matrices --- exclusion, the `Inf` case, the labels, the all-degenerate return,
the non-finite and the negative variance, the per-draw report, the named `NA`
draws, the report with no equation name, and equality with the plain formula
where nothing is degenerate. Two tests use the fitted fixture: the end-to-end
case, and the empty-vector branch of `dispersion()`, reached by substituting the
return of `pearson_dispersion()` since a fit degenerate at every observation is
not one anyone would produce. The fixture is fitted once and reused; the file
runs in 5m22s.

### Not in this pull request

The `vignette("example9")` passage that explains the `NA` is on the branch of
#238, so it is narrowed there in a separate commit rather than here. That
vignette needs a re-knit before #238 merges, which it needs in any case.

</details>

Closes #298

https://claude.ai/code/session_01CGpuM3TVYB7RtMn9sj7Gk7

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -263,24 +263,38 @@
   (1 - mu) n` for a binomial and `mu` for a Poisson --- and the response at such
   an observation is the fitted value, so its residual is `0/0`. One of them made
   every draw `NaN`, and the whole vector was returned empty with a message
-  attributing it to a bad model fit. Such observations are now excluded from
-  both the observed and the simulated sum, every draw is retained, and a message
-  names them. Where instead the response differs from a fitted value of zero
-  variance, the observed Pearson residual is infinite; that is a property of the
-  fit rather than a numerical limit, so the statistic is reported as `Inf` and a
-  warning names the observations rather than the case being swallowed by the
-  same branch. It is a warning rather than an error because `dispersion()` is
-  called once per equation from `expand_nec()`, where stopping would abandon
-  construction of the whole `bayesmanecfit`. Measured on the `nassarius`
-  contaminant A survival set of `vignette("example9")` (binomial, `decline`,
-  nine equations, 2026-09-08, R 4.6.1): three equations returned `NA`, among
-  them `ecxsigm` at 0.767 of the model weight with a Bayesian R-squared of 0.91,
-  while `ecxexp` at 0.002 of the weight reported a value. The statistic was
-  therefore least available for the equations describing the data best, since
-  underflow requires a fast decay. `expand_nec()` writes the four `dispersion_*`
-  columns of the weights table from this vector, so those columns now populate
-  for such a fit. An empty vector is returned only where every observation is
-  degenerate, with a message that says so (#298).
+  attributing it to a bad model fit.
+
+  Such a term is now excluded from both the observed and the simulated sum and
+  every draw is retained. The exclusion is the limit rather than an
+  approximation to it: as the fitted mean tends to zero with a response of zero
+  both terms tend to zero, so excluding them is arithmetically identical to
+  contributing zero to each sum, and the ratio is self-normalising, so its null
+  value of 1 holds whatever the size of the retained set. It is made draw by
+  draw, since an observation whose fitted mean underflows in one draw gives an
+  ordinary residual in another, and a message reports the observations and the
+  draws affected. Measured on the fixture added with this change, 400 post
+  warm-up draws: 229 draws exclude nothing, 60 exclude three observations and
+  111 exclude six, and no observation is excluded in every draw.
+
+  Where instead the response differs from a fitted value of zero variance, the
+  statistic is reported as `Inf` and a warning names the observations, rather
+  than the case being swallowed by the same branch. The infinity is produced by
+  the underflow, since in exact arithmetic the residual there is large and
+  finite, but the misfit it reports is real. It is a warning rather than an
+  error because `dispersion()` is called once per equation from `expand_nec()`,
+  where stopping would abandon construction of the whole `bayesmanecfit`.
+
+  Measured on the `nassarius` contaminant A survival set of
+  `vignette("example9")` (binomial, `decline`, nine equations, 2026-09-08, R
+  4.6.1): three equations returned `NA`, among them `ecxsigm` at 0.767 of the
+  model weight with a Bayesian R-squared of 0.91, while `ecxexp` at 0.002 of the
+  weight reported a value. The statistic was therefore least available for the
+  equations describing the data best, since underflow requires a fast decay.
+  `expand_nec()` writes the four `dispersion_*` columns of the weights table
+  from this vector, so those columns now populate for such a fit. An empty
+  vector is returned only where no observation contributes a residual in any
+  draw, with a message that says so (#298).
 
 - `update()` on a `bayesmanecfit` returned an object classed `bayesmanecfit`
   with none of that class's structure whenever all but one model failed to
--- a/NEWS.md
+++ b/NEWS.md
@@ -257,6 +257,31 @@
 
 ## Bug fixes
 
+- `dispersion()` no longer discards the statistic where a single observation is
+  reproduced exactly. The Pearson denominator is the fitted standard deviation,
+  which underflows to exactly zero for a curve that decays fast enough --- `mu
+  (1 - mu) n` for a binomial and `mu` for a Poisson --- and the response at such
+  an observation is the fitted value, so its residual is `0/0`. One of them made
+  every draw `NaN`, and the whole vector was returned empty with a message
+  attributing it to a bad model fit. Such observations are now excluded from
+  both the observed and the simulated sum, every draw is retained, and a message
+  names them. Where instead the response differs from a fitted value of zero
+  variance, the observed Pearson residual is infinite; that is a property of the
+  fit rather than a numerical limit, so the statistic is reported as `Inf` and a
+  warning names the observations rather than the case being swallowed by the
+  same branch. It is a warning rather than an error because `dispersion()` is
+  called once per equation from `expand_nec()`, where stopping would abandon
+  construction of the whole `bayesmanecfit`. Measured on the `nassarius`
+  contaminant A survival set of `vignette("example9")` (binomial, `decline`,
+  nine equations, 2026-09-08, R 4.6.1): three equations returned `NA`, among
+  them `ecxsigm` at 0.767 of the model weight with a Bayesian R-squared of 0.91,
+  while `ecxexp` at 0.002 of the weight reported a value. The statistic was
+  therefore least available for the equations describing the data best, since
+  underflow requires a fast decay. `expand_nec()` writes the four `dispersion_*`
+  columns of the weights table from this vector, so those columns now populate
+  for such a fit. An empty vector is returned only where every observation is
+  degenerate, with a message that says so (#298).
+
 - `update()` on a `bayesmanecfit` returned an object classed `bayesmanecfit`
   with none of that class's structure whenever all but one model failed to
   refit. The guard tested the length of the candidate set going in rather than
--- a/R/dispersion.R
+++ b/R/dispersion.R
@@ -28,6 +28,20 @@
 #' posterior probability of under-dispersion, which no other summary here
 #' addresses.
 #'
+#' \bold{An observation the model reproduces exactly is excluded.} The Pearson
+#' denominator is the fitted standard deviation, which is exactly zero wherever
+#' the fitted mean underflows --- \code{mu (1 - mu) n} for a binomial and
+#' \code{mu} for a Poisson both do so for a curve that decays fast enough. An
+#' observation whose response equals that fitted value contributes \code{0/0}
+#' to both sums and says nothing about dispersion in either direction, so it is
+#' excluded from both and every draw is retained. The number of observations
+#' excluded is reported. Where instead the response differs from a fitted value
+#' of zero variance, the model has assigned zero variance to a value it did not
+#' predict; the Pearson residual is infinite, the statistic is reported as
+#' \code{Inf}, and a warning names the observations. An empty vector is
+#' returned only where every observation has zero variance and is reproduced
+#' exactly, in which case there are no residuals to compare.
+#'
 #' \bold{A beta-binomial fit does not address under-dispersion.}
 #' \code{beta_binomial} adds a variance component to the binomial, so it can
 #' represent a variance above the binomial's and not one below it. Where
@@ -69,6 +83,13 @@ dispersion <- function(model, summary = FALSE, seed = 10) {
   }
   chk_lgl(summary)
   chk_number(seed)
+  # Captured before `model` is replaced by the brmsfit. expand_nec() calls this
+  # once per equation, so an unnamed message on a model set says only that
+  # something somewhere was excluded, which is not enough to act on.
+  mod_name <- model$model
+  if (is.null(mod_name)) {
+    mod_name <- "fitted"
+  }
   formula <- model$bayesnecformula
   model <- model$fit
   mod_dat <- model.frame(formula, data = model$data)
@@ -92,10 +113,12 @@ dispersion <- function(model, summary = FALSE, seed = 10) {
     prd_out <- posterior_epred(model)
     set.seed(seed)
     ppd_out <- posterior_predict(model)
-    prd_sr <- matrix(0, nrow(prd_out), ncol(prd_out))
-    sim_sr <- matrix(0, nrow(prd_out), ncol(prd_out))
+    # The per-observation fitted variance is collected first and the residual
+    # arithmetic done in one place, because an observation with zero variance
+    # has to be handled differently depending on whether the response agrees
+    # with the fitted value. See pearson_dispersion() and #298.
+    var_out <- matrix(0, nrow(prd_out), ncol(prd_out))
     for (i in seq_len(nrow(prd_out))) {
-      prd_y <- prd_out[i, ]
       prd_mu <- fam_fcts$linkinv(lpd_out[i, ])
       prd_var_y <- fam_fcts$variance(prd_mu)
       if (fam == "binomial") {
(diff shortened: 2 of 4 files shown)


beckyfisher:
## Independent review

Reviewed on the branch at `b1fa4041`, R 4.6.1, brms 2.23.0, WSL Debian, 2026-09-09.
`tests/testthat/test-dispersion.R` was run with `NOT_CRAN=true`: 25 pass, 0 fail, 0
warn. The end-to-end fixture was refitted at seeds 298, 1, 2 and 7. At seed 298 it
reproduces the reported statistic exactly (0.969, 0.304 to 2.476) and the fitted
mean underflows at 6 of the 15 observations under all four seeds, so the test is
not seed-fragile. `pearson_dispersion()` was also exercised directly on
constructed matrices covering the cases the test file does not reach.

## Assessment

The statistical reasoning is correct and the refactor into `pearson_dispersion()`
is the right structure. Two findings below should be settled before merge: the
message describes a fixed exclusion set where the exclusion is per draw, which the
pull request's own fixture contradicts; and `dispersion()` prints "The fitted
fitted mean" whenever the equation name is absent. The remainder are accuracy
improvements to code and documentation that already state an intent they do not
quite meet. Nothing found changes the value the function returns for the case the
pull request is about.

## The derivation behind the exclusion

Excluding a zero-variance observation whose response equals the fitted value from
both sums is not a judgement call, and the justification is stronger than the one
recorded. As `mu` tends to zero with `obs = 0`, the observed Pearson term tends to
`mu n`, which tends to zero; the simulated term tends to zero as well, because a
predictive distribution with vanishing variance returns its mean with probability
approaching one. Excluding a term is arithmetically identical to contributing zero
to both sums, so the exclusion **is** the limit rather than an approximation to
it, and it introduces no bias in either direction. Recording that in the roxygen
would convert "the observation says nothing about dispersion" from an assertion
into a derivation.

The ratio is also self-normalising: under the null each retained observation
contributes approximately 1 to the denominator, so the null value stays at 1
whatever the size of the retained set. Calibration is preserved when the retained
set varies. `Inf` is likewise the correct limit for the disagreeing case, and the
placement argument for warning rather than stopping follows the reasoning already
recorded in `R/bnec.R`.

## A fixed exclusion set is reported where the exclusion is per draw

`degenerate` is a draws-by-observations matrix and the exclusion is elementwise,
but the message at `R/dispersion.R:258-265` counts observations degenerate in any
draw and reports "zero variance at N of M observations ... are excluded from the
statistic". Measured on the fixture in this pull request, seed 298, 400 post
warm-up draws: 229 draws exclude nothing, 60 exclude three, 111 exclude six, and
no observation is excluded in every draw. The user is told that six of fifteen
observations were dropped when the majority of draws dropped none.

The per-draw implementation is the correct choice and should be kept. The
alternative of fixing membership across the posterior, by dropping the six columns
everywhere, was measured on the same fit: the summary changes little (median 0.975
against 0.969) but individual draws differ by up to 74 per cent in relative terms
and 1.22 in absolute terms, because in the 229 draws where the curve is shallower
those observations produce genuine residuals that a fixed set would discard. The
defect is in the reporting alone. Reporting the draw count, in the form
`nsec_off_curve()` already uses, describes what the code does; a count of
observations implies a fixed set.

## Doubled word in the message text

`R/dispersion.R:89-92` sets `mod_name` to `"fitted"` where `model$model` is `NULL`
and then passes it as `model_name`, so the `is.null(model_name)` branch at line 221
is unreachable and `who` becomes `paste0("The fitted ", "fitted", " mean")`.
Reproduced:

```
The fitted fitted mean has zero variance at 1 of 3 observations (3) which the
model reproduces exactly...
```

Passing `model$model` through unchanged and letting `pearson_dispersion()` supply
its own fallback resolves it. The `"fitted"` default is needed only for the
message inside `dispersion()` at line 148, where it reads correctly.

## Negative variance and the unusable guard

The comment at lines 233-235 states the intent: a variance that no accepted family
should produce is excluded and reported "rather than left to disappear into the
na.rm below". `!is.finite(var_out)` does not catch a negative variance. `sqrt()`
then produces `NaN` with base R's unattributed "NaNs produced" warning, `na.rm =
TRUE` drops the term, and `n_used` still counts the observation. Reproduced on a
three-observation input with `var = -1`: the returned value equals the
two-observation statistic while `n_used` reports 3. This is not reachable for
`binomial` or `poisson`, whose variance functions are non-negative over the range
the likelihood admits, so the severity is low, but `!is.finite(var_out) | var_out
< 0` closes it and matches the stated intent.

## Draws returned as NA

The cited precedent, #39 as implemented in `nsec_off_curve()`, names the number of
excluded draws in a message. Where a subset of draws is fully degenerate,
`estimates_summary()` and `mean(disp > 1, na.rm = TRUE)` condition on the
remainder in silence. Demonstrated on a two-draw input whose first draw is fully
degenerate: the returned vector is `c(NA, 0.189)` and the summary reports only the
second. Following the cited precedent means naming that count.

Related: `summary = FALSE` can now return a vector containing `NA` or `Inf`, which
is a change to the exported return contract. `@return` in `man/dispersion.Rd`
still reads "an n-long vector containing the dispersion metric" and mentions
neither.

## The infinity and its cause

"A property of the fit rather than a numerical limit" appears in the roxygen, in
`pearson_dispersion()`'s details, and in `NEWS.md`. A variance of exactly zero is
produced by underflow, and in exact arithmetic the Pearson residual at such an
observation is very large but finite, so the infinity is the numerical limit. The
misfit it reveals is real, and that is the claim the sentence should make.

A stronger statement is available. For a converged fit the disagreeing case should
be close to unreachable, because Stan evaluates the same `binomial_lpmf` in the
same double precision and rejects any draw with `mu = 0` and `y > 0`. If that
warning fires it is more likely to indicate a scale mismatch between `prd_out` and
`obs_y` than a failure of curve shape, and saying so would make the warning
actionable.

## The equality predicate and the family list

`disagreeing <- degenerate & obs_mat != prd_out` compares an integer response with
a floating-point fitted value. That is safe for `binomial` and `poisson`, because
their variance functions equal the mean in the underflow regime: `1 - mu` is
exactly `1` for any `mu` below machine epsilon, so `mu (1 - mu) trials` and
`mu trials` are the same product and underflow together. Verified for `mu` from
1e-320 to 1e-324.

It is not safe for a family whose variance underflows before the mean. At
`mu = 1e-170` a Gamma-style `mu^2 / shape` gives `var == 0` with
`prd_out == 1e-170`, so a response of zero would be classified as disagreeing and
the whole statistic reported as `Inf` with a warning naming a severe misfit. This
belongs beside the existing #136 comment at lines 130-138, which already addresses
whoever widens `allowed_fams`.

<details>
<summary>Smaller points and test coverage</summary>

- `disp[n_used == 0 & rowSums(disagreeing) == 0] <- NA_real_` at line 250 is
  redundant. That expression is already `NaN` from `0/0`, and `is.na(NaN)` is
  `TRUE`, so every downstream test treats it as missing already. `n_used` exists
  only to serve this line.
- `sim_sr[unusable | degenerate] <- NA` at line 241 has no effect for the
  `degenerate` half. A zero-variance predictive distribution is a point mass, so
  `ppd_out == prd_out` and the term is already `0/0`. Only the `unusable` half
  changes anything, where the term would otherwise contribute a silent `0`. The
  code is correct; the surrounding comment implies it does more than it does.
- The `all(is.na(disp))` message at line 147 asserts "Every observation has a
  fitted variance of zero and is reproduced exactly", but the condition is also met
  where `disp` is `NaN` for any other reason.
- `R/dispersion.R:213`, "observations carry no information about dispersion", uses
  a verb the prose convention rules out. The roxygen at line 199 already has the
  form to use: "says nothing about dispersion in either direction".
- Test coverage: the all-degenerate message and `numeric()` return at lines
  146-151 are untested, as is the mixed case where some draws are `NA` and others
  are not, which is the one that changes the exported return contract.

</details>

https://claude.ai/code/session_01RqwRH6CBCHT1qf4uWiasoA

beckyfisher:
Worked through the review at `299bee2e`. Seven findings accepted, one accepted in part. `tests/testthat/test-dispersion.R` passes at 36 tests, 0 failures, 0 warnings, `NOT_CRAN=true`, R 4.6.1.

## The per-draw exclusion

Confirmed independently before changing anything. The fixture was refitted at seed 298 and the per-draw exclusion count tabulated: of 400 post warm-up draws, 229 exclude nothing, 60 exclude three observations and 111 exclude six, and no observation is excluded in every draw. The message did describe a fixed set that is not what the code does.

The elementwise implementation is kept, for the reason given in the review. The message now reports both counts: *"has zero variance at 6 of 15 observations (10, 11, 12, 13, 14, 15) which the model reproduces exactly, in 171 of 400 draws"*, and the reason the exclusion is made draw by draw is recorded at the code.

## Also accepted

- **"The fitted fitted mean".** `model$model` is now passed through unchanged and `pearson_dispersion()` supplies its own fallback; `dispersion()` keeps a separate label for its own message. Tested.
- **Negative variance.** The guard is now `!is.finite(var_out) | var_out < 0`. The value is set missing *before* the square root, so base R's unattributed "NaNs produced" warning is not raised at all rather than being raised alongside the informative one. The warning text says "negative or not finite". Tested, including that the returned value equals the statistic over the observations that remain.
- **Draws returned as NA.** Named in a message, in the form `nsec_off_curve()` uses. Tested on a two-draw input whose first draw is fully degenerate.
- **The infinity and its cause.** "A property of the fit rather than a numerical limit" was wrong and is removed from the roxygen, the helper's details and `NEWS.md`. The infinity is the underflow; the misfit it reports is real. The warning now also states what to check first, since for a converged fit the likelihood rejects a draw assigning zero probability to an observed value, which makes a scale mismatch between the response and the fitted mean the more likely cause.
- **The derivation.** Recorded in the roxygen: both terms tend to zero in the limit, so excluding them is arithmetically identical to contributing zero to each sum and introduces no bias, and the ratio is self-normalising, so the null value of 1 holds whatever the size of the retained set.
- **The equality predicate.** The assumption that the variance and the mean underflow together is recorded beside the #136 note for whoever widens `allowed_fams`, with the Gamma-style counter-example.
- **`@return`.** Records that the vector can contain `Inf` or `NA`.
- **Smaller points.** The `all(is.na(disp))` message states the condition tested rather than asserting a cause; the comment on the simulated-term assignment no longer implies it does more than it does; "observations carry no information" is replaced. Tests added for the empty-vector branch of `dispersion()` (by substituting the return of `pearson_dispersion()`, since a fit degenerate at every observation is not one anyone would produce) and for the mixed case. The end-to-end fixture is now fitted once and reused, which takes the file from 9m11s to 5m22s.

## Accepted in part

`disp[n_used == 0 & rowSums(disagreeing) == 0] <- NA_real_` is not removed. The point that `0/0` is already `NaN` and that `is.na(NaN)` is `TRUE` is correct, and nothing downstream depends on the substitution. It is kept because the vector is exported and `NA` reads as a missing value where `NaN` reads as an arithmetic accident. It is simplified to `disp[is.nan(disp)] <- NA_real_`, which drops `n_used` --- the variable that existed only to serve the longer form --- and does not change the returned value.

https://claude.ai/code/session_01CGpuM3TVYB7RtMn9sj7Gk7

beckyfisher:
The body has been updated to the state at `299bee2e`: the description of the exclusion now says it is elementwise and reports both counts, the `Inf` case states what the infinity is and what to check, the third case covers a negative variance as well as a non-finite one, and the test count and runtime are current. Nothing moved between the open sections and the fold.

https://claude.ai/code/session_01CGpuM3TVYB7RtMn9sj7Gk7

beckyfisher:
The family caveat from the review is withdrawn at `19abbe49`, both the note beside the #136 comment and the paragraph in `pearson_dispersion()`'s details.

`dispersion()` accepts `poisson` and `binomial` because those are the two families whose variance is fixed by the mean, which is what makes the ratio of observed to simulated Pearson sums a test of anything. A family that estimates a variance or dispersion parameter of its own has no equivalent null to compare against, so it is not a candidate for the statistic and the list is not one that would be widened to include it. A caveat about a family whose variance underflows before its mean therefore describes a change that would not be made, and it invites a future reader to guard against it.

The `#136` comment beside the rate branch is unchanged. It addresses the `negbinomial` count-scale variance, which is a real derivation someone might copy incorrectly if the rate handling were reused.

https://claude.ai/code/session_01CGpuM3TVYB7RtMn9sj7Gk7

Issue #305, opened 2026-09-10
The regularizing prior set is not a narrowed version of the uninformative one, and its bot anchor is biased for a count response


beckyfisher:
## What

The `"regularizing"` prior set does not mean the same thing in every family, and
on a count response its `bot` prior excludes the asymptote it is meant to
locate. Both are properties of the response-scaled `top` and `bot` entries,
which are the only ones `prior_type` selects between.

`prior_type` does not exist on `master` (2.1.3.1, the CRAN release), so
`"regularizing"` is entirely unreleased and can be changed without affecting any
published analysis.

## Why it matters

A user selects `prior_type = "regularizing"` when the data are sparse or noisy
and they want the prior to do more work. On a `poisson` or `negbinomial`
response fitted on the identity link they instead get a `bot` prior whose
central 95% interval excludes the true lower asymptote, so the parameter is
determined by the prior rather than by the data. The `"uninformative"` set,
which is the default, does not have this defect.

Separately, the word "regularizing" does not describe one thing. For a Gaussian
response it narrows the prior to 0.40 of its uninformative width; for a
`negbinomial` `top` it makes it 1.15 times *wider*; for a Beta response it
changes the width by 12 per cent and does not change what the prior is anchored
to at all. A user cannot reason about what selecting it will do without knowing
which branch their family takes.

## Evidence

Priors only; nothing was fitted. Responses were simulated from a `nec4param`
curve with a known `top` and `bot`, over three designs --- a series spaced evenly
from zero, a two-fold dilution series, and the `nassarius` contaminant A series
--- by four families on the identity link. Priors were read through
`get_priors()`, which runs the same `check_data()` and `define_prior()` sequence
`bnec()` runs, and each was truncated to its own bounds before being summarised.
`dev` at `aa2c7dbd`, R 4.6.1, brms 2.23.0, 2026-09-10.

**The `regularizing` `bot` prior excludes the true value on a count response.**
It is `gamma(5, 5 / (min(y) + min(y > 0) / 10))`, whose maximum density is at
about 0.8 times the smallest observation. For a continuous response that is
defensible. For a count the smallest observation sits well below the asymptote:
against a true `bot` of 5 the observed minimum was 1 to 3 across cells.

| family, identity link | cells with the true `bot` outside the central 95% |
|---|---|
| `poisson`, `regularizing` | 3 of 3 |
| `negbinomial`, `regularizing` | 3 of 3 |
[...]

PR #307, opened 2026-09-10
Define the regularizing prior set once and anchor it on the predictor

beckyfisher:
## What

`prior_type = "regularizing"` is defined once, as a location and a spread, and
every family's entry is derived from it. The location is the mean response at
the end of the concentration series where the parameter is the level of the
curve --- the control group for `top`, the highest concentrations for `bot` ---
rather than an extremum of the response. The `"uninformative"` set is unchanged.

`nec` and `ec50` are narrowed under `"regularizing"` as well, which they were
not. That is the follow-up comment on the issue.

## Why it matters

Selecting `"regularizing"` did not mean one thing. Measured over 420 `top` and
`bot` entries built from one simulated response each --- five designs, three
predictor transforms, seven families, two links --- the ratio of the
regularizing prior standard deviation to the uninformative one ran from 0.032 to
2.18, and the regularizing prior was the **wider** of the two in 17 of them. It
is now 0.4 in 385 of the 420 and above 1 in none.

More seriously, the released regularizing prior excluded the value it was built
to find. Over the 720 `top` and `bot` cells of the prior audit it placed the
true value outside the central 95% of its own prior in 65; the set here does so
in 1. The `"uninformative"` set does so in none, under either version.

For `nec` and `ec50`, `prior_type` was inert: the predictor-scaled prior was
identical under both sets, so a user selecting the narrower set got nothing for
the two parameters they most often select it for.

## Evidence

Priors only; nothing was fitted. Priors were read through `get_priors()`, which
runs the same `check_data()` and `define_prior()` sequence `bnec()` runs, over
the sweep archived at `notes/scripts/prior_audit.R`: five designs, three
predictor transforms, twelve families, two links, two prior types, two
equations, 4,320 parameter-prior pairs. `top` and `bot` are reported from the
`nec4param` cells and `nec`/`ec50` from the `ecx4param` cells, so each parameter
is assessed on the equation whose curve identifies it. Each prior is truncated
to its own bounds before being summarised. R 4.6.1, brms 2.23.0, 2026-09-10;
`dev` at `bd57f6f3` is the baseline.

**Coverage.** Cells placing the true value outside the central 95% of the
truncated prior, of 720 `top` and `bot` cells:

| set | dev | this branch |
|---|---|---|
| `uninformative` | 0 | 0 |
| `regularizing` | 65 | 1 |

Mean distance of the truncated prior CDF at the true value from 0.5 falls from
0.269 to 0.140. The one remaining exclusion is a `zero_inflated_beta` `bot` at
0.0217, just outside the band.

**Six anchors were measured, not assumed.** Each was run through the whole
sweep on the same simulated data and scored the same way:

| anchor for the location | cells of 720 outside the central 95% |
|---|---|
| `max(y)` / `min(y)` (released) | 65 |
| 95th / 5th percentile of the pooled response | 40 |
| 90th / 10th percentile | 51 |
| 75th / 25th percentile | 134 |
| extreme of the per-concentration group means | 4 |
| mean of the extreme tenth of the predictor, by rank | 27 |
| mean at the extreme concentration (adopted) | 11 |
| the same, with the spread floored at its standard error | **1** |

A quantile of the pooled response is a proxy for the level of one plateau whose
quality depends on what share of the design sits on that plateau: if a share
*f* of the observations are at the upper plateau, its own median sits at level
*1 - f/2* of the pooled response, which is 0.775 for a design with 45% of its
points above the threshold and 0.59 for one with 82%. The error runs in
opposite directions for different response types --- against a true `bot` of 5
the smallest count observed ran 1 to 3, and against a true `top` of 40 the 95th
percentile of a Negative Binomial response reached 72 --- so no single level
serves both parameters.

The per-concentration group-mean extreme scores best on these designs and is
**not** adopted, because it has no meaning without replication: on a continuous
predictor with one observation per value it reduces to the extremum. Checked
separately over five seeds, five families and both parameters on such a design,
it placed the true value outside the band in 4 of 10 `Gamma` cells, where the
rule adopted here failed none of the 50.

**The standard-error floor.** A mean of six control observations is not a
precise estimate of a plateau. The spread is `0.4` of the uninformative width or
the standard error of the location, whichever is larger, and never more than the
uninformative width. It binds in 35 of the 420 same-response entries, of which
26 reach the cap, on the responses whose anchor is least precise: a `bernoulli`
one, where a single observation states only whether one individual responded,
and an over-dispersed count.

The two halves of the rule are not measured on the same thing, and it is worth
being clear about that. The location is read at one end of the series; the
spread is inherited from the `"uninformative"` entry, which is built from a
quantile of the pooled response. Deriving the spread from the same six
observations as the location would make this a different prior rather than a
narrowing of the released one, and would put the width at the mercy of one dose
group. The consequence of inheriting it is that where the `"uninformative"`
entry is itself poorly scaled, so is this one, and the cap can then hold the
spread below the standard error of the location.

**Two designs the archived sweep does not contain were measured separately.**
The sweep's true `bot` is never at or near zero, and its hurdle cell simulates a
*flat* survival rate, so neither complete effect at the highest concentration
nor a declining survival series appears in it. Both were run: ten seeds, three
designs, five families, 420 cells, archived at
`notes/scripts/prior_hard_cases.R`.

| | `dev` regularizing | this branch | uninformative |
|---|---|---|---|
| true value outside the central 95% | 31 | 12 | 127 |
| density at the truth below 0.15 of the maximum | 60 | **1** | 61 |

The second row is reported because a beta prior whose maximum density is at
0.014 has nearly all its mass above it, so the interval measure penalises a
well-placed prior on a bounded parameter near its boundary. Both are given here
and in `NEWS.md`.

**All 12 of those cells are sampling error**, not a property of the rule. Their
prior modes are the observed group proportions: 0.167 against a true survival of
0.014, and 0.833 against 0.99. Six replicates resolve 0.014 only to the nearest
sixth, and at least one individual survives 8.1% of the time, which over 60
second-block lower-asymptote cells predicts 4.9 such draws against the 7
observed. The twelfth is the same thing on a mu block whose highest surviving
concentration held two survivors, 0.420 and 0.514, giving a location of 0.467
against a true 0.6. That is the one respect in which reading the response is a
liability rather than an improvement: the `"uninformative"` entry for those
families is a constant and is unaffected.

Three properties of the anchor come from those designs, each a fix made in
response to review rather than something the sweep found.

- A zero at the highest concentration is the endpoint responding and is kept in
  the average; a zero at the control is a structural one and is excluded.
- Where every observation at the highest concentration is zero the location is a
  tenth of the smallest positive observation; where the *control* is entirely
  zero it is the extreme quantile of the positive part, since a detection floor
  is the wrong location for `top` and applying it there was the #210 collapse
  reintroduced.
- The subset never extends past a fifth of the concentrations tested, which
  stops the second block of a hurdle fit --- primed from one survival proportion
  per concentration --- from averaging half the design. A design with few
  concentrations and no replication presents the same input and is resolved the
  same way, so its subset is one observation; a range-based limit resolves that
  case better and the hurdle block worse (6 density failures against 1). What
  the single-observation subset gives up is that the standard-error stand-in
  widens the entry towards the uninformative one, so the set relocates the prior
  rather than narrowing it: measured ratios of 0.55 and 1.00 on an unreplicated
  eight-concentration Gamma series.

**`nec` and `ec50` cannot take the same factor.** Their width is not a free
choice: `predictor_prior()` sets it to the smallest width whose central 95%
interval still reaches the farthest concentration tested, which is what #302
exists to guarantee. Narrowing by each factor, over the 30 design by transform
by parameter cells that exhaust the prior:

| factor | cells outside the central 95% |
|---|---|
| 1.00 (uninformative) | 0 |
| 0.84 (adopted) | 0 |
| 0.80 | 0 |
| 0.75 | 2 |
| 0.40 | 8 |

`0.84` is `qnorm(0.975) / qnorm(0.99)`: the central 95% interval reaches the
farthest concentration under `"uninformative"` and the central 98% interval
does under `"regularizing"`. Over the whole sweep the true threshold falls
inside the central 95% of the prior in every one of the 360 `nec` and 360 `ec50`
cells of each prior type, with the truncated CDF at the true value running 0.42
to 0.969 under `"regularizing"` against 0.474 to 0.953 under `"uninformative"`.

**One measurement in the issue does not reproduce, and its own table is
affected.** `notes/scripts/prior_audit.R` draws a new response inside its
`prior_type` loop, so a ratio taken across the two types there compares priors
built from different draws. The table of prior standard deviation ratios in the
issue body comes from that comparison. Measured on one response per cell the
direction of every finding holds and the numbers differ: the issue reports 0.40,
0.87, 0.83 and 1.15 for the `gaussian`, `Gamma`, `poisson` and `negbinomial`
`top` against 0.40, 0.90, 0.84 and 1.48 here, and 0.40, 0.34, 0.10 and 0.16 for
`bot` against 0.40, 0.26, 0.11 and 0.10. The coverage results are unaffected,
since each prior there is scored against its own data.

## What a user sees

`prior_type = "regularizing"` on a Negative Binomial response simulated from a
`nec4param` curve with a true `top` of 40 and `bot` of 5:

| parameter | dev | this branch |
|---|---|---|
| `top` | `gamma(5, 0.0416667)` | `gamma(17.0943, 0.442962)` |
| `bot` | `gamma(5, 4.54545)` | `gamma(7.76467, 1.22994)` |
| `nec` | `lognormal(1.70060, 0.867668)` | `lognormal(1.70060, 0.731017)` |

(The `0.84` is applied to the standard deviation of the normal prior on the log
of the concentration, not to the standard deviation of the lognormal written out
on the concentration scale, which is not a linear function of it.)

The released `bot` prior has its maximum density at 0.88 against a true `bot` of
5; the one here has it at 5.5. The released `top` prior peaks at 96 against a
true `top` of 40; the one here peaks at 36.3. The `nec` prior keeps its location
and takes 0.84 of its standard deviation, where before it did not change at all.
The `"uninformative"` priors are identical on this branch and on `dev`, byte
for byte. They are not identical to CRAN's, for two reasons that predate this
change and neither of which #305 touches: #302 and PR #304 replaced the `nec`
and `ec50` entry, and #210 and #232 put `positive_scale()` in place of the raw
quantile on the gamma branch of `top` and `bot`. On the series above CRAN gives
`gamma(5, 0.4)` for `nec`; on a `poisson` response with 22 zeros of 66 it gives
`gamma(2, 100)` for `bot` against `gamma(2, 0.2849)` here, the first being the
collapse #210 exists to remove.

<details>

### The change

All in `R/define_prior.R`.

- `regularizing_factor` (0.4) and `regularizing_predictor_factor`
  (`qnorm(0.975) / qnorm(0.99)`), each documented with the measurement behind
  it. The statement of what the set *is* lives above `regularizing_factor` and
  nowhere else.
- `gamma_from_mode_sd()` and `beta_from_mode_sd()` solve a branch's parameters
  from a mode and a standard deviation. The gamma has a closed form; the beta is
  found by root-finding on the concentration `a + b`, which is unique because
  the standard deviation falls monotonically in it from `1/sqrt(12)` at the
  uniform. A request at or above that limit returns the uniform.
- `regularizing_location()` returns the location and its standard error.
- `regularizing_entry()` applies the contract to one branch.
- `predictor_prior()` gains `prior_type`.
- `define_group_prior()`'s narrowing changes from one half to
  `regularizing_factor`, so one number states the whole set. It was two before,
  chosen as "the same order as" a narrowing that ran from 2.5 to 1.3 by family.
  Its `ogl` log-branch cap was a constant 1 and now scales with the prior type,
  so selecting the narrower set is not inert where the cap binds.

### Specifying the mode rather than the mean

A normal is specified by a parameter that is also its mode, so a gamma specified
by its mean peaks somewhere else: `gamma(2, 2/q75)` has its maximum density at
`q75/2` while `normal(q90, 2.5 sd)` peaks at `q90`. Stating the contract on the
mode is what makes one rule mean the same thing on three branches. It is the
same ambiguity that produced #273 and #302 on the predictor side.

### Edge cases

- A response with no spread: the entry falls back rather than dividing by zero,
  and the degenerate predictor fallback of `sigma = 1` is **not** narrowed, for
  the reason `define_group_prior()` already leaves its own fallback alone.
- Zero-inflation: the location is the mean of the *positive* observations on the
  zero-bounded branch, for the reason `positive_scale()` records --- `top` is
  the mean of the count process, not of the mixture. On the other branches
  `response_link_scale()` has already moved a zero onto the link scale.
- No replication: the rule extends to the next distinct concentration until at
  least a twentieth of the observations, and no fewer than three, are included.
- The `beta` entries are built only for the four unit-interval families, and the
  gamma entries only for the three positive ones, so #229's all-negative
  gaussian response still reaches neither.

### Tests

Seventeen tests added or rewritten in `tests/testthat/test-define_prior.R`,
covering the two solvers and both of their boundary branches, the ratio bounds
in all three branches, the mode placement (pinned to a group mean computed in
the test file, not read back from the function under test),
`regularizing_location()` with and without replication, the count `bot` that
#305 reports against the released anchor computed alongside it, the beta branch
reading the response, the predictor narrowing and its degenerate fallback, and
one test per defect found in review: complete effect at the highest
concentration, the group mean against the mean of its survivors, structural
zeros at the control, the hurdle second block, a group with no spread of its
own, and the `ogl` cap. One test asserts that each `"uninformative"` entry is
still what Fisher et al. (2024) state, reconstructed from the published rule
rather than pinned as a literal, so the claim that nothing released changes is
self-enforcing.

Four existing tests were updated: the two that pinned `beta(5, 1)` and the
extreme-quantile strings, the two group-level ones that pinned one half, and the
`nec` invariance test, which now asserts invariance across family and link and
tests `prior_type` separately.

### The vignette is stale

`vignettes/example3.Rmd.orig` is rewritten: the "Selecting the default prior
set" section states the contract, and the demonstration changes from
`nec3param` to `nec4param` so both response-scaled parameters are present ---
`nec3param` has no `bot`, and `bot` is where the sets differ most. A
`get_priors()` chunk shows the count branch without fitting anything, which is
where the released set failed outright and where the published demonstration
showed nothing.

`vignettes/example3.Rmd` is **not** rebuilt. Per
`notes/implementation/00_protocol.md` precompiling is not a merge gate; the
rendered vignette needs the `precompile-vignettes` workflow before release.

### Reproducing

`notes/scripts/prior_audit.R`, which is updated on this branch so that it runs:
its prior parser gained a `lognormal` case, without which every `nec` and
`ec50` cell returns `NA` and the run stops on the first of them; its output path
is relative and overridable; and it now prints the coverage tables the issue and
this body cite, with a note in the source that a ratio taken across prior types
from its output compares priors built from different draws. The anchor
comparison was run by the same script with the location rule varied.
`notes/scripts/prior_hard_cases.R` is the second sweep.

</details>

https://claude.ai/code/session_011ftghup233K2iaTrveZ3Q3


beckyfisher:
## Review findings that were real, and what they changed

Four rounds of independent review. Eight were defects in the code and are fixed;
the rest were inaccurate claims or missing evidence and are corrected. Every one
was found on a design the archived prior sweep does not contain, which is the
single most useful thing to take from this.

### Defects in the code

| what was wrong | what it did | fix |
|---|---|---|
| the `bot` anchor filtered its subset to positive observations | under complete effect at the highest concentration the prior peaked at 10.8 against a true `bot` of 0 | a zero at that end is the endpoint responding and is kept; only the `top` end filters |
| the same filter could leave one observation | with five zeros and one survivor the location was the survivor, 6x the group mean, and its standard error was 0 | as above |
| the detection-floor fallback was not conditioned on the end of the series | a zero-inflated control put `top` at 0.1 against a true 40, which is the #210 collapse at the other end | `top` falls back to the extreme quantile of the positive part |
| the subset could span half a hurdle fit's concentrations | that block is primed from one survival proportion per concentration, so `hubot` peaked at 0.34 against a true 0.014 | the subset stops at a fifth of the concentrations |
| `sd()` is zero on a homogeneous group | the standard-error floor was inert on a binary control where every individual survived, which is the modal case | the spread of the whole response stands in |
| `define_group_prior()`'s `ogl` log-branch cap was a constant | both prior types returned `student_t(3, 0, 1)` wherever it bound, so `prior_type` was inert there | the cap scales with the prior type, as #294 made the parameter-level one do |

Two more were mine to answer rather than to fix. The width cap can hold the
spread below the standard error of the location, which is the one case where the
floor does not do what it is for; that is inherited from the `"uninformative"`
entry and is now stated rather than implied. And a design with fewer than ten
concentrations and no replication presents the same input as a hurdle second
block, so no rule reading only the data separates them; the choice is made for
the hurdle block and what the other gives up is written down.

### Claims that were wrong

- The explanation offered for twelve hard-case cells. Their prior modes are the
  observed group proportions, not the true values: six replicates resolve a
  survival of 0.014 only to the nearest sixth. Sampling error, not a scoring
  artefact.
- The equations that return `top` exactly at a predictor of zero are thirteen of
  23, not fifteen; `nechormepwr01` was missing from the excess-term list; the
  shallow-curve range omitted `ecxwb1p3` and so was 17.6 to 29.7 rather than
  20.4 to 29.7; and how close the other ten come is a property of the curve
  rather than of the equation.
- The `bot` anchor's justification holds for a plain count response and not
  under zero-inflation, where it inherits the zero-inflation share as a downward
  bias. Measured: 3.92, 3.27 and 2.39 against a true `bot` of 5.

### Evidence that was missing

- `notes/scripts/prior_audit.R`, cited by the roxygen and `NEWS.md` as the
  evidence, could not be run: its parser had no `lognormal` case, so every `nec`
  and `ec50` cell returned `NA` and the run stopped on the first. It now runs
  and prints the tables quoted here.
- The measurement behind the round-2 fixes existed only in this thread. It is
  archived at `notes/scripts/prior_hard_cases.R`.
- Nine tests asserted the implementation against itself or left a branch
  unreachable. The prior's mode is now pinned to a group mean computed in the
  test file; one test reconstructs each `"uninformative"` entry from the
  published rule rather than pinning a literal, so the claim that nothing
  released changes is self-enforcing.

<details>
<summary>One suggestion declined</summary>

`define_hurdle_prior()` knows which block it is building, so it could pass a
flag and let the ordinary path use a range-based extension limit instead of a
count-based one. Declined: it puts an argument on `define_prior()`, which is
also the ordinary entry point, to change behaviour only on a design with fewer
than ten concentrations and no replication, where the current result is safe
rather than wrong --- the entry is relocated and widened towards the
uninformative one, which is the right answer for an anchor of one observation.
Worth its own issue if the design turns out to be common.

Two smaller items were noted and not acted on: `check_data()` refuses a constant
predictor by dying on an empty subset rather than by a message that names the
cause, and the archived sweep redraws the response inside its `prior_type` loop
so a ratio taken across the two types from its output compares different draws.
The first is pre-existing and out of scope; the second is now recorded in
`notes/prior_audit.md`.

</details>

beckyfisher:
## Correction: the `"uninformative"` set is not the set on CRAN

The roxygen above `regularizing_factor` and `vignette("example3")` both
described the `"uninformative"` entries as the ones on CRAN. That is wrong, and
it has been corrected in `1fdfbe58` rather than left for a reader to trip over.

The load-bearing claim of this PR is unaffected: `prior_type` does not exist on
`master`, so the whole `"regularizing"` set is unreleased, and #305 does not
touch the `"uninformative"` entries. They are byte-identical on this branch and
on `dev`. But two earlier changes on `dev` already moved them away from the
released ones:

- **#302, PR #304** replaced the `nec` and `ec50` entry outright. On a series of
  `0, 1, ... 10`: `gamma(5, 0.4)` on CRAN, `lognormal(1.70060, 0.867668)` here.
- **#210, #232** put `positive_scale()` in place of the raw quantile on the
  `Gamma` branch of `top` and `bot`. On a response with no zeros the two agree
  exactly --- both give `gamma(2, 0.0529801324503311)` for `top` on the fixture
  below. On a `poisson` response with 22 zeros of 66, CRAN gives
  `gamma(2, 100)` for `bot` against `gamma(2, 0.284900284900285)` here; the
  first is the collapse #210 exists to remove.

Measured directly against a `master` worktree at `6f6bd25d`, `nec4param`,
`poisson(link = "identity")`, eleven concentrations by six replicates, R 4.6.1.

This does not change the test that asserts the `"uninformative"` entries are
what Fisher et al. (2024) state: it reconstructs them from the published rule on
a response it first asserts has no zeros, which is the case the guard is
documented to leave untouched.

Issue #296, opened 2026-09-08
Redesign the negative-sgr simulation study around the model-averaged workflow


beckyfisher:
## What

Replace the `negative-sgr` simulation study rather than patch it. Phases 5 and
10 are retired; a single redesigned study answers the questions
`vignette("example7")` actually needs, and the vignette is rewritten against it
alone.

Supersedes the original text of this issue, which proposed re-running the
existing sweep over the current candidate set. That is subsumed here.

## Why it matters

Three problems in the existing design limit what the vignette can claim, and
they are cheaper to fix together than separately.

**The emphasis is inverted.** Phase 5 fixed `nec4param` for every arm and tested
model averaging afterwards on three cells. Fitting a candidate set is what
`bayesnec` recommends and what people do, so it should be the study, not an
appendix. Fixing the equation also meant fitting the generating equation to data
that had been altered, so part of the measured bias follows from the design
rather than from the convention under test.

**Two arms no longer correspond to a decision anyone makes.** Old B2 and B3
fixed the lower asymptote at zero through a `constant(0)` prior. Since 2.2.0 the
zero-asymptote equations are retained under a Gaussian family, so that is one
candidate among fourteen and the weights decide it. Removing those arms removes
the two-by-two design and the shared-prior machinery with it.

**A common practice was never tested.** Deleting the negative rows outright is
distinct from replacing them with zero and from truncating the concentration
series, and no arm covered it.

Alongside those, the study generated a residual scale rising 8.1-fold along the
curve and fitted a single `sigma` everywhere, so reported coverage confounded
each convention with a dispersion misspecification shared by all of them.

## Evidence

- `analysis/phase10_metrics.csv` and `analysis/phase10_vignette_numbers.txt` at
  `0181d66e` hold the eight-equation model-averaged results.
- `R/arms.R` and `R/data_prep.R` define old arm D as predictor truncation at an
  estimated zero crossing (`keep <- dat$x < crossing`), not deletion of negative
  values.
- The dispersion check is in `bayesnec:notes/disp_sensitivity_example7.md`:
  scenario 8, 50 datasets, six approaches fitted twice each, paired. Arm B2's
  ErC50 bias moves from -12.7% to -1.2% under `disp("loglinear")`.
- Cost basis: `analysis/phase10_gate2_pilot.R` measured 31.5 worker-minutes per
  iteration for five arms over eight equations, i.e. 0.79 worker-minutes per
  model fit.

## The design

[...]

Issue #283, opened 2026-09-06
example7: demonstrate a dispersion sub-model, and re-run the precision sweep under disp("loglinear")


beckyfisher:
## What

Two things #193 scoped for `example7` that did not land in PR #243, split out so
that PR can close its issue.

1. **A dispersion sub-model as a vignette section.** #193 called this "the
   payoff": `example7` fits growth-rate data where the variance is not constant
   in the mean, and `disp()` is the package feature that addresses it, but the
   vignette does not demonstrate it.

2. **Re-run the precision sweep under `disp("loglinear")`.** The sweep in the
   vignette was run with a pooled `sigma`.

## Why it matters

The bounded check recorded in `notes/disp_sensitivity_example7.md` shows that
**arm B2's headline numbers are largely a pooled-`sigma` artefact**. That is one
scenario at n = 50, so it is not enough to re-score the study, but it is enough
to say that the vignette's claims about B2 are conditioned on a variance
assumption the vignette does not state and does not test. Either the sweep is
re-run under a dispersion sub-model, or the claim is qualified.

## Evidence

`notes/disp_sensitivity_example7.md`, on the branch of PR #243. One scenario,
n = 50, bounded response.

<details>
<summary>Implementation detail</summary>

Ordering: this is downstream of PR #243 merging, and downstream of the ECx
redefinition in #281, which changes every ECx `example7` reports and therefore
requires a re-render regardless. Do not start it before both have landed.

The re-run is the expensive half: `example7` alone takes about 137 minutes
locally and about 2 h 56 m in CI, for 24 `alga` fits, and the sweep multiplies
that. Read PR #243's handoff note before starting a precompile -- it records two
traps that cost a previous session time, the stale installed library and the
figure paths.

The third item PR #243 flagged, that `example7` is about 14,700 words and the
longest vignette in the package, is a judgement for RF and is not part of this
issue. Adding a dispersion section makes it longer.

</details>

beckyfisher:
## Outcome of the re-run

The precision sweep has been re-run with a dispersion sub-model, on the same
cells and realisations as the study: five arms on the four precision cells, 100
realisations each, 2,000 units, all estimable. `floored`, `gamma` and `beta`
under `disp("loglinear")`, and the two bounded families under `disp("power")` as
well, since a power law in the fitted mean is the shape the Gamma's failure
calls for and choosing between the two forms in advance would have assumed the
answer.

A sub-model does not repair the conventions that impose the zero boundary
[...]

PR #308, opened 2026-09-10
Precompile the vignettes on the HPC in a container

### LINKED ISSUE #306: Precompile vignettes in a container on the HPC, not on a workstation
## What

Move `vignettes/precompile.R` off individual workstations and onto the HPC, running inside an Apptainer container that holds the toolchain but **not** `bayesnec`. The branch's own `bayesnec` is installed at job start from the checked-out source, so one image serves every branch and every vignette.

## Why it matters

#190 requires a full re-run of `precompile.R` across every vignette before the next CRAN release. On a workstation that is days of wall-clock, on the machine the person needs for everything else. #283 and #296 each queue a further `example7` re-run behind it, and `example7` is now the most expensive vignette in the repository.

It also removes a correctness trap. A container that pins `bayesnec` to a fixed commit cannot honestly precompile a vignette from a different branch. `example7`'s case studies are computed at build time precisely so that they track the version of the package they ship with, and its provenance section says so. Installing the package from the checked-out source at job start is what makes that statement true regardless of which branch is being built.

## Evidence

Measured on `example7` over 2026-09-09 and 2026-09-10. One vignette: six data-handling conventions across four datasets, model-averaged over the declining candidate set, 320 individual model fits and 304 distinct Stan programs.

| run | Stan cache | result |
|---|---|---|
| first | cold, all 304 programs compiled | 309 min end to end |
| second | warm, nothing recompiled | 147 min of fitting; still running at 4 h 31 m |

Compilation is roughly half the cost of a cold precompile and is entirely avoidable between runs. The cache is keyed by a hash of the Stan source, so it is reusable across branches and across vignettes wherever the data and priors are unchanged. A persistent cache on shared scratch is the single largest lever available here, larger than the choice of hardware.

The second run also exposed a cost that is not fitting at all: extracting `ecx()` and `nsec()` from 24 model-averaged fits has taken over two hours on its own, because each call solves across every posterior draw of thirteen equations. That is a property of the vignette rather than of the build, but it means an HPC job for `example7` should be sized for post-processing as well as sampling.

<details>

### Container

Base image pinned by digest. Install: cmdstan and `cmdstanr`, `brms`, and the packages the vignettes use directly — `ggplot2`, `dplyr`, `tidyr`, `knitr`, `rmarkdown`. Do **not** install `bayesnec`.

Pandoc is not required. Precompiling stops at `knitr::knit()`; rendering to HTML happens later, at package build or pkgdown time.

The existing definition file in the `negative-response-conventions` compendium is the starting point. It differs in two ways that matter: it pins `bayesnec` to a commit, and it has no `knitr`.

### Job

1. Check out the branch into a job directory.
2. `R CMD INSTALL --library=<job-local library>` from that source, and put the library first on `R_LIBS`.
3. Set `cmdstanr_write_stan_file_dir` to a persistent directory on shared scratch, shared by all precompile jobs.
4. Knit only the requested vignettes.
5. Copy back `vignettes/<name>.Rmd` and `vignettes/vignette-fig-*.png`, nothing else.

### Changes needed in this repository

`vignettes/precompile.R` knits every `.Rmd.orig` it finds. It needs to accept a vignette name, or a vector of them, so that one vignette can be rebuilt without rebuilding the rest and so that vignettes can be run as separate array elements rather than one serial job. Keep the existing all-vignettes behaviour as the default.

The existing failure check at the end of that script — which greps the knitted output for error markers, because knitr renders a failed chunk and carries on — must run inside the job and set the exit status, so that a failed precompile does not return a plausible-looking `.Rmd`.

### Open questions

**Whether unprivileged container build is permitted on the HPC — not verified.** I attempted this check and could not reach the login node; the workstation has no VPN interface up, so this is untested rather than answered. The check is, on a login node: whether the module provides a build subcommand at all, whether the invoking account has entries in `/etc/subuid` and `/etc/subgid`, and whether a `--fakeroot` build of a trivial definition file succeeds.

If it is not permitted, the fallback is to build the image elsewhere and copy it once. Under this design that is a much weaker objection than it first appears: the image changes only when a dependency changes, not when the package or the branch changes, so the copy is rare rather than per-run.

**Provenance of the committed output.** Precompiling in a container makes the committed `.Rmd` files depend on the image's R and cmdstan versions rather than on whichever workstation ran them. That is an improvement in reproducibility, but it means the image identity becomes part of what the output should record, and that rebuilding the image is a change that can alter committed vignette output. Decide where that identity is recorded before the first production run, not after.

### Definition of done

- A documented single command that precompiles one named vignette on the HPC and returns its `.Rmd` and figures.
- `example7` reproduced through it, with the non-fitting chunks byte-identical to a local run.
- Two consecutive runs demonstrating that the Stan cache is reused and that the second compiles nothing.
- `precompile.R` accepting a vignette selector, with the all-vignettes default unchanged and the error check setting the exit status.

Any change that would alter committed vignette output as a side effect — a different base image, a different cmdstan version, or a change to the default candidate set — needs to be raised before it is made, because the output is committed and reviewed.

</details>

beckyfisher:
## What

Precompiling the vignettes moves off a workstation and onto the AIMS HPC, inside
an Apptainer container that holds the toolchain but **not** `bayesnec`. The job
installs `bayesnec` from the checked-out source at start-up, so one image
precompiles a vignette from any branch. One command does it:

```sh
./hpc/build.sh                     # once, and again only when a dependency changes
./hpc/precompile-hpc.sh example3   # deploy, submit, wait, collect
```

`vignettes/precompile.R` now takes a vignette selector, keeps the all-vignettes
default, and every vignette is knitted with the `cmdstanr` backend.

## Why it matters

#190 requires a full re-run of `precompile.R` across every vignette before the
next release, which is days of wall clock on the machine the person needs for
everything else. #283 and #296 each queue a further `example7` re-run behind it.

It also removes a correctness trap. An image that pinned `bayesnec` to a commit
could not honestly precompile a vignette from a different branch, and `example7`
states that its case studies are computed at build time so that they track the
version of the package they ship with. Installing from the checked-out source is
what makes that statement true whichever branch is built.

The `cmdstanr` backend is what makes the Stan compile cache available.
`cmdstanr` names a compiled program after a hash of its Stan source, so a
program is reused only where the source is identical, and the cache is reusable
between runs, between vignettes and between branches. `rstan`, the `brms`
default and what example1--example6 used until now, has no equivalent cache that
survives a session.

**This changes the sampler that produces committed vignette output.** The next
precompile of example1--example6 is expected to change their numbers. That was
decided deliberately rather than discovered; #190 is where it becomes visible
across the set.

## Evidence

Measured on 2026-09-10 against the AIMS HPC, partition `cpuq`, one task, four
cores.

**The issue's open question is answered.** Unprivileged container build is not
permitted on the cluster. `singularity build --fakeroot` on a trivial definition
file returns

```
FATAL: could not use fakeroot: no mapping entry found in /etc/subuid for <account>
```

so a definition file with a `%post` section cannot be built there. `singularity
exec docker://…` does work, so the restriction is on building rather than on
images. The image is therefore built with apptainer 1.4.5 on the workstation and
copied once; a 1.4.5-built SIF was confirmed to run under the cluster's
singularity 3.6.3.

**Container and workstation agree where they can be compared.** `example2b` was
precompiled both ways and the two `.Rmd` files differ in no byte. What that
covers should be stated rather than implied: `example2b` runs no `bnec()` call,
so the comparison exercises R, `knitr` and `ggplot` text output and nothing of
the Stan toolchain, the source install or sampling. It is a control on the
container's R environment, not on its sampler.

The corresponding comparison for a vignette that does fit could not be made
against this workstation, and finding out why is the more useful result. A local
`example4` run disagreed with two container runs that agreed with each other,
because the workstation was loading `bayesnec` 2.1.3.7, built 2026-08-17, from
the user library -- twenty-six versions behind the `dev` the container installs
from source. This is the failure the design exists to prevent, observed rather
than argued.

**A `bnec()` fit does not reproduce under a fixed seed, and that is not caused
by this change.** `example3` was precompiled three times, twice on the same image
and once on its predecessor, all on the same node, and no two agreed. The
difference is confined to the `fixef()` summaries of `exmp_f` and `exmp_g` and
the three `check_priors` figures — but that is because those are the only
posterior-dependent outputs the vignette prints. Everything else it prints is a
prior specification or prose, so the 416 lines that agreed would have agreed
either way and are not evidence of anything.

What settles it is a measurement that needs no cluster. Two `bnec()` calls in one
R session, each preceded by `set.seed(333)`, `nec3param` on `nec_data` with
`Beta(link = "identity")`: `fixef()` differs by about 5e-5, and
`sum(.Random.seed)` after the two calls differs, so a fit consumes a number of
random draws that is not fixed by the seed. `check_priors()` called twice on one
saved fit gives byte-identical PNGs, so the plotting is deterministic and the
figures differ because the fits do.

This is a property of `bayesnec` rather than of the container — same session,
same machine, no container involved — and it is raised as #310 rather than
addressed here. It does mean the `example3` render committed here is one draw
among many, as every committed vignette render is, and that its three
`check_priors` figures will differ on any re-render without anything having
changed. The render committed is the one produced by the image
`hpc/image.lock` records.

**A failed precompile returns nothing.** A deliberate `stop()` was added to one
chunk and the vignette precompiled through the same command. `knitr` rendered the
error and carried on, the check in `precompile.R` caught it, and the job ended
`FAILED` with exit code 1. `out/example2b` was absent and `--fetch` reported
`no output staged for example2b -- the job failed, or has not finished`.

**The Stan cache is reused and the second run compiles nothing.** `example4` was
precompiled twice through the same command, onto a cache emptied beforehand. The
first run compiled 14 Stan programs; the second found the cache holding the same
28 files (14 programs and 14 executables) with the same names, sizes and
modification times, so nothing was written and nothing was recompiled. Elapsed
13 m 58 s and 13 m 06 s.

What that is worth depends on how much sampling a vignette does, and the
"roughly half the elapsed time" figure that #306 reports for `example7` is not
established by the numbers given there: the cold run has a total, 309 min, and
the warm one has 147 min of fitting and no total, having not finished when it
was last observed. What is established is that 304 programs were compiled cold
and none warm. The claim is stated that way here and in the scripts.

**A production re-render was done through it.** `example3` was rebuilt on `dev`
by `./hpc/precompile-hpc.sh example3` and the result is committed here: 11 m 25 s,
no errored chunks, three figures returned. The committed render predated the
predictor prior from #302, and the rebuilt one reports
`lognormal(-0.1327, 1.6829)` where the old one reported `gamma(5, 2.2831)`. That
re-render is the only change to committed vignette output in this pull request;
example1, example2 and example4 keep their existing renders, which are now older
than their sources, and #190 is where the set is brought up to date together.

### Definition of done

| item | state |
|---|---|
| a documented single command | done |
| two consecutive runs, second compiles nothing | done, on `example4` |
| `precompile.R` selector, all-vignettes default, error check sets exit status | done |
| `example7` reproduced through it, non-fitting chunks byte-identical to a local run | **not done** |

`example7` is not on this branch, and it is work in progress. `example3` was
re-rendered instead, which was wanted independently. The byte-identity comparison
was made on `example2b`, and its limits are stated above. The `example7` item
stays open; it is the natural first job for whoever runs #283 or #296.

### Two things the backend switch broke

Both were found by running it, and both are fixed here.

`open_progress = FALSE` is an `rstan` argument. Under `cmdstanr` `brms` raises
`unused argument (open_progress = ...)`, so **every** model failed and `bnec()`
reported only `None of the models fit successfully`. It appears in 20 `bnec()`
calls across example1--example4 and is removed. It was a no-op during a knit in
any case: it defaults to `interactive()`.

`.github/workflows/precompile-vignettes.yaml` runs this same script on a runner
that has neither `cmdstanr`, which is not on CRAN and not in `DESCRIPTION`, nor a
cmdstan installation. An unconditional `cmdstanr` default would have failed every
leg of it. That workflow now sets `BAYESNEC_BACKEND=rstan`. **The consequence is
that a vignette precompiled in CI and one precompiled on the HPC are sampled by
different back ends**, which is a real difference and is not resolved by this
pull request: resolving it means installing `cmdstanr` and cmdstan on the runner
and caching them. Raised here because it is a decision rather than an oversight.

<details>

### `vignettes/precompile.R`

A selector, with the all-vignettes default unchanged:

```sh
Rscript vignettes/precompile.R example3
BAYESNEC_VIGNETTES=example2,example3 Rscript vignettes/precompile.R
```

The environment variable exists because the routes that cannot pass arguments --
an interactive `source()`, a container `%runscript` -- are the ones this is for.
`commandArgs()` is read only when the session is not interactive, so a stray
`--args` from an IDE cannot silently narrow a precompile to one vignette. A name
may be given as `example3`, `example3.Rmd`, `example3.Rmd.orig` or with the
directory attached. An unknown name stops and lists what is available.

`options(brms.backend = ...)` defaults to `cmdstanr` and is overridable with
`BAYESNEC_BACKEND`. `cmdstanr_write_stan_file_dir` is set from
`BAYESNEC_STAN_CACHE` when that is given, and is what makes the compiled
programs persist; unset, they go to the session tempdir and nothing survives the
run.

Figure handling changed with the selector. The script copied every
`vignette-fig-*.png` in the working directory into `vignettes/` and then deleted
them all, which is right for the all-vignettes default but wrong for a partial
rebuild: it would copy a figure left behind by an interrupted run over a good
committed one, and delete figures belonging to vignettes it had not rebuilt. It
now records the figures present before knitting, by name, modification time and
size, and acts only on those that are new or whose contents changed.

The existing check -- grep the knitted output for `^#> Error`, because `knitr`
renders a failed chunk and carries on -- is unchanged in what it tests. What is
new is that it now sets the exit status of a job: under `Rscript` an uncaught
`stop()` exits non-zero, `set -e` in `run.precompile` turns that into a failed
job, and the outputs are staged only after it has passed.

### `hpc/`

| file | what it is |
|---|---|
| `bayesnec-precompile.def` | the image definition; base pinned by digest |
| `build.sh` | builds the image locally and writes `image.lock` |
| `image.lock` | the image's identity, committed |
| `precompile-hpc.sh` | deploy, submit, wait, collect |
| `run.precompile` | the SLURM array task: install, knit, stage |

`^hpc$` and `^.*\.sif$` are added to `.Rbuildignore`; `*.sif` to `.gitignore`.

**Image identity.** Pinning the base image by digest is not enough on its own:
`rocker/r-ver` configures `p3m.dev/cran/__linux__/noble/latest`, which is a
moving target, so the package snapshot is pinned to a date as well and the
derivation is asserted rather than written out. `build.sh` writes `hpc/image.lock`
from a manifest generated inside the image -- base digest, R and cmdstan
versions, the resolved repository, every package version -- and appends the
SHA-256 of the `.sif`. Before anything is fitted, `run.precompile` compares the
SHA-256 of the image it has been given against that lock and refuses to run on a
mismatch, so rebuilding the image cannot alter committed vignette output as a
side effect. `./hpc/build.sh --check` performs the same comparison locally, by
digest alone, and reads the manifest only to report what differs.

**Library isolation.** Apptainer bind-mounts `$HOME` by default, so
`R_LIBS_USER` otherwise resolves to the account's own library on the host and the
container reads the host's packages -- an image built with no `bayesnec` in it
reported `bayesnec` as available, because it was seeing the workstation's library
through the bound home directory. `run.precompile` sets both `R_LIBS` and
`R_LIBS_USER` to the job-local library and then asserts that `find.package("bayesnec")`
resolved there.

**Staging.** `out/<vignette>/` is cleared at the start of the job rather than
before a successful stage. Staging on success alone is not enough: a rerun that
failed would leave the previous run's output in place and `--fetch` would return
it as though it were this run's.

**One task at a time.** The array is submitted with `%1`. `cmdstanr` does not
lock the compile cache, and vignettes share Stan programs wherever a family and
equation coincide, so two concurrent tasks could run `make` on the same
executable path. Serialising affects wall clock only on the first, cold run.

**Not done here.** Pandoc is not installed: precompiling stops at
`knitr::knit()`, and rendering to HTML happens at package build or pkgdown time.
The image copy takes about 20 minutes over the VPN, and is skipped when the
cluster already holds a byte-identical image.

### What to raise rather than guess

Any change that would alter committed vignette output as a side effect -- a
different base image, a different cmdstan version, a different package snapshot
date, or a change to the default candidate set -- needs to be raised before it is
made. `hpc/image.lock` is what makes such a change visible in review.

</details>

Closes #306

https://claude.ai/code/session_01Hg4DT12BxPoJdZ5T4cWueU


beckyfisher:
## Review cycle: what was actually wrong

Three rounds with an independent reviewer, ending in agreement that this is
sound. Twenty-five findings were raised. These are the ones that were real
defects and are fixed; the rest were either accepted as claim corrections
(below) or judged out of scope and raised as their own issues.

### Would have broken something

**The precompile GitHub workflow.** `.github/workflows/precompile-vignettes.yaml`
runs `vignettes/precompile.R` on a runner that has neither `cmdstanr`, which is
not on CRAN and not in `DESCRIPTION`, nor a cmdstan installation. Defaulting the
script to the `cmdstanr` backend would have failed every leg of that matrix, with
`bnec()` reporting only `None of the models fit successfully`. The workflow now
sets `BAYESNEC_BACKEND=rstan`. Consequence recorded rather than hidden: a vignette
precompiled in CI and one precompiled on the HPC are now sampled by different
back ends.

**`--fetch` exited silently.** `ssh …; rc=$?` is a simple command, so `errexit`
ended the function before either branch could run. `./hpc/precompile-hpc.sh
--fetch <vignette>` on a job that had not finished — the commonest case there is
— printed nothing and exited 1. Introduced while fixing a different finding, and
caught on the round after.

**The Stan cache was shared across images.** `cmdstanr` names a cached program
after a hash of its Stan source and nothing else, so the executable beside it
records neither the cmdstan nor the compiler that built it. The job refused to
run against an image the branch does not record, and then handed that run
binaries built by a different one. The cache is now keyed on the base image
digest and the cmdstan version.

### Would have returned wrong output

**`--fetch` could return a previous run's vignette.** Staged output survives a
redeploy, and was cleared only once a task started. Deploying a second branch and
collecting before its job began returned the first branch's `.Rmd` and reported
success. Output is now stamped with the deployment that produced it and checked
on collection; `out/<vignette>/` is also cleared before anything that can fail.

**A dependency check that read an empty variable**, failing a real job on
`vignettes/.Rmd.orig`. And its first version was too strict in the other
direction, demanding every `Suggests` entry including `testthat`. It now asserts
`Depends`, `Imports`, what `precompile.R` loads, and what the vignette itself
attaches — read from its `library()` calls, because `bayesplot`, `posterior` and
`cmdstanr` appear in no field of `DESCRIPTION`.

**A second deployment while an array was queued** would have rewritten the source
tree and `vignettes.txt` underneath the tasks that had not started, so task *n*
would precompile whatever was then on line *n*. Refused.

### Would have made a failure hard to diagnose

- `hpc/image.lock` named 20 packages, not the 138 installed. It omitted
  everything `brms` pulls in, several of which change the Stan code it generates.
  Now taken from `installed.packages()`.
- `build.sh` truncated the committed lock before it could fail. Now written to a
  temporary file and moved. This was then confirmed by accident: a build failed on
  a truncated cmdstan download and the lock survived intact.
- The container read the host's R library and startup files through apptainer's
  bound `$HOME`. An image built with no `bayesnec` in it reported `bayesnec` as
  available.
- Vignette names were validated inside the job, so a typo cost a tree sync, a
  lock check, a 20-minute image copy and an install first. Now validated and
  normalised before any contact with the cluster.
- The three shell scripts were committed non-executable. `core.fileMode` is
  `false` on the 9p mount this was written on, which hid it.
- Two vignettes were sampling four chains one after another on a four-core
  allocation, because only three of the six set `mc.cores`.

### Claims that were wrong, and are corrected rather than defended

**rstan is not required by the cmdstanr path.** `brms` lists it under `Imports`,
so it was in every image already, and the failure that prompted the claim was
`open_progress` and nothing else.

**"Compilation is roughly half the elapsed time"** is not established by the
numbers on #306: the cold run has a total and the warm one does not. Measured
here, `example4` was 13 m 58 s cold and 13 m 06 s warm.

**"The pipeline is reproducible and two fits are not"** was the reverse of what
the evidence supported. `example3` prints posterior output for only two of its
eleven fits, so the 416 lines that agreed across three runs are prior
specifications and prose and would have agreed either way. Two `bnec()` calls in
one session under `set.seed(333)` give different `fixef()`, and leave R's random
stream in a different state; `check_priors()` twice on one saved fit is
byte-identical. So no fit reproduces, and the plotting is not the cause. #310.

### Raised, not fixed here

- #310 — a `bnec()` fit does not reproduce under a fixed seed. This is the
  load-bearing open question for #190: until it is resolved, no vignette render
  produced by this pipeline is reproducible either.
- #311 — twelve `bnec()` calls in the test suite pass `open_progress`, which
  fails under `cmdstanr`. Pre-existing.
- `g++` is not in the toolchain cache key: `%post` installs `build-essential`
  from the noble archive, which is neither digest- nor date-pinned, so two images
  with the same base digest could hold different compilers. Adding
  `g++ -dumpfullversion` to the manifest and the key would close it.
- `example5.Rmd.orig` attaches `cmdstanr` and calls `set_cmdstan_path("C:/cmdstan")`
  in evaluated chunks, so that leg of the CI matrix cannot pass on a runner
  whatever the backend. Pre-existing on `dev`.

beckyfisher:
The decision to precompile every vignette with `cmdstanr` rather than `rstan` is recorded on #313, together with the open question of whether to install `cmdstanr` on the Actions runner so that both routes sample identically.

That decision is separable from this pull request and is worth reading separately. #306 asked for the move to the cluster, and needed no backend change to get what it was after: `example7` already set `options(brms.backend = "cmdstanr")` in a chunk of its own. Extending it to every vignette is what makes the compiled-program cache general, and it is also what brought in the workflow change, the removal of `open_progress` from 20 calls, and the expectation that example1, example2 and example4 will change their numbers when next built.

beckyfisher:
The body has been edited, and so has #310's: an account name, the login node and a compute node name are replaced with placeholders. Nothing about the change moved, only those identifiers. The same thing has been taken out of the repository itself in 278f4da7 — cluster settings now live in `hpc/local.conf`, which is gitignored, with `hpc/local.conf.example` as the template.

That commit also makes a local image optional. It is needed to build one and to copy it across; once the cluster holds an image matching `hpc/image.lock`, `precompile-hpc.sh` verifies the remote one against the lock instead of refusing, so the day-to-day call needs nothing set beyond `HOST`.

Issue #309, opened 2026-09-10
Initial-value search: all chains must pass at once, and the acceptance threshold sits at the centre of the prior


beckyfisher:
## What

Two defects in `make_good_inits()`, taken together because they multiply.

**The structure.** The check requires all chains to pass **at the same time**, and re-draws every chain when any one fails. The per-chain acceptance rate is 8–31%, so requiring four at once makes the per-attempt rate 0.004–0.9%. Accepting chains individually returns the same initial values for two orders of magnitude fewer draws.

**The criterion.** The clause `max(x) < max(limits)` compares the initial curve's upper asymptote — an estimate of the **mean** control response — against the **largest single observation** in the dataset. Both are derived from the same response and on a declining curve both live in the control, so the threshold sits at the centre of the prior it is filtering and rejects about half of every equation's draws by construction.

## Why it matters

The search is a single-threaded R loop and, on a wide prior relative to the response range, it is most of the wall-clock of a model-averaged fit. Two equations of the default declining set cannot succeed within the 10,000-attempt cap at all: they exhaust it and fall back to Stan's own initialisation, so the budget is spent to reach the outcome that was available at the first attempt.

The two defects have different characters and that is why both belong here. Accepting chains individually is **not a trade-off**: the four draws are independent, so a chain accepted on its own is a draw from the same distribution as one accepted alongside three others. The set of accepted initial values is unchanged and only the number of proposals changes. Changing the criterion **is** a behaviour change — it alters which starting values are accepted — so it needs its own evidence. Fixing only the structure leaves a criterion that discards half of all proposals for a reason unrelated to whether the starting values are any good; fixing only the criterion leaves the fourth-power multiplier in place.

#266 raised the same symptom — 561 seconds for a single model — and the reasoning recorded in `inits_functions.R` addressed the size of the cap rather than either of these. The cap is not the problem and should stay where it is.

## Evidence

Measured on the `alga` `c_proliferum` contaminant A series (n = 85, 14 concentrations, response range -0.577 to 0.137), Gaussian identity link, default `uninformative` priors, package version 2.1.3.33. 400 single-chain draws per equation, each scored by `check_init_predictions()`.

| equation | per-chain pass | all four | expected attempts |
|---|---|---|---|
| `ecxll4` | 30.8% | 0.900% | 111 |
| `ecxwb1` | 27.2% | 0.547% | 183 |
| `ecxwb2` | 26.5% | 0.493% | 203 |
[...]

PR #312, opened 2026-09-10
Accept initial-value chains individually, and band the curve on the level means

beckyfisher:
## What

Two changes to the initial-value search, `make_good_inits()`.

**A chain is accepted on its own.** The search required all four chains to pass
at the same time and re-drew the complete set whenever any one failed. The four
draws are independent, so this changed nothing about which values are accepted
and raised the number of proposals to the fourth power of the per-chain rate.
Accepted chains are now kept and only the empty slots are re-drawn.

**The initial curve is tested against a band read from the level means**, rather
than against `range(y)`. The band is every mean response the design estimates —
the mean at each replicated predictor value, plus the two
`regularizing_location()` anchors that #307 adopted for the regularizing `top`
and `bot` priors — widened by four within-group standard deviations,
intersected with the support of the mean mapped through the link, and stopped
short of the boundary of that support.

`n_trials` is unchanged at 1e4.

## Why it matters

The search is a single-threaded R loop and is most of the wall clock of a
model-averaged fit on a wide prior. Measured over the fourteen equations of the
`decline` set, four chains, five seeds, the released 10,000-attempt cap in
force,
counting chain proposals drawn:

| design | released | changed | ratio |
|---|---|---|---|
| `alga` `c_proliferum` A | 148,397 | 237 | 626 |
| `alga` `r_salina` B | 210,180 | 414 | 508 |
| `nec_data` | 552 | 87 | 6 |
| small replicated | 236,632 | 2,805 | 84 |
| small unreplicated | 334,178 | 308 | 1,085 |

**Of the 340 searches, 71 exhausted the cap under the released rule and none
under the change.** A search that exhausts the cap falls through to Stan's own
initialisation, so that budget was spent to reach the outcome available at the
first attempt. On the `c_proliferum` series those were `neclin`, `ecxlin` and
`ecxsigm`.

The criterion is a separate matter and is not a speed fix. `range(y)` was wrong
on three counts, none of which the acceptance rate expresses. It compared the
initial curve's upper asymptote, an estimate of the **mean** control response,
against the **largest single observation**. It was tied to the prior it
filtered,
because both `max(y)` and the location of the `top` prior are read from the
control, so the threshold landed near the centre of the prior it was testing.
And it got **looser as replicates were added**, because an extremum drifts
outward with sample size, which is the wrong way round for a starting-value
check.

The consequence is not only that draws were rejected but that the accepted ones
were displaced. Initial values are a draw from the prior restricted to the
region
the criterion admits, and a threshold at the centre of the `top` prior puts
every
accepted `top` on one side of it.

## Evidence

Archived at `notes/scripts/init_search_audit.R`, which takes the name of a
measurement as an argument. The full test suite passes with 0 failures. R 4.6.1,
brms 2.23.0, rstan 2.32.7, 2026-09-10. Five
designs: the two packaged `alga` growth series, the packaged `nec_data`
unit-interval series, and the small replicated and unreplicated designs the
`n_trials` comment in `R/inits_functions.R` describes. Fourteen equations of the
`decline` set where the family admits them.

**Per-chain acceptance**, 2,000 single-chain draws per equation and design,
median over the equations of a design:

| design | released | band | ratio |
|---|---|---|---|
| `alga` `c_proliferum` A | 0.186 | 0.245 | 1.23 |
| `alga` `r_salina` B | 0.138 | 0.202 | 1.48 |
| `nec_data` | 0.682 | 0.683 | 1.00 |
| small replicated | 0.119 | 0.204 | 1.54 |
| small unreplicated | 0.103 | 0.219 | 1.89 |

The clause the issue identifies, `max(x) < max(limits)`, passed 46.2% to 61.4%
of
draws on the `c_proliferum` series and passes 55.6% to 76.5% under the band.

`nec_data` is the case where the two criteria very nearly coincide. Its band
reaches both boundaries of (0, 1) and is set back from each by a small fraction,
so on an unreplicated series it is close to `range(y)` itself. The criterion is
not looser everywhere; it is differently placed, and on that design the
difference is almost nothing.

**What each criterion keeps of the prior**, 6,000 draws per equation over the
three replicated designs, reported as the proportion of the prior retained and
the median of the accepted draws relative to the unfiltered prior, in units of
the prior's own standard deviation:

| parameter | kept, released | kept, band | shift, released | shift, band |
|---|---|---|---|---|
| `top` | 0.14 | 0.21 | −0.149 | **−0.031** |
| `bot` | 0.14 | 0.20 | −0.067 | −0.047 |
| `nec` | 0.07 | 0.12 | −0.077 | −0.078 |
| `ec50` | 0.17 | 0.25 | −0.050 | −0.054 |
| `beta` | 0.16 | 0.23 | −0.700 | −0.574 |

The median displacement of the accepted `top` falls from 0.149 prior standard
deviations below the prior median to 0.031. That is the defect the criterion
fixes: not that draws were rejected, which any criterion does, but that the
survivors were all on one side.

These are medians over 42 design and equation combinations, whose individual
`top` shifts run −0.429 to +0.208 under the band against −0.554 to +0.261 under
the released criterion. What is removed is the systematic displacement, not the
equation-to-equation spread: `neclin` and `ecxlin` remain displaced by −0.216 to
−0.429 on all three designs.

**The log density at the accepted starting points.** What a starting value has
to
do is give Stan a finite log density and a finite gradient, and not sit so far
into a tail that warmup is spent travelling. Both were read off the compiled
Stan
program of the fit `bnec()` builds, with no sampling: 400 accepted starting
points per equation and criterion on the `c_proliferum` series, scored through
`rstan::log_prob()` and `rstan::grad_log_prob()` with `sigma` held at the
residual spread so the comparison is of the curve parameters alone.

| equation | criterion | acceptance | median log density | 10th percentile | median max abs gradient |
|---|---|---|---|---|---|
| `nec3param` | released | 0.121 | −1230 | −1426 | 4105 |
| `nec3param` | band | 0.176 | −1211 | −1847 | 4512 |
| `nec4param` | released | 0.182 | −2246 | −10010 | 14051 |
| `nec4param` | band | 0.202 | −2126 | −9199 | 12451 |
| `ecx4param` | released | 0.310 | −3385 | −11003 | 13712 |
| `ecx4param` | band | 0.334 | −2727 | −11216 | 12005 |
| `ecxwb1` | released | 0.314 | −4169 | −12540 | 18085 |
| `ecxwb1` | band | 0.325 | −3394 | −13023 | 15755 |
| `ecxll4` | released | 0.310 | −3385 | −11003 | 13541 |
| `ecxll4` | band | 0.334 | −2727 | −11216 | 11776 |
| `ecxlin` | released | 0.105 | −2083 | −12569 | 17392 |
| `ecxlin` | band | 0.117 | −1951 | −9243 | 15130 |

The median starting point is closer to the mode under the band for all six
equations and the gradient is smaller for five. The trade the width makes is
visible in the tenth percentile: `nec3param` falls from −1426 to −1847 and
`ecxwb1` from −12540 to −13023, so a wider band does admit a deeper start.

The same measurement on the small unreplicated design, where the criterion
loosens most, shows the trade more plainly. Acceptance roughly doubles and the
starting point is deeper: for `ecx4param`, 0.120 to 0.235 with the median log
density falling from −127 to −152 and the tenth percentile from −233 to −375.
`nec3param` is the exception, its median improving from −133 to −121. Every
accepted point is finite on both designs.

Under **both** criteria a small share of accepted starting points return a
non-finite log density or gradient — 4.5% for `ecxll4` and 3.8% for `ecxwb1`
under the released criterion, 1.8% and 1.5% under the band. That is a property
of
those equations in Stan rather than of either criterion, and this pull request
neither causes nor fixes it.

**What was not done.** #309 asks for confirmation that fits started from the new
initial values reach the same posterior as fits started from the current ones.
That comparison is close to vacuous and was replaced: given convergence the
posterior does not depend on where the chains start, so a match confirms the
sampler rather than the starting values. The four measurements above were run in
its place, at the issue author's direction.

<details>

### The structure

`make_good_inits()` keeps a `filled` flag per chain and a list of accepted
inits.
Each round draws `sum(!filled)` proposals, scores them, applies `refine_inits()`
to the ones that failed only, and fills the slots that passed. `n_t` counts
rounds, so the cap is now exactly `n_trials` rounds where the released loop
allowed one more — it drew the first set before the loop and then tested
`n_t <= n_trials` from `n_t = 1`.

The accepted set is unchanged by this, exactly for a fresh proposal: the four
draws are independent, so conditioning on the other three having passed does not
change the law of the first. Once `refine_inits()` is included it is unchanged
up
to the weight of the fourth power of the per-chain rate, because the released
loop did not refine the set it drew before entering the loop and this one
refines
from the first round. That weight is between 4e−5 and 9e−3 on the designs
measured.

The fallback stays all-or-nothing: where the cap is reached with a slot still
empty, the whole fit is handed to Stan. There are no values for the empty slots
to supply, and starting some chains from the search and the rest from Stan's
`uniform(-2, 2)` draw would make the chains of one fit incomparable during
warmup. The message now says how many chains were found.

`check_init_predictions()` returns `isTRUE(...)` of the conjunction. Its first
two clauses are `NA` on a curve containing `NA`, and the value is read in an
`if`
by `refine_inits()`; the finiteness clauses make the conjunction `FALSE` in
every
case reachable today, so this is a guard rather than a fix for an observed
failure.

### The band

`init_limits(x, y, width = 5, zero_bounded, support)` in `R/inits_functions.R`,
with `replicated_group_means()`, `group_spread()` and `boundary_inset()` beside
it.

**Why the level means and not the two anchors alone.** The anchors are read at
the ends of the predictor. The `c_proliferum` series is not monotone at its
lower
end — its lowest group mean is −0.443 at 15 units against −0.310 at 20 — so the
`bot` anchor sits above the lowest level the design recorded, and an anchor-only
band excluded a curve whose asymptote is at an observed group mean. Only
replicated values contribute: an unreplicated value's "mean" is that
observation,
and admitting those would put the band back on the extrema it exists to leave.

**Why the pooled within-group standard deviation.** Three were measured. The
largest single group's is the least robust: a single aberrant observation
widened
the band by a factor of 6.7, against 4.1 for the pooled spread and 3.0 for
`range(y)`. The standard deviation of the whole response is the most robust, at
1.6, and is not a spread — it grows with the size of the effect, giving a band
2.6 to 3.8 times the response range on the replicated designs against 1.1 to 1.6
for the pooled one. A pooled variance is not robust either, but its failure
direction is permissive: the wider band contains the narrower one, so every
starting value accepted before is still accepted.

The branch is chosen on the total within-group degrees of freedom rather than on
whether any replication exists. One repeated predictor value contributes a
variance on a single degree of freedom, and preferring it to the twenty-odd
differences the rest of the design offers made the spread a function of that one
pair: on a 24-point series, duplicating one value with an equal response sent
the
pooled estimate to zero and the band from 1.35 to 4.83 times the response range.

**Where nothing is replicated** the spread is read from the successive
differences of the response ordered by the predictor, over each half of the
series with the larger taken. `sd(y)` is not used, because it is the statistic
rejected above; on the eight-concentration unreplicated design it gives 0.395
against 0.064 and a band 4.4 times the response range against 1.5. The halves
rather than the whole series because a single number is the wrong summary where
the dispersion changes across the predictor, which is the ordinary case for a
growth or count endpoint and is what `disp()` exists for. It is a maximum over
exactly two estimates, each on at least three differences, so it does not
reproduce the failure that ruled out the largest-group spread: on the same
series
an aberrant observation inflates the band by 20.8× under the halves and 19.0×
under the whole series, against 6.7× versus 4.1× for that comparison.

**Why a width of four.** The rule is the smallest width that covers the
asymptotes of the curve that generated the data in every cell measured, because
a
band excluding them rejects a correct starting value and a wider one admits a
starting point further into a tail. Measured over 6,480 simulated responses:
three generating processes — gaussian, Beta and poisson, so that the clamp and
the boundary inset are exercised and not only the band — by three predictor
grids, three replication levels, three equations, a steep and a shallow curve,
constant and fivefold-rising dispersion, twenty seeds. Of the 270 in-scope
cells,
four covers all of them; three covers 268, two 264 and one 241. By process at
four: 90 of 90 for each of the three.

Two kinds of draw are out of scope, both because no width covers them and the
rule would otherwise never be satisfied: a design whose predictor stops short of
the crossing, and a zero-bounded response whose extreme concentration returns
nothing but zeros, where the `bot` location becomes `min(y[y > 0]) / 10` and the
floor a hundredth of it. Both were confirmed covered at no width before being
excluded.

**Coverage does not choose the spread.** All three estimators reach 270 of 270
at
a width of four and separate only at one, so the case for the pooled and
half-series spread rests on the robustness and band-width measurements above and
not on coverage.

The width was five for three of the five review rounds, under a whole-series
spread on unreplicated designs. The cell that decided it was an unreplicated
design with rising dispersion, missed at the lower end of the response where the
noise is largest — a property of the estimator and not of the band, paid on
every
design including the replicated ones. The reduction is not felt equally: on a
replicated design the pooled spread is unchanged and the band narrows by the
full
fifth; on an unreplicated one the half-series estimator is 7 to 21 per cent
wider
than the whole-series one it replaces, so the band narrows by 5 to 11 per cent.

**The support, and the boundary.** Under the identity link `bnec()` assigns, an
initial curve outside the interval the likelihood permits is invalid rather than
merely poor: Stan rejects it and the fit ends on "Initialization failed".
`range(y)` kept the curve inside the support by accident, because a response is
inside its own support. A band built from a location and a spread has no such
guarantee, so it is intersected with `init_support()` — the support of the mean
mapped through the link, because `pred_<model>()` returns the linear predictor
and the band is built from the response on the link scale, so under `logit` an
unmapped clamp would admit means between 0.5 and 0.73 alone.

Where the band reaches a boundary it stops short of it, by a tenth of the
smaller
of the distance to the nearest level mean inside the boundary and one spread,
and
never further than the distance to the closest value the response takes strictly
inside. A mean at the boundary is one the likelihood cannot evaluate: on the
second block of a hurdle fit the band spans the whole of (0, 1) — one survival
proportion per concentration is an unreplicated series of a few points — and a
curve accepted at a survival of 1e−66 made the joint log likelihood −Inf and the
fit fail to initialise. The gap is read from level means and a spread rather
than
from an observation, because `min(y[y > 0])` is 1 for any integer response and a
floor read from it is the same for eight replicates as for eight hundred; the
observed value enters only as a cap, so the inset can never be stricter at a
boundary than `range(y)` was where the response does not reach it.

`make_good_inits()` takes `family` rather than the answers derived from it, and
has no default for it: a `NULL` default silently drops the clamp, which is how
`test-nechormepwr-bounded.R` began passing when it should not.
`make_good_hurdle_inits()` primes each block under the family that block's
priors
are built from — `hurdle_mu_family()` for the mu block and
`bernoulli(link = "identity")` for the second.

`prior_family_tag()` is factored out of `define_prior()`, which applied two
rewrites inline where only one caller could read them: a `log` or `logit` link
makes the response-scaled priors gaussian whatever the family, and the mu block
of a zero-inflated count is an ordinary count. `zero_bounded_family()` is keyed
on that tag, so the regularizing prior's anchor and the init band cannot
disagree
about where the ends of the curve are.

### What the band does not fix

Where the highest concentration has not reached the lower asymptote, every level
mean and both anchors sit above the true `bot` and widening does not reach it:
over the simulated designs whose predictor stops short of the crossing, coverage
is 0.160 at a width of four and 0.172 at five, against 0.003 at one. That is a
bias and not noise, and
it is the same limitation `regularizing_location()` records for the regularizing
prior. The released criterion is affected identically and worse, because
`min(y)`
is above the true asymptote on such a design as well.

The anchor is shared with the `"regularizing"` prior set only. The
`"uninformative"` set, which is the default, locates `top` and `bot` at
quantiles
of the pooled response and is unchanged, so the shared definition covers one of
the two prior sets.

### Reproducibility

The initial values a given seed produces change, so a fit at a given seed is not
bit-identical to one from 2.1.x. Determinism is preserved: two runs at the same
seed agree, which `test-inits_functions.R` asserts.

### What the proposal count does and does not measure

Both reimplementations in `measure_proposals()` exclude `refine_inits()`, so the
table counts proposals under the two **rules** and is not the wall clock of
either implementation: the released one calls `refine_inits()` on every failing
round, on all four chains, each call making up to 500 single-parameter re-draws
per tunable parameter, so a failing round does far more work than the four
proposals it draws. `measure_timing()` runs both with the rescue in place.

### Tests

`tests/testthat/test-inits_functions.R` gains 100 assertions: `group_spread()`
and its degrees-of-freedom threshold, `successive_difference_spread()`,
`replicated_group_means()`, the band's anchoring on group means, its stability
under added replicates, its containment of every group mean of the `alga`
series, the degenerate fallback, the support clamp, the boundary inset and the
count floor, `init_support()`'s link mapping, `prior_family_tag()` and
`zero_bounded_family()`, the requirement that `family` be supplied, `ecxsigm` on
the `c_proliferum` series, the second block of a hurdle fit not being started at
zero survival, and the structural fix — the last by reimplementing the released
rule beside the new one and comparing proposal counts at three seeds, so the
comparison does not depend on a second working tree.

</details>

https://claude.ai/code/session_01M6NcvKGbgS3hES9AXzfhSD


beckyfisher:
This change was reviewed over five rounds before merge. The defects below were
found to be real and are fixed in the branch; points that were matters of prose
are not listed.

### Defects in the change

1. **Both zero-inflated count families stopped fitting, and a `Gamma(link = "log")` fit reached the #229 failure.** `zero_bounded_family()` read `family$family`, where `define_prior()` reads a tag it rewrites twice — a log or logit link makes the response-scaled priors gaussian, and the mu block of a zero-inflated count is an ordinary count. Both rewrites are now in `prior_family_tag()`, which both sites share.

2. **The support clamp was applied on the response scale to a curve on the link scale.** `pred_<model>()` returns the linear predictor, so under a logit link an unmapped clamp admitted means between 0.5 and 0.73 alone. `init_support()` maps the support through the link.

3. **`family` defaulted to `NULL`, which silently dropped that clamp.** That is how `test-nechormepwr-bounded.R` began passing when it should have failed. It is now a required argument.

4. **The second block of a hurdle fit could be started at zero survival.** Its band spanned the whole of (0, 1), a curve at 1e−66 was accepted, and the joint log likelihood was −Inf — reproduced against `dev`, which fits both models where the branch fitted one. The band now stops short of a boundary of the support.

5. **That fix was then too strict on a zero-bounded response containing zeros**, flooring the band above a generating `bot` of 0.3 and making the criterion stricter than `range(y)` for two of four equations. The gap is now read from the level means and the spread, capped by the closest value the response takes inside the boundary.

6. **One repeated predictor value could decide the spread.** `group_spread()` chose the pooled branch on whether any replication existed, so a single tied pair with an equal response sent the estimate to zero and the band from 1.35 to 4.83 times the response range. The branch is chosen on within-group degrees of freedom.

7. **An empty band accepted every draw rather than rejecting them.** `c(Inf, -Inf)` is read through `min()` and `max()`, which reorder it. `range(y)` had the same defect on the same input.

### Defects in the measurements

8. **The archived script built a different band from the package's**, so one design's figures were of a criterion that does not ship.

9. **The coverage study exercised neither the clamp nor the boundary inset.** It generated only gaussian responses on an unbounded support, so neither fired in any of its 2,160 draws — which is why defects 4 and 5 reached review rather than the script. It now generates Beta and poisson responses as well.

10. **Only the shipped spread was routed through `init_limits()`**, so the two rejected alternatives were scored without the clamp and the inset and appeared better than it.

11. **The width was set by an estimator flaw rather than by the band.** At five it was compensating for a whole-series spread that under-covers on heteroscedastic unreplicated designs, and was paid on every design including the replicated ones. Reading the spread over each half of the series and taking the larger, the width is four, and all three spread estimators cover 270 of 270 in-scope cells there.

### Claims corrected

12. **The boundary floor does not fall as a count design gains replicates.** Over 200 seeds its median is 0.025 at eight replicates and at four hundred; what falls is the interquartile range and the share of designs whose top group is entirely zero. A test had pinned the wrong claim and passed on 86 of 170 seeds.

13. `NAMESPACE` was not regenerated, and six figures in the documentation were quoted from superseded runs.

The full test suite passes with 0 failures. Two measurements in the archived
script have never been run — `measure_timing()`, and the log density on a second
design — and nothing in the pull request is claimed from them.

https://claude.ai/code/session_01M6NcvKGbgS3hES9AXzfhSD

Issue #314, opened 2026-09-11
The regularizing `nec` and `ec50` prior is uniform where the predictor is supplied logged


beckyfisher:
## What

`predictor_prior()` sets the `"regularizing"` spread for `nec` and `ec50` by
multiplying the `"uninformative"` spread by `regularizing_predictor_factor`
(`R/define_prior.R:919-930`). The multiplication means one thing where the user
supplies concentrations as recorded and another where the user supplies
`log(concentration)`.

Where concentrations are supplied as recorded, the uninformative spread is
`half-width / qnorm(0.975)` and the factor is `qnorm(0.975) / qnorm(0.99)`, so
the `qnorm(0.975)` cancels exactly. What is in force is an absolute rule: the
central 98% of the prior reaches the farthest concentration tested.

Where the user supplies `log(concentration)`, the uninformative spread is
`10 * sd(z)`, the constant Fisher et al. (2024) state. On a 0.1 to 100 series
that is 24.872 against a tested range of 6.908 on the log scale, so the prior is
uniform after truncation to the range, and 0.8425 of it is also uniform.

Proposed: state the regularizing spread directly, as `half-width / qnorm(0.99)`,
in both cases. Leave both `"uninformative"` entries exactly as they are.

## Why it matters

`prior_type = "regularizing"` is inert for `nec` and `ec50` where the user has
logged the predictor, which is the ordinary way of supplying a wide dilution
series. It is the case #305 exists to fix, and #305 fixed it for one of the two
ways the predictor can be supplied.

The same experiment also gets a different regularizing prior depending on
whether the user logged the x variable before calling `bnec()`. That is a defect
on its own terms: `prior_type` is a statement about belief, not about which
column the user passed.

The initial-value search draws its proposals from whichever prior is in force
(`R/inits_functions.R:964`, passed through to `make_inits()` at
`R/inits_functions.R:156`), so the entry is not only a statement about the
posterior. Measured below, restating it raises per-chain acceptance against the
band #312 installed and halves the displacement of the accepted `nec`, which is
the one row of #312's evidence table that the band left unchanged.

Nothing released is affected. The whole `"regularizing"` set is unreleased, and
neither `"uninformative"` entry changes.

## Evidence

R 4.6.1, brms 2.23.0, `dev` at 771f9547, 2026-09-11. Prior-only; no model was
fitted.

**Truncated prior CDF at each tested concentration**, series 0.1 to 100
[...]

PR #315, opened 2026-09-11
State the regularizing nec and ec50 spread as a coverage rule (#314)

beckyfisher:
## What

`predictor_prior()` set the `"regularizing"` spread for `nec` and `ec50` by
multiplying the `"uninformative"` spread by `regularizing_predictor_factor`.
The spread is now stated directly, as `half_width / qnorm(0.99)`, on both
branches. `regularizing_predictor_factor` is removed. Neither
`"uninformative"` entry changes, and neither does the regularizing entry
where concentrations are supplied as recorded.

## Why it matters

The multiplication meant one thing on each of the two branches. Where
concentrations are supplied as recorded the uninformative spread is
`half-width / qnorm(0.975)` and the factor is `qnorm(0.975) / qnorm(0.99)`, so
the `qnorm(0.975)` cancels and what was in force was an absolute rule: the
central 98% of the prior reaches the farthest concentration tested. Where the
user supplies `log(concentration)` the uninformative spread is the constant
`10 * sd(z)`, which is not a coverage width, so the same multiplication left a
prior that is uniform across the tested series.

`prior_type = "regularizing"` was therefore inert for `nec` and `ec50` for a
user who had logged the predictor, which is the ordinary way of supplying a
wide dilution series and the case #305 exists to fix. The same experiment also
received a different regularizing prior according to which column was passed to
`bnec()`, which is a defect on its own terms: `prior_type` is a statement about
belief, not about the shape of the input.

Nothing released is affected. `prior_type` does not exist on `master` (2.1.3.1,
the CRAN release), so the whole `"regularizing"` set is unreleased.

## Evidence

R 4.6.1, brms 2.23.0, branch at 4053ba7d against `dev` at fc33ca84,
2026-09-11. No model is fitted anywhere below; every measurement draws from
priors or evaluates a `pred_*()` curve.

**Regression gate.** `notes/scripts/prior_audit.R` was rerun in full: 4,320
rows, 0 errors. Of its 30 `nec` and `ec50` cells, 0 fall outside the central
95% of the truncated prior under either prior type, and the `top` and `bot`
tables are as #305 recorded them (1 regularizing cell of 720, 0
uninformative). Comparing each cell's prior string against the rule this
branch replaces, the `"uninformative"` table is unchanged in every cell, and
so is the `"regularizing"` table on the `identity` and `sqrt` transforms.

One qualification to the issue's statement that the recorded-concentration
string is bit-identical. The two expressions are the same quantity but not the
same double: `(a/b) * (b/c)` and `a/c` agree to about 20 units in the last
place. Over 5,000 randomly generated dilution series the 15 significant digits
`paste0()` writes differed in 36 of them, and one cell of the audit is affected
--- the nassarius contaminant A series read on the recorded and on the
square-root scale, where sigma is `1.93333703285197` against
`1.93333703285198`. The tests therefore assert the quantity numerically on that
branch and reserve literal prior strings for the `"uninformative"` entries,
whose code path is untouched.

**Change gate.** The ten cells that change are the audit's five designs read on
the log scale, by `nec` and `ec50`. The truncated prior CDF at the true value
in those cells:

| | range over the 10 cells | outside the central 95% |
|---|---|---|
| `"uninformative"` | 0.626 to 0.889 | 0 of 10 |
| `"regularizing"`, before | 0.626 to 0.890 | 0 of 10 |
| `"regularizing"`, after | 0.536 to 0.974 | 0 of 10 |

The first two rows differ in the third decimal place, which is the defect: the
regularizing entry was the uninformative one to within a factor that the
truncation then removed.

**The prior on a series supplied logged.** Truncated prior CDF at each tested
concentration, 0.1 to 100 over seven doses supplied as `log(conc)`:

| conc | uninformative | regularizing, before | regularizing, after |
|---|---|---|---|
| 0.3 | 0.158 | 0.158 | 0.052 |
| 1 | 0.333 | 0.333 | 0.226 |
| 3 | 0.492 | 0.492 | 0.499 |
| 10 | 0.667 | 0.667 | 0.793 |
| 30 | 0.826 | 0.827 | 0.945 |

The first two columns are the positions of those doses within the tested range,
which is what a uniform prior gives. The spread is 24.87 before and 1.51 after,
against a tested range of 6.91 on the log scale, and the density ratio across
that range rises from 1.0008 to 1.174.

**Why `q = 0.99` and not a larger quantile.** Central 50% of the truncated
prior as a fraction of the tested range, where 0.50 is uniform, and the prior
mass below the second-lowest dose, over four designs supplied as `log(conc)`:

| design | before | q = 0.99 | q = 0.999 |
|---|---|---|---|
| 0.1 to 100, 7 doses | 0.498 / 0.158 | 0.287 / 0.052 | 0.221 / 0.020 |
| 1 to 20, 5 doses | 0.499 / 0.249 | 0.283 / 0.115 | 0.218 / 0.060 |
| 1 to 1000, 7 doses | 0.498 / 0.166 | 0.283 / 0.051 | 0.218 / 0.019 |
| linear 1 to 10 | 0.498 / 0.297 | 0.340 / 0.094 | 0.278 / 0.038 |

`q = 0.99` is the level a user supplying concentrations as recorded already
receives, so it needs no separate justification. At `q = 0.999` the mass below
the second-lowest dose falls to 0.02 on a wide series, which begins to exclude
a threshold at the bottom of the series --- the case #302 exists to protect.

**Initial-value gate.** `notes/scripts/init_search_audit.R` never set
`prior_type`, so all 340 searches and every table in #307 and #312 were measured
on the `"uninformative"` set. It now takes `prior_type` as a dimension of its
three prior-drawing measurements, and has a sixth design whose predictor is
supplied as `log(conc)`. Per-chain acceptance, 2,000 single-chain draws per
cell, fourteen equations of the declining set per design, released criterion and
the band #312 installed:

| design | uninformative, released / band | regularizing, released / band |
|---|---|---|
| `alga_cp_A` | 0.187 / 0.248 | 0.394 / 0.507 |
| `alga_rs_B` | 0.139 / 0.206 | 0.290 / 0.413 |
| `nec_data` | 0.692 / 0.694 | 0.707 / 0.712 |
| `small_rep` | 0.116 / 0.208 | 0.223 / 0.357 |
| `small_unrep` | 0.112 / 0.259 | 0.226 / 0.467 |
| `log_supplied` | 0.198 / 0.276 | 0.381 / 0.506 |
| **mean** | **0.230 / 0.306** | **0.362 / 0.488** |

`log_supplied` is the design this change reaches: 0.1 to 100 over seven levels
with six replicates, gaussian, supplied as `log(conc)`. Per-chain acceptance
against the band roughly doubles on it, 0.276 to 0.506.

No threshold is set on this measurement; the measurement is the deliverable, and
what it closes is that there was no measurement of the initial-value search
under `prior_type = "regularizing"` at all.

**Proposals drawn to fill four chains**, five seeds, summed over the declining
set, released rule against the change #312 installed:

| design | uninformative | regularizing |
|---|---|---|
| `alga_cp_A` | 148,397 / 237 | 24,292 / 139 |
| `alga_rs_B` | 210,180 / 414 | 90,011 / 188 |
| `nec_data` | 552 / 87 | 538 / 81 |
| `small_rep` | 236,632 / 2,805 | 149,715 / 4,201 |
| `small_unrep` | 334,178 / 308 | 120,602 / 153 |
| `log_supplied` | 128,816 / 257 | 9,478 / 116 |

Of 810 searches under the released rule 111 exhausted the cap of 10,000 rounds
and none did under the change. On the logged design the regularizing set draws
128,816 proposals against 9,478, a factor of 13.6 under the released rule.

**Displacement of the accepted value** relative to the drawn median, in units of
the drawn standard deviation, on the logged design, 6,000 draws per cell:

| parameter | prior set | released criterion | band |
|---|---|---|---|
| `nec`, 3 equations | uninformative | −0.492 | −0.514 |
| | regularizing | −0.233 | −0.244 |
| `ec50`, 8 equations | uninformative | +0.010 | −0.041 |
| | regularizing | +0.037 | +0.001 |

This is the row of #312's evidence table the band left unchanged, and it is the
measurement this change is for. The band corrects a displacement caused by the
response criterion; it cannot correct one caused by a predictor prior that is
uniform across the tested range. `ec50` is unaffected under both sets, as #312
also found.

What this does not cover: the script runs against one working tree and
reimplements the released *search*, not the released *prior*, so it cannot
produce a "regularizing, before" column. The before-and-after prior comparisons
above were measured standalone.

<details>

### The change

`predictor_prior()`, `R/define_prior.R`. `z` is already the predictor on the
log scale on both paths --- `log(u[u > 0])` where concentrations are supplied
as recorded, and `u` itself where the user supplied them logged --- so
`half_width` is computed before the branch and only the uninformative spread
stays case-dependent:

- uninformative, concentrations as recorded: `half_width / qnorm(0.975)`,
  unchanged;
- uninformative, predictor supplied logged: `sd(z) * 10`, unchanged;
- regularizing, both: `half_width / qnorm(0.99)`.

The distribution, the location and the truncation to
`[min(predictor), max(predictor)]` are unchanged.

The degenerate guard keeps its ordering: where the uninformative spread is not
finite or not positive the fallback of 1 stands, and the regularizing rule is
applied after it and not to it. That guard covers the regularizing branch as
well, because the two spreads are positive and finite on exactly the same
designs --- `sd(z)` is so only where `z` has two or more distinct finite
values, which is also when `half_width` is, and the lognormal spread is a
positive multiple of `half_width`. This is stated in a comment rather than
guarded twice.

`regularizing_predictor_factor` is removed. Its roxygen block recorded why the
regularizing set cannot be narrowed to `regularizing_factor` here --- 0.4 puts
the true threshold outside the central 95% in 8 of 30 audit cells, against none
at 0.8 --- and that argument is now in `predictor_prior()`'s own documentation,
where the rule it justifies is stated.

### Documentation

`R/bnec.R` described `prior_type` as narrowing `nec` and `ec50` "to 0.84" of
the uninformative width. That is no longer what the package does, so it now
states the coverage rule. `vignettes/example3.Rmd.orig` stated the same
factor framing, including the sentence that the coverage argument does not
apply to a logged predictor and the factor is applied there anyway, which is
the defect this branch removes; that passage is rewritten. Only the `.orig` is edited. The rendered
`vignettes/example3.Rmd` is stale and still describes the pre-#305 package ---
it states that the regularizing set differs only in the response-scaled
parameters, that the predictor-scaled priors are identical under both, and that
`top` is `beta(5, 1)` under `"regularizing"`. That is the subject of #190,
which is open and covers every vignette in the repository; re-rendering requires
model fits and is done once before release. Hand-editing the prose would leave
it inconsistent with the stale printed `pull_prior` output beside it.

### Tests

`tests/testthat/test-define_prior.R`, six tests:

- the uninformative entry is unchanged for both ways of supplying the
  predictor, as literal strings, and across the audit's design list on the
  `identity` and `sqrt` transforms;
- the regularizing entry is unchanged where concentrations are supplied as
  recorded, asserted numerically against the factor form, which is the
  cancellation identity;
- the same experiment supplied as `conc` and as `log(conc)` gives the same
  spread on the log scale;
- the #305 entry is uniform across the tested range on a logged predictor and
  the stated rule is not;
- the prior mass below the second-lowest tested concentration is above 0.025 on
  each of four designs, supplied both ways;
- the true threshold is inside the central 95% of the truncated prior in each
  of the ten audit cells that change.

The degenerate-fallback test is unchanged and still passes: the fallback of 1
is not narrowed under either prior type.

### `init_search_audit.R`

The script never set `prior_type`, so every table in #307 and #312 was measured
on the `"uninformative"` set. Two changes close that:

- `measure_acceptance()`, `measure_proposals()` and `measure_truncation()` take
  a `prior_type` argument and record it. The runner calls each once per prior
  set under the same seed rather than making `prior_type` an inner loop:
  interleaving would change the draws every design receives, and run this way
  the `"uninformative"` pass reproduces #312. The three Stan measurements are
  untouched.
- A sixth design, `log_supplied`, whose predictor is supplied as `log(conc)`.
  The five designs the script already had are all on the recorded concentration
  scale, where the regularizing `nec` and `ec50` entry does not change, so
  without it the new dimension would measure nothing about this change. It is
  appended last, so the five keep the draws they had.

What the script still cannot produce is the "regularizing, before" column: it
runs against one working tree and reimplements the released *search*, not the
released *prior*. The before-and-after prior comparisons above were measured
standalone.

### Relationship to #312

`init_limits()` uses `regularizing_location()` for the `top` and `bot` anchors.
That is the location half of the regularizing set and is not touched here.
Nothing in `R/inits_functions.R` reads `predictor_prior()` or
`regularizing_factor`. The band, the width of four, the pooled spread, the
support clamp and the boundary inset are unaffected.

### Noted, not changed

`predictor_prior()` selects its branch on `min(unique(predictor)) < 0`. A user
who logs a series whose lowest dose is at or above 1 --- `log(1 to 1000)` gives
0 to 6.9 --- is therefore treated as having supplied raw concentrations, and
receives a prior built from the log of a log, with the value at `conc = 1`
dropped by `u[u > 0]`. Which branch is taken then depends on the units the doses
were recorded in.

On a four-decade series `c(1, 10, 100, 1000)` with six replicates and a true
threshold of 10, supplied as `log(conc)`:

| units | branch | `"regularizing"` entry | CDF at the true threshold |
|---|---|---|---|
| mg/L | lognormal | `lognormal(1.52718, 0.297955)` | **0.0110** |
| mg/mL | normal | `normal(-3.45388, 1.48468)` | 0.2133 |

So on that series the recorded units decide whether the true threshold is
inside the central 95% of its own prior. Two qualifications matter. The effect
is an interaction, not a property of the branch selection alone: the same mg/L
cell sits at 0.0286 under `"uninformative"`, just inside, so it is the branch
selection together with the regularizing narrowing that puts it outside. And
that narrowing is #305's, not #314's --- the lognormal regularizing entry is
bit-for-bit what #305 produced, which the fifteen literal strings and the
identity and sqrt assertions now pin.

None of the audit's 30 cells is affected, because every audit design has a zero
control and the log transform substitutes for it, which puts the minimum below
zero and selects the normal branch. Correcting the branch selection would change
a published `"uninformative"` default, so it is out of scope for #314 and is
recorded here rather than fixed.

</details>

https://claude.ai/code/session_01KUgY59Xc43CyX3yTUnvzhe


beckyfisher:
## Review cycle: what was wrong and what changed

Three rounds of independent review, concluded with no blocking findings. The
body above has been updated to the current state of the branch; what moved is
listed at the end of this comment. The defects below were real and are fixed.
Commits `f7be7a26`, `7c752583` and `a98b6a9d`.

**1. A claim in a committed comment was false, and it was the evidence claim.**
The first version stated that appending the logged design left the five designs
#312 measured on the draws they had. `designs()` calls `set.seed()` inside
itself and the measurement loops call `designs()` once per design iteration, so
every design starts from the same reseeded state and the 42 draws added inside
`designs()` shifted that state for all six. Measured: the first three normals in
each shared design iteration ran −0.7693, 0.9019, 0.8418 before and −0.0003,
−0.2092, −0.8552 after. The logged design's response is now drawn under its own
seed with the global state restored, and the five shared designs are
bit-identical to `dev` on every column. The comment states the real mechanism,
including that reseeding per iteration is what makes list order and length
immaterial.

**2. The audit script asked `check_models()` a question it could not answer.**
`design_parts()` passed a bare `list(predictor = x)`. `check_models()` reads the
predictor through `retrieve_var()`, which looks the column up by the `bnec_pop`
attribute of a `bayesnecformula` model frame, so it returned `NULL` and the
exclusion keyed on a negative predictor was skipped silently. Latent while every
design was on the recorded concentration scale. On the logged design it retained
`ecxsigm`, which returns `NaN` at every `x` below zero — 0 of 2,000 draws gave a
finite prediction — so the cell accepted nothing, ran every search to the cap of
10,000 rounds, and measured a candidate `bnec()` would have dropped. The five
pre-existing designs retain the same equations either way.

**3. The regression gate did not test what it claimed.** One assertion compared
`predictor_prior()` against its own default argument, which is a tautology;
another restated the implementation in the test file. Neither would catch a
change. The `log` transform — the normal branch, which is the branch this diff
refactors — was not covered at all. Replaced with fifteen literal prior strings,
one per cell of the #302 audit, verified against `dev`'s own `predictor_prior()`
with 0 mismatches of 15.

**4. The change gate asserted coverage but never asserted change.** An
implementation that did nothing would have passed it, because the #305 entry
covers those cells too — by being uniform, which is the defect. It now asserts
both halves.

**5. Two numbers were wrong.** The difference between the factor form and the
stated form is **one** unit in the last place, 2.22e-16 over 200,000 series, not
the twenty first reported: that measurement had parsed the printed strings back,
so it measured decimal granularity rather than the doubles. And the two density
ratios were transposed — 1.000588 belongs to a spread of 24.87 and 1.000829 to
0.8425 of it.

**6. The route-equivalence claim was overstated.** The two ways of supplying the
predictor agree on the spread only for a series with **no zero control**. With a
control they cannot agree: the recorded-concentration branch drops the zero,
while a logged predictor includes whatever was substituted for it. On
`c(0, 0.1, 0.3, 1, 3, 10, 30, 100)` the entries are
`lognormal(1.0986, 1.5073)` and `normal(0.5493, 1.7434)`. That is a property of
the substitution, not of the rule, and no choice of spread removes it. The claim
is now scoped wherever it appears.

**7. `?bnec` and `NEWS.md` described the change inaccurately.** The exported
documentation said the regularizing spread reaches the farthest concentration at
the 98% level "against the central 95% under `"uninformative"`", which holds
only on the lognormal branch — on the other the uninformative spread is the
constant `10 sd(x)`, roughly fourteen times past the farthest dose. `NEWS.md`
asserted the two sets differ in the confidence level "and in nothing else" and
then, two paragraphs later, gave the branch on which that is untrue.

**8. A better argument for `q = 0.99` was available than the one given.** The
first version rejected narrower quantiles on the soft ground that prior mass
below the second-lowest dose falls to 0.02. In fact `q = 0.995` and `q = 0.999`
**fail the change gate**: the `log_unit` `ec50` cell sits at 0.9823 and 0.9929.
The cell crosses 0.975 at `q = 0.9905`. That settles the choice, and the margin
at 0.99 — 0.9743 — is now recorded as thin.

<details>
<summary>Smaller corrections, and what was declined</summary>

- Two prior strings differ by one ulp on the unchanged branch, which is four of
  the audit's 30 cells rather than one, since `nec` and `ec50` share a string.
  Both sigmas are now quoted.
- The one-ulp figure was asserted as a derivation. Three roundings against one
  bound the difference at about two; one is what is measured. Stated as such.
- A test comment kept all four retracted figures after the roxygen had corrected
  them, so the file stated both versions.
- `init_search_audit.R`'s header said five designs; `bnec_frame()` passes the
  response as recorded, which is what `bnec()` passes, and now says so; two
  instances of prose ruled out by the house style were removed.
- Recorded rather than changed: the branch is selected by `min(x) < 0`, not by
  the user having logged the predictor, so any predictor spanning negative
  values now receives the narrower prior, which is unaudited outside the logged
  concentration case; and `prior_type` still narrows this entry by different
  amounts on the two routes, because the `"uninformative"` entry remains two
  rules.
- **Declined:** the rendered `vignettes/example3.Rmd` still describes the
  pre-#305 package. That is #190, which covers every vignette in the repository
  and requires model fits to resolve. Hand-editing the prose would leave it
  inconsistent with the stale printed `pull_prior` output beside it. The `.orig`
  is current.

**Changes to the body above:** the evidence section now reports the corrected
ulp and density-ratio figures and the `q` gate result; the route-equivalence
claim is scoped; the implementation fold records the `check_models` fix, the
seeding fix and the branch-selection consequences; and the note on the
`min(x) < 0` branch selection gained the measured counter-example and its
attribution to the interaction with #305's narrowing rather than to the branch
selection alone.

</details>

beckyfisher:
[claude]

Branch `issue-314-regularizing-predictor`, reviewed at a98b6a9d. The
implementation is correct and I verified it independently of the tests: over the
audit's 15 design by transform cells the `"uninformative"` string is identical
to `dev` in every one, the `"regularizing"` entry agrees numerically on the
`identity` and `sqrt` transforms in all five designs, all ten `log` cells change
and none falls outside the central 95%, and the degenerate fallback of 1 is
unnarrowed under both prior types. `test-define_prior.R` passes under
`NOT_CRAN=true`.

One claim needs correcting. It is a documentation and test-scope change, not a
code change.

### The condition on route agreement is not the zero control

The branch states, in `predictor_prior()`'s roxygen and in the pull request
body, that the two routes now agree on the spread "for a series with no zero
control". `tests/testthat/test-define_prior.R:1047-1060` asserts it under the
name "the regularizing spread ignores how x was supplied", with the comment
"The series has no zero control, so `log()` takes it whole and the two routes
describe the same experiment."

Absence of a zero control is necessary but not sufficient. The route is selected
by `spans_negative <- min(u) < 0`, so a series whose lowest tested concentration
is at or above 1 stays non-negative after logging, takes the
recorded-concentration path, and is logged a second time. Measured on this
branch, `prior_type = "regularizing"`, none of the three series having a zero
control:

| series | supplied as `conc` | supplied as `log(conc)` | agree |
|---|---|---|---|
| 0.1 to 100 | `lognormal(1.0986, 1.5073)` | `normal(1.0986, 1.5073)` | yes |
| 1 to 1000 | `lognormal(3.4657, 1.4898)` | `lognormal(1.3851, 0.5304)` | no |
| 10 to 10000 | `lognormal(5.7683, 1.4898)` | `lognormal(1.7524, 0.3948)` | no |

The location differs as well as the spread. The test passes because its series
starts at 0.1.

The correct statement is that the two routes agree where the series has no zero
control **and** its lowest tested concentration is below 1. Both conditions are
needed and only the first is recorded.

### What to change

- `predictor_prior()` roxygen: state both conditions, and state the mechanism —
  a logged series that stays non-negative is not recognised as logged and is
  logged again.
- `tests/testthat/test-define_prior.R:1047`: the test itself is correct and
  should stay. Rename it so it does not claim more than it tests, correct the
  comment, and add a second series with its lowest dose above 1 asserting that
  the two routes differ, so the bound is pinned rather than left to be
  rediscovered.
- The pull request body: same correction to the "What this does not do"
  passage.

### Out of scope here

Do not change the discriminator. Correcting the classification would change the
`"uninformative"` entry for those users as well --- `lognormal(1.7524, 0.4686)`
becomes `normal(5.7683, 1.7683)` on the 10 to 10000 series --- and #314
explicitly excluded changing an uninformative default. It is also only partly
fixable from the data: the model frame column for `crf(log(x), ...)` is named
`log(x)` and is recoverable, but `crf(logx, ...)`, where the user logged the
column in their own data frame, is indistinguishable from concentrations.

The consequence is second-order compared with what this branch fixes. On the 10
to 10000 series supplied logged the misclassified regularizing prior has its
mode at 4.94 against a series median of 5.77, and an interquartile range of
0.365 of the tested range against 0.284 if classified correctly. It is still
peaked inside the tested range and `prior_type` still narrows it, 0.394 to
0.365. What this branch removes is a prior that was uniform and on which
`prior_type` did nothing. A separate issue will cover the discriminator.

### Definition of done for this comment

The three items under "What to change" are applied, the new test asserting the
bound passes, and `test-define_prior.R` still passes in full under
`NOT_CRAN=true`. No model is fitted.

beckyfisher:
Applied in `7c8b0562`, pushed after the check on `a98b6a9d` finished so that run
was not superseded. All five checks passed there.

The correction is right and I reproduced the table. On the three series, none
with a zero control, under `prior_type = "regularizing"`:

| series | as `conc` | as `log(conc)` | agree |
|---|---|---|---|
| 0.1 to 100 | `lognormal(1.0986, 1.5073)` | `normal(1.0986, 1.5073)` | yes |
| 1 to 1000 | `lognormal(3.4539, 1.4847)` | `lognormal(1.3833, 0.5341)` | no |
| 10 to 10000 | `lognormal(5.7565, 1.4847)` | `lognormal(1.7503, 0.3939)` | no |

The series are `round(exp(seq(log(lo), log(hi), length.out = 7)), 3)`, so the
last two rows differ in the third decimal from the figures in the comment; the
conclusion is unchanged.

**The three items.** The roxygen now states both conditions and the mechanism.
`NEWS.md` and `vignettes/example3.Rmd.orig` stated the same claim and are
corrected the same way — they were not in the list but would otherwise have been
left wrong. The test is renamed to "both routes agree below a lowest dose of 1
and not above" and now asserts both conditions, with the second series pinning
that the location differs as well as the spread.

**One addition beyond what was asked.** The test also pins the second-order
claim, since the comment relies on it to argue the discriminator is out of
scope: on the 10 to 10000 series supplied logged, the misclassified entry still
has its mode inside the tested range and `prior_type` still narrows it. If that
ever stops holding, the out-of-scope argument stops holding with it, and the
test will say so.

**Discriminator unchanged**, for the reasons given. `test-define_prior.R` passes
in full under `NOT_CRAN=true`, 748 assertions. No model was fitted.

beckyfisher:
The branch-selection limitation recorded in `predictor_prior()`'s roxygen and in the "Noted, not changed" section above is now tracked as #317, with the `master` comparison: the discriminator is `min(x) < 0` on the CRAN release as well, so a logged series with no negative values receives `gamma(5, 4/m)` there and a log-of-a-log `lognormal` here. The four coverage failures measured under `"regularizing"` are that misclassification combined with #305's narrowing, not anything this pull request changes.

#316 tracks the separate finding that the test suite runs at its parallel floor, which came out of measuring why the check matrix takes 38 minutes.

PR #318, opened 2026-09-11
Fit the model set under the user's future plan (#184)

beckyfisher:
## What

`bnec()` and `amend()` now fit their model sets under whatever `future` plan is
set when they are called. Setting `plan(multisession, workers = 4)` before the
call fits four models at a time; no plan, or `plan(sequential)`, fits them one
at a time exactly as every earlier version did.

```r
library(future)
plan(multisession, workers = 4)
fit <- bnec(y ~ crf(x, model = "decline"), data = my_data, seed = 17)
plan(sequential)
```

There is no new argument to `bnec()` and no new hard dependency. `future` and
`future.apply` are Suggests, reached through `requireNamespace()` the way
`R.utils` is for `timeout`.

## Why it matters

A 23-model run is 23 sequential `brm()` calls, each of which leaves most of the
machine idle. Fitting the models at the same time is the largest speed-up
available for a multi-model run, which is the normal way the package is used.

Two things had to be made true rather than assumed, and the first is the reason
this issue was held for attended work:

**Each fitted model reproduces exactly under a parallel plan, for the same
seed.** It would not have, by default, and nothing would have said so.
`future.seed = TRUE` installs an L'Ecuyer-CMRG generator in each worker.
`make_good_inits()` seeds the initial-value search with
`set.seed(brm_args$seed)`, and `set.seed()` called with the default
`kind = NULL` sets a seed without restoring the generator kind. The same seed
therefore draws from a different generator in a worker, and the fits differ. No
error, no warning. The worker now restores the parent's RNG kind before fitting.

**The model-averaged quantities are the exception, and it is stated rather than
papered over.** After the models are fitted, `expand_manec()` draws a seed from
the session's RNG stream and uses it to decide which posterior draws each
equation contributes to the averaged estimate. Fitted in sequence, that stream
has been advanced by every model's initial-value search and so is fixed by
`seed`; fitted in parallel those searches run in workers and it is not. The
averaged `nec`, its interval and the stored prediction grid therefore differ
between the two plans, as two valid realisations of the same weighting.

`bnec()` now leaves the stream undisturbed under a parallel plan, so
`set.seed()` in the calling session fixes that draw and two parallel runs agree
— which is what `expand_manec()` says the draw is for (#216). Closing the
remaining gap means deriving it from `seed` rather than from the ambient stream,
which is a change to `expand_manec()` and to what #216 decided deliberately. It
is not attempted here.

**Chains are sampled in sequence inside each worker.** `brm()` parallelises
across chains already, so models in parallel on top of that requests
`workers x chains` processes, which on most machines is slower than fitting in
sequence. Under a parallel plan of more than one worker `bnec()` therefore
passes `cores = 1` to `brm()`. A `cores` you passed yourself is left alone and
reaches `brm()` through `...` as it always did, which is how the two levels are
nested deliberately: `bnec(..., cores = 2)` with `workers = 4` runs four worker
processes sampling eight chains between them.

Worth being exact about what the clamp prevents, because the obvious account of
it is wrong and this PR asserted it for two rounds. `future` sets `mc.cores` to
1 inside a worker of its own accord, so a value set in a profile does not in
fact reach `brms` there — measured at 1 under `multisession`, `multicore` and
`cluster` with `options(mc.cores = 7)` in `R_PROFILE_USER`. `cores = 1` is
therefore belt and braces, not the only thing standing between you and sixteen
processes. It is kept because it makes what `brms` is asked to do a property of
this package rather than of `future`'s internals.

A plan that names a parallel strategy can still resolve to a single worker —
`plan(multicore)` does exactly that wherever forking is unavailable, which is
Windows, and RStudio or Positron on Linux and macOS. Clamping there would fit one
model at a time with its chains in sequence, slower than the current release for
anyone who has set `mc.cores`. Neither the clamp nor the parallel announcement
applies in that case.

## Evidence

The `future`/WSL testing pass this issue asked for before implementation is
posted as a comment on #184. In summary: all five local backends work under WSL2
(`sequential`, `multisession`, `multicore`, `cluster`, and an explicit
`makeClusterPSOCK`), as do five plan transitions within one session; `try()` and
custom conditions survive the round trip; `R.utils::withTimeout` still fires
inside a worker; and the trouble reported under WSL was not reproduced across
three runs.

The RNG defect above, measured 2026-09-11 under R 4.6.1, `future` 1.70.0, on
every backend including `plan(sequential)`:

| body | parent | worker |
|---|---|---|
| `set.seed(43); runif(1)` | 0.48503768 | 0.47899959 |

Restoring the RNG kind removes the difference on all five backends.

No timing is reported. The development machine was running three other R
sessions' simulations throughout, and a wall-clock figure taken under that load
would not mean anything. What the mechanism does to run time on an idle machine
is therefore not established by this PR, and the acceptance criterion asking for
a measured speed-up is not met. The correctness criteria are.

Three further measurements, all made while responding to review:

| what | before | after |
|---|---|---|
| applied function, serialised, over an 8 MB object in the calling frame | 7.630 MiB | 0.000 MiB |
| `draw_seed` over two parallel runs of one call | differed | identical, and equal to an undisturbed stream |
| `cores` under `plan(multicore, workers = 4)` with forking unavailable | 1 | unset |

`future` exports the applied function together with its enclosing environment,
and that environment was the `bnec()` or `amend_model_set()` frame — which holds
the data, the model frame, the priors, and for `amend()` every fit already in the
set. `narrow_environment()` rebuilds the function with only the names it reads,
parented on the package namespace. It is applied on the sequential path too, so a
name left out of the list fails on the first ordinary call rather than only for
whoever sets a plan.

`future_lapply()` also chunks one-per-worker by default, so a worker held every
fit in its chunk until the chunk finished and a worker drawing the slow equations
set the wall clock. `future.chunk.size = 1` assigns dynamically and holds one fit
at a time.

Compilation is not duplicated by this change. Each bayesnec equation is a
distinct Stan program, so a set of *n* models compiles *n* programs whether they
are fitted in sequence or at the same time; what changes is that the compiles
overlap. The `cmdstanr` backend #308 adopted caches a compiled program by the
hash of its Stan source, so a second run of the same set reuses it, which is
strictly better under a parallel plan than under a sequential one.

<details>

### Files

`R/parallel_models.R` is new and holds three internals:

- `bnec_plan_is_parallel()` reads the strategy off `future::plan("list")[[1]]`
  rather than counting workers. A `cluster` plan pointed at one remote node
  reports a single worker but is still a parallel plan, and the question is
  whether the user asked for one. Returns `FALSE` where either package is
  absent, which is what keeps the sequential path free of the dependency.
- `plan_model_set(brm_args, n_models)` makes the decision and returns the
  `brm_args` that go with it, so the two are made in one place and cannot
  disagree. Fewer than two models to fit is never parallel: a parallel plan
  provides no benefit over one `brm()` call, and applying the `cores` limit to a
  lone fit would make it slower than it is today. It also reports what it
  decided, because a parallel run emits its per-model messages out of order or
  not at all.
- `bnec_model_lapply(X, FUN, parallel)` is `lapply()` when `parallel` is
  `FALSE`, which is every call made today, and `future.apply::future_lapply()`
  with the RNG-kind restore when it is not.

`R/bnec.R` and `R/amend.R` each replace their `for` loop with a call to
`bnec_model_lapply()`. `try()` stays *inside* the applied function: the
`NA`-on-failure contract is what several downstream functions read, and an error
escaping a worker aborts the whole batch rather than one model of it. The
assembly of `mod_fits` and `failed` stays in the parent, so what the applied
function returns is exactly what the sequential path returned -- the fit, or the
`try-error` holding its condition -- and `failure_record()` reads the same object
either way.

`amend()` needed one thing `bnec()` did not. Its loop carries an existing fit
over from `old_fits` without a `brm()` call, and returning those from the applied
function would serialise every existing fit out to a worker and straight back for
a model that is not refitted -- the memory hazard the issue raises, incurred for
nothing. Carry-over is done in the parent and only the models needing a fit reach
the applied function, which is also what `plan_model_set()` counts.

### `future.seed = TRUE`, with the kind restored

`future.seed = NULL` also gives draws matching `lapply()`, and was checked rather
than assumed: it is silent, and only `future.seed = FALSE` reports
`UNRELIABLE VALUE`. It was not chosen because it leaves the worker's RNG state to
the backend, making correctness a property of every code path inside
`fit_bayesnec()` seeding itself rather than of the dispatcher.

Restoring the kind discards the L'Ecuyer stream, and does not leave workers
drawing in step: `RNGkind()` re-initialises `.Random.seed` from the clock and
process id. Measured under `plan(multicore, workers = 3)`, three of three draws
distinct, with the restore in place as without it.

Where no seed is supplied, the search calls `set.seed(NULL)` and reseeds from
entropy, so no run repeats under any plan. That is what the package does today.

### Tests

`tests/testthat/test-parallel_models.R`. The plan is restored after each test and
the whole file skips on CRAN.

- the sequential path is `lapply()` and leaves `.Random.seed` untouched;
- `plan_model_set()` declines to parallelise fewer than two models, sets
  `cores = 1` when it does, and leaves a supplied `cores` alone;
- a parallel run draws the same values from the same seed as a sequential one --
  the acceptance criterion, reduced to the mechanism it depends on;
- a canary asserting that `future`'s own seeding *would* have changed those
  draws. A failure there means `future` changed how it seeds a worker, not that
  bayesnec is broken, and the comment says so;
- a failing element yields its `try-error` with the `bnec_fit_failure` condition
  and its fields intact, and the rest still run, on both paths;
- the initial-value search itself draws the same values in a worker as in the
  parent for the same seed, which is the same assertion at the call site that
  matters and costs no sampling time;
- the parallel loop leaves the caller's `.Random.seed` exactly as it found it,
  and leaves none behind where there was none;
- a one-worker parallel plan does not take chains away from `brms`;
- the applied function's environment holds only what it names, and is parented
  on the namespace;
- a two-model `bnec()` set fitted under `plan(multisession)` reproduces the
  sequential fit draw for draw. This one is not cheap: it compiles four Stan
  programs. It is kept because it is the acceptance criterion stated end to end
  rather than at the mechanism;
- a timeout ends a parallel run exactly as it ends a sequential one.

Two things about that last one. `timeout = 0.001` lets the `TimeoutException`
escape `fit_bayesnec()`'s `tryCatch` rather than being recorded as a failure --
identically under both plans, so it is pre-existing behaviour and not something
this change introduced. The test asserts the two agree rather than asserting an
outcome. Worth a separate issue if it is not already one.

### A development trap worth knowing

A `multisession` or `cluster` worker loads the **installed** bayesnec, not one
loaded with `pkgload::load_all()`. Under `devtools::test()` against an older
installation a worker reports `object 'bnec_model_lapply' not found`;
`multicore`, which forks, inherits the loaded namespace. `R CMD check` installs
first, so CI is unaffected, and the test file detects this and skips with a
reason rather than failing. It is also stated in `?bnec`.

Relatedly, `future` sets the plan inside a worker to `sequential`, so a `bnec()`
call made from within one does not parallelise again.

### Out of scope, as the issue asks

`bnec_hurdle()`'s two components stay sequential; each of its `bnec()` calls uses
the plan for its own model set. Nothing in `expand_manec()` or the post-fit path
is changed.

</details>

Addresses #184. Not marked as closing it: the issue's acceptance criteria
include a measured speed-up including compilation, and no timing is reported
here for the reason given above. The correctness criteria are met. Whether to
close on this PR and open a separate benchmarking issue, or hold #184 open for
it, is your call.

https://claude.ai/code/session_01EgdwdPQL8KNApFDgYWkPeK

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,6 +28,8 @@ Suggests:
     knitr,
     rmarkdown,
     R.utils,
+    future,
+    future.apply,
     testthat (>= 3.1.8)
 VignetteBuilder: knitr
 URL: https://open-aims.github.io/bayesnec/
--- a/NEWS.md
+++ b/NEWS.md
@@ -608,12 +608,13 @@
   across chains already, so models in parallel on top of that would request
   `workers x chains` processes --- sixteen for four workers and the default
   `chains = 4`. Under a parallel plan of more than one worker `bnec()` passes
-  `cores = 1` to `brm()`, which also stops a `mc.cores` set in a profile from
-  applying per worker. Passing `cores` yourself is left alone and is how the two
-  levels are nested deliberately. A plan that names a parallel strategy but
+  `cores = 1` to `brm()`. Passing `cores` yourself is left alone and is how the
+  two levels are nested deliberately. A plan that names a parallel strategy but
   resolves to a single worker --- `plan(multicore)` wherever forking is
   unavailable, which includes Windows --- is clamped in neither respect, and
-  says so.
+  says so. Note that `future` already sets `mc.cores` to 1 inside a worker, so
+  a value set in a profile does not reach `brms` there in any case; `cores = 1`
+  makes that a property of this package rather than of `future`'s internals.
 
   **Each fitted model reproduces exactly under a parallel plan, for the same
   `seed`.** This needed making true rather than being inherited:
@@ -626,14 +627,15 @@
 
   **The model-averaged quantities are the exception.** `expand_manec()` draws
   the seed for the weighted posterior draw from the session's RNG stream after
-  the models are fitted. Fitted in sequence that stream has been advanced by
-  every model's initial-value search and so is fixed by `seed`; fitted in
-  parallel those searches run in workers and it is not. The averaged `nec`, its
-  interval and the stored prediction grid therefore differ between the two
-  plans, as two valid realisations of the same weighting. `bnec()` leaves the
-  stream undisturbed under a parallel plan, so `set.seed()` in the calling
-  session fixes that draw and two parallel runs agree --- which is what
-  `expand_manec()` intends it to answer to (#216).
+  the models are fitted. A sequential run advances that stream, through the
+  `set.seed(seed)` each initial-value search makes, so the draw is fixed by
+  `seed`; under a parallel plan the model loop leaves the stream alone. The
+  averaged `nec`, its interval and the stored prediction grid therefore differ
+  between the two plans, as two valid realisations of the same weighting.
+  `set.seed()` in the calling session fixes that draw under a parallel plan, so
+  two parallel runs agree --- which is what `expand_manec()` intends it to
+  answer to (#216). One further consequence: a parallel call does not advance
+  the calling session's RNG stream, where a sequential one does.
 
 - **`bnec_record()`** reports what `bnec()` did to the request before fitting:
   the candidate set as requested, the set attempted, the equations excluded with
--- a/NEWS.md
+++ b/NEWS.md
@@ -607,19 +607,33 @@
   **Chains are sampled in sequence inside each worker.** `brm()` parallelises
   across chains already, so models in parallel on top of that would request
   `workers x chains` processes --- sixteen for four workers and the default
-  `chains = 4`. Under a parallel plan `bnec()` passes `cores = 1` to `brm()`,
-  which also stops a `mc.cores` set in a profile from applying per worker.
-  Supplying `cores` in `brm_args` is left alone and is how the two levels are
-  nested deliberately.
+  `chains = 4`. Under a parallel plan of more than one worker `bnec()` passes
+  `cores = 1` to `brm()`, which also stops a `mc.cores` set in a profile from
+  applying per worker. Passing `cores` yourself is left alone and is how the two
+  levels are nested deliberately. A plan that names a parallel strategy but
+  resolves to a single worker --- `plan(multicore)` wherever forking is
+  unavailable, which includes Windows --- is clamped in neither respect, and
+  says so.
 
-  **A parallel run reproduces the sequential run exactly for the same `seed`.**
-  This needed making true rather than being inherited: `future.seed = TRUE`
-  installs an L'Ecuyer-CMRG generator in each worker, `set.seed()` does not
-  restore the generator kind, and the initial-value search would therefore have
-  drawn different starting values in a worker from the same seed --- silently,
-  with no error and no warning. The worker restores the parent's RNG kind before
-  fitting. Without a `seed` the search reseeds from entropy and no run repeats,
-  under a plan or otherwise, which is unchanged.
+  **Each fitted model reproduces exactly under a parallel plan, for the same
+  `seed`.** This needed making true rather than being inherited:
+  `future.seed = TRUE` installs an L'Ecuyer-CMRG generator in each worker,
+  `set.seed()` does not restore the generator kind, and the initial-value search
+  would therefore have drawn different starting values in a worker from the same
+  seed --- silently, with no error and no warning. The worker restores the
+  parent's RNG kind before fitting. Without a `seed` the search reseeds from
+  entropy and no run repeats, under a plan or otherwise, which is unchanged.
+
+  **The model-averaged quantities are the exception.** `expand_manec()` draws
+  the seed for the weighted posterior draw from the session's RNG stream after
+  the models are fitted. Fitted in sequence that stream has been advanced by
+  every model's initial-value search and so is fixed by `seed`; fitted in
+  parallel those searches run in workers and it is not. The averaged `nec`, its
+  interval and the stored prediction grid therefore differ between the two
+  plans, as two valid realisations of the same weighting. `bnec()` leaves the
+  stream undisturbed under a parallel plan, so `set.seed()` in the calling
+  session fixes that draw and two parallel runs agree --- which is what
+  `expand_manec()` intends it to answer to (#216).
 
 - **`bnec_record()`** reports what `bnec()` did to the request before fitting:
   the candidate set as requested, the set attempted, the equations excluded with
--- a/NEWS.md
+++ b/NEWS.md
@@ -595,6 +595,32 @@
 
 ## New
 
+- **A model set can now be fitted in parallel.** `bnec()` and `amend()` fit
+  their models under whatever `future` plan is set when they are called, so
+  `plan(multisession, workers = 4)` before the call fits four models at a time
+  and `plan(sequential)`, or no plan at all, fits them one at a time exactly as
+  every earlier version did. There is no new argument: the plan already holds
+  that state, and a `cores` argument beside it would be ambiguous the moment a
+  user set both. `future` and `future.apply` are Suggests, so a session without
+  them takes the sequential path and reports nothing (#184).
+
+  **Chains are sampled in sequence inside each worker.** `brm()` parallelises
+  across chains already, so models in parallel on top of that would request
+  `workers x chains` processes --- sixteen for four workers and the default
+  `chains = 4`. Under a parallel plan `bnec()` passes `cores = 1` to `brm()`,
+  which also stops a `mc.cores` set in a profile from applying per worker.
+  Supplying `cores` in `brm_args` is left alone and is how the two levels are
+  nested deliberately.
+
+  **A parallel run reproduces the sequential run exactly for the same `seed`.**
+  This needed making true rather than being inherited: `future.seed = TRUE`
+  installs an L'Ecuyer-CMRG generator in each worker, `set.seed()` does not
+  restore the generator kind, and the initial-value search would therefore have
+  drawn different starting values in a worker from the same seed --- silently,
+  with no error and no warning. The worker restores the parent's RNG kind before
+  fitting. Without a `seed` the search reseeds from entropy and no run repeats,
+  under a plan or otherwise, which is unchanged.
+
 - **`bnec_record()`** reports what `bnec()` did to the request before fitting:
   the candidate set as requested, the set attempted, the equations excluded with
   the reason for each, and any substitution made in the response. Both were
--- a/R/amend.R
+++ b/R/amend.R
@@ -239,65 +239,88 @@ amend_model_set <- function(object, mod_fits, old_method, drop = NULL,
   mod_fits <- vector(mode = "list", length = length(model_set))
   names(mod_fits) <- model_set
   failed <- list()
-  for (m in seq_along(model_set)) {
+  # Which models this call has to fit. A model already in the set is carried
+  # over from old_fits without a brm() call, so a set of twenty with one
+  # addition is a one-model run, and only the additions decide whether a
+  # parallel plan is worth using.
+  needs_fit <- !vapply(model_set, function(model) {
+    inherits(try(old_fits[[model]], silent = TRUE), "prebayesnecfit")
+  }, logical(1))
+  # Carried over here, in the parent, rather than returned from the applied
+  # function. Under a parallel plan the alternative serialises every existing
+  # fit out to a worker and straight back again, for a model that is not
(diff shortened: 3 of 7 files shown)


beckyfisher:
## What review changed in this PR

Four review rounds against the branch. Recording the defects that were real,
since the commits alone do not say which were found rather than intended.

### Defects

**1. A parallel run was not reproducible against itself.** `expand_manec()`
draws the seed for the weighted posterior draw from the session's RNG stream
immediately after the model loop, and a parallel loop advances that stream by
generating its per-element seeds. Two parallel runs of one call gave different
`w_draw_seed`, and so different model-averaged `nec`, interval and prediction
grid. The loop now leaves the caller's stream where it found it, as
`weighted_draw_index()` already does: `draw_seed` is identical across runs and
equal to an undisturbed stream.

It still does not equal the sequential run's, because a sequential run advances
the stream through each model's `set.seed(seed)`, and nothing in a worker can
replay that. The PR claimed it did. The claim is corrected rather than the
behaviour forced, since closing the gap means changing what #216 decided
deliberately. `?bnec` and NEWS now state exactly which quantities reproduce.

**2. Every worker was sent the whole calling frame.** `future` exports the
applied function together with its enclosing environment, which was the
`bnec()` or `amend_model_set()` frame — the data, the model frame, the priors,
and for `amend()` every fit already in the set. Carrying existing fits over in
the parent, which the first commit did for this reason, saved nothing while they
stayed reachable from the frame. `narrow_environment()` rebuilds the applied
function with only the names it reads, parented on the namespace: **14.166 MiB
to 0.017 MiB** serialised on the two-equation `manec_example`. It is applied on
the sequential path too, so a name left out fails on the first ordinary call.

**3. A one-worker parallel plan was slower than the release.**
`bnec_plan_is_parallel()` tests the strategy, not the worker count, and
`plan(multicore)` resolves to one worker wherever forking is unavailable —
Windows, and RStudio or Positron. `cores` was clamped to 1 there, so a user with
`mc.cores` set fitted one model at a time with its chains in sequence, while the
message said models were being fitted in parallel. The clamp now requires more
than one worker.

**4. Static chunking removed load balancing and multiplied peak memory.**
`future_lapply()` chunks one per worker by default, so a worker held every fit
in its chunk until the chunk ended and whichever worker drew the slow equations
set the wall clock. `future.chunk.size = 1` assigns dynamically and holds one
fit. The trade — data and priors sent once per model rather than once per worker
— is stated in the code, and no timing has been taken either way.

**5. A warning per model under the pre-3.6.0 sampler.** The worker's `RNGkind()`
restore relayed `non-uniform 'Rounding' sampler used` once per model where the
session had set `sample.kind = "Rounding"`. Suppressed, as
`weighted_draw_index()` does at its own restore.

### Claims that were wrong

**The stated reason for `cores = 1` does not occur.** Three places asserted that
without it, `mc.cores` from a profile would give every worker four chains and a
four-worker plan sixteen processes. `future` sets `mc.cores` to 1 inside a worker
itself: measured with `options(mc.cores = 7)` in `R_PROFILE_USER`,
`getOption("mc.cores")` in a future is 1 under `multisession` at two and four
workers, under `multicore`, and under a two-node `cluster`. The clamp is kept —
it makes what brms is asked to do a property of this package rather than of
another's internals — and the justification is replaced with the measurement.

Four smaller ones, each a single clause: the averaged-quantity difference is a
property of the plan and not of where the initial-value searches ran, so it holds
at one worker too; `init` alone does not stop a sequential run advancing the
stream, because brms defaults `seed` to `NA` and `rstan:::check_seed` then draws
one per `brm()` call; a formula keeps the environment it was written in, so one
built inside a function sends that frame to every worker (6.18 MiB beside a
6.2 MB matrix), as does any function passed through `...`; and the one-worker
message told a single-node `cluster` user that nothing had changed about chain
sampling when `future` had in fact taken it.

### Raised and not changed, with reasons

- **`w_ne` equality is not asserted in the end-to-end test.** It is expected to
  differ; asserting a difference is a fragile pin. The mechanism is asserted
  instead — that the loop does not disturb the caller's stream.
- **The expensive test is not behind a new environment variable.** The
  convention here is `NOT_CRAN`, and a variable nobody sets is a test that never
  runs. It compiles four Stan programs and is labelled as the acceptance
  criterion stated end to end.
- **The bold lead-ins in `NEWS.md` are left.** §12's 2026-09-08 ruling does
  apply to new prose — the `?bnec` ones were removed on that basis — but every
  entry in the 2.2.0 section uses the list-item pattern, and converting only
  this one would make the file inconsistent. **Recording it as a deferral rather
  than an oversight**, for whoever does the whole-file pass.

### Still open

The acceptance criterion asking for a measured speed-up including compilation is
unmet, which is why this says "Addresses #184" rather than closing it. The
machine has been running other sessions' simulations throughout and any wall
clock taken under that load would not mean anything. The correctness criteria
are met and are tested.

Two things measured along the way that bear on when a plan is worth setting: a
`multisession` worker must load bayesnec and with it brms, Rcpp and rstan before
it can fit anything, about three seconds per worker on an idle machine; and
memory rather than cores is what binds, since a worker killed by the OS cannot
raise a condition for `try()` and ends the whole call. Both are in `?bnec`.

### Two defects CI found in the tests, after review had finished

Neither is in the package, but both are worth recording because the first was
the kind of mistake that makes a green test meaningless.

**The RNG-stream test took its baseline in the wrong place.**
`skip_unless_worker_sees_internals()` sits inside the timed block and calls
`future_lapply()` itself, which advances the parent's stream. The baseline was
captured before that, so it was a state the model loop was never given, and the
comparison failed for that reason rather than for anything the restore did. The
forking test added shortly after repeated it. Both now take the baseline inside
the plan, after the skip check, and return what they measured instead of reading
the global afterwards. The restore itself was correct throughout.

**`parallelly::supportsMulticore()` was an undeclared import**, and this
workflow runs `error_on: "warning"`, so it would have failed the build on its
own. `future::nbrOfWorkers()` answers the same question from a package already
declared: `future` reports one worker wherever it cannot fork.

### On the evidence behind this

Every measurement quoted above was taken on the WSL2 side of a workstation
running several other sessions' Stan fits throughout, and each unit check was
run in a fresh process for that reason. The local run of the full test file was
killed three times by the OS, twice with about 1 GB of 31 free — which is the
memory constraint described above, arriving unprompted.

The platform evidence is therefore the CI matrix, not a local run:

| | result | `test-parallel_models.R` |
|---|---|---|
| Windows (release) | pass, 41m | all run bar the forking block, skipped as "forking is not available on this platform" |
| macOS (release) | pass, 33m | all run, forking block included |
| Ubuntu (release) | pass, 31m | all run |
| Ubuntu (devel) | pass, 31m | all run |

`FAIL 0` on all four. Neither the socket nor the fork path is assumed anywhere:
each is exercised on every platform that has it.

beckyfisher:
Correction to the body, since you are reading it now.

The paragraph on chain parallelism was the pre-review version. It said "bayesnec passes no `cores` of its own, so brms falls back to `getOption("mc.cores")`, and a value set in a profile would give every worker four chains". Two things wrong with that.

`cores` supplied by the user has always reached `brm()`: `bnec()`'s `...` is collected into `brm_args` and `fit_bayesnec()` passes that straight to `do.call(brm, ...)`. That is what makes `"cores" %in% names(brm_args)` a valid test for "the user asked for it". The sentence meant that bayesnec originates no value of its own, which after this PR is not true either, since it sets `cores = 1` under a parallel plan of more than one worker.

And the `mc.cores` hazard it described does not occur: `future` sets `mc.cores` to 1 inside a worker itself, measured at 1 under multisession, multicore and cluster with `options(mc.cores = 7)` in `R_PROFILE_USER`. That was found in review, corrected in `?bnec`, `R/parallel_models.R` and `NEWS.md`, and I failed to propagate it here. The body now matches the code. No code changed.

beckyfisher (commented):
This looks great. My only comments are that the addition to the bnec help file is very large and there is a lot of prose in the model files that explain what has been done. This should be oppened as a an issue to build vignette, "Running models in parallel". This current version can be merged so it can be used, the vignette issue is docs only and doesn't impact any existing behavior.

PR #321, opened 2026-09-11
Supply the model-averaging weight default wherever a set is assembled

### LINKED ISSUE #320: Model weights use stacking rather than the pseudobma default when a set is assembled with c(), + or amend()
## What

A model-averaged set assembled with `c()`, `+` or `amend()` is weighted by
stacking, not by the documented default of pseudo-BMA. The same set fitted in a
single `bnec()` call is weighted by pseudo-BMA. The weighting method therefore
depends on how the set was assembled rather than on what was requested.

On that path `attr(mod_stats$wi, "method")` is also empty, so the object does not
record which method produced its weights.

## Why it matters

Weights determine the model-averaged NEC, NSEC and ECx estimates, so two
analysts fitting the same equations to the same data get different toxicity
estimates depending on whether they called `bnec()` with a model vector or
combined single fits with `c()`.

Combining single fits is a documented workflow and is how the `bnec` training
material introduces model averaging, so this is a common path rather than an
edge case.

The empty `method` attribute means an affected result cannot be identified after
the fact from the fitted object.

## Evidence

`nec_data`, equations `nec3param` and `ecxll3`, 8000 draws, `bayesnec` 2.1.3.33
at `dev` `fc33ca84`, R 4.6.1 on Debian, `cmdstanr` 0.9.0, CmdStan 2.39.0.

| Route | `attr(wi, "method")` | `nec3param` | `ecxll3` |
|---|---|---|---|
| `bnec(y ~ crf(x, model = c("nec3param", "ecxll3")))` | `pseudobma` | 0.809 | 0.191 |
| `c(fit_nec3param, fit_ecxll3)` | *(empty)* | 0.867 | 0.133 |
| `amend(combined, loo_controls = list(weights = list(method = "pseudobma")))` | `pseudobma` | 0.781 | 0.219 |

Rows two and three are the controlled comparison: the same two fitted models,
differing only in the weighting method. The weight on `nec3param` changes by 8.7
percentage points.

Row one is not directly comparable with the others, because those models were
fitted in a separate call and so are a different sample from the posterior.

The printed summary does report `Method: stacking_weights` on the affected path,
so the output is not misleading to anyone who reads it; the defect is that
stacking is not what was asked for, and that the attribute is unset.

<details>

**Cause.** `expand_manec()` builds its own `loo_controls` when the argument is
missing:

```r
loo_controls <- list(fitting = list(), weights = list())
loo_controls <- validate_loo_controls(loo_controls, fam_tag)
```

`validate_loo_controls()` only validates. Its `else` branch sets
`loo_controls$weights <- list()` and returns, so no method is injected.
`loo_w_controls$method` is then `NULL`, `do.call(loo_model_weights, loo_mw_args)`
receives no `method`, and `loo` applies its own default of `"stacking"`.
`attr(mod_stats$wi, "method") <- loo_w_controls$method` assigns `NULL`, which is
why the attribute is empty.

`define_loo_controls()` is the function that injects the default, and it is not
called here.

**Which entry points are affected.**

- `bnec()` calls `define_loo_controls(loo_controls, brm_args$family$family)` and
  passes the result down. Not affected.
- `c.bnecfit()` calls `expand_manec(mod_fits, formulas)` with no `loo_controls`.
  Affected.
- `amend()` forwards `loo_controls` only when it was supplied. Affected when it
  is not.
- `+` shares the `c()` path. Affected.

**Proposed fix.** In the `missing(loo_controls)` branch of `expand_manec()`, call
`define_loo_controls()` rather than constructing and validating an empty list.
That is the function whose purpose is to supply the default, and it already
handles both the missing and the partially specified case.

Worth checking in the same pass whether `bnec_group()` and `bnec_hurdle()` reach
`expand_manec()` by a route that supplies `loo_controls`.

**Tests.** A test that the weighting method is the same for a set assembled by
`bnec()` and by `c()` would have caught this, and none of the existing weight
tests compares the two routes. Assert on `attr(mod_stats$wi, "method")` rather
than on the weights themselves, so the test does not depend on sampling.

**Reproduction.**

```r
library(bayesnec)
options(mc.cores = 4)
set.seed(333); data(nec_data)

a <- bnec(y ~ crf(x, model = c("nec3param", "ecxll3")), data = nec_data)
set.seed(333); f1 <- bnec(y ~ crf(x, model = "nec3param"), data = nec_data)
set.seed(333); f2 <- bnec(y ~ crf(x, model = "ecxll3"), data = nec_data)
b <- c(f1, f2)

attr(a$mod_stats$wi, "method")   # "pseudobma"
attr(b$mod_stats$wi, "method")   # NULL
```

</details>

beckyfisher:
## What

The weighting method of a model-averaged set no longer depends on which
function assembled it. A set built with `c()`, `+`, `amend()` or `update()`
from fits that record no method of their own is weighted by pseudo-BMA, the
documented default, as a set fitted in one `bnec()` call already was, and
`attr(mod_stats$wi, "method")` records it. Where the set being operated on does
record a method, that method is kept.

Closes #320.

## Why it matters

Weights determine the model-averaged NEC, NSEC and ECx estimates, so two
analysts fitting the same equations to the same data got different toxicity
estimates depending on whether they called `bnec()` with a model vector or
combined single fits with `c()`. Combining single fits is a documented workflow
and is how the training material introduces model averaging.

The empty `method` attribute meant an affected result could not be identified
from the fitted object after the fact.

## Evidence

`expand_manec()` called `validate_loo_controls()`, which checks the shape of
`loo_controls` and supplies nothing. With no method named, `method` reached
`loo::loo_model_weights()` as `NULL`, `match.arg()` resolved that to `loo`'s own
first choice of `"stacking"`, and the attribute was assigned `NULL`. `bnec()`
was unaffected because it calls `define_loo_controls()` before the model loop.

Measured on the two packaged `manec_example` fits pulled out and recombined
(R 4.6.1, `loo` 2.8.0, no resampling):

| Route | `attr(wi, "method")` | weight on `nec4param` |
|---|---|---|
| `bnec()`, as recorded in `manec_example` | `pseudobma` | 0.827 |
| `c(nec4param, ecx4param)` on `dev` | *(empty)* | 0.892 |
| `c(nec4param, ecx4param)` on this branch | `pseudobma` | 0.821–0.862 |

The third row is a range over twenty repeats. Pseudo-BMA uses a Bayesian
bootstrap and is not deterministic; stacking is, which is why the second row is
a single value. The range brackets the value the `bnec()` call that fitted these
two models recorded, which is the comparison the change is for.

## User-visible changes

Three behaviours differ from `dev`, and one thing this PR deliberately does not
do.

A set assembled by `c()`, `+`, `amend()` or `update()` from fits that record no
weighting method, where the caller names none either, is weighted by pseudo-BMA
rather than stacking, so its model-averaged estimates change. Every set built
before this change is of that kind on these routes, since none of them recorded
a method. Anyone who wants the previous behaviour asks
for it: `amend(fit, loo_controls = list(weights = list(method = "stacking")))`.
A second consequence is that repeated calls no longer agree with each other
unless the session sets a seed, because pseudo-BMA draws a Bayesian bootstrap
where stacking solves a fixed optimisation.

`update()` now keeps the weighting method the set was built with, unless the
caller names another. Refitting a set is not a request to reweight it, and
neither is changing a LOO fitting argument: `loo_controls` names its `fitting`
and `weights` arguments separately, so a call that names only the first names
no method. `amend()` already behaved this way.

`c()` and `+` inherit the method where every input that records one names the
same method, and report a fallback to the default where two disagree. They take
no `loo_controls` argument, so the objects being combined are the only place an
explicit request can come from.

No migration is provided for objects already saved with stacking weights. They
keep the weights they were built with; `amend()` or `update()` on one now
reweights it by pseudo-BMA, which is the correction rather than a further
defect. Re-running such a call is how an affected result is repaired, and the
method the object now records is how it can be identified.

<details>

The default is supplied at `expand_manec()`, on both branches of its
`missing(loo_controls)` test, rather than at each entry point:

```r
fam_tag <- object[[1]]$fit$family$family
loo_controls <- if (missing(loo_controls)) {
  define_loo_controls(family_str = fam_tag)
} else {
  define_loo_controls(loo_controls, fam_tag)
}
```

`expand_manec()` is the only function in the package that calls
`loo_model_weights()`, so it is where the invariant can be stated once. Fixing
the four entry points separately would require four call sites to be right, and
would leave the next one to be added unprotected.

`define_loo_controls()` now tests `is.null(loo_controls$weights$method)` rather
than whether the name is present. A name present with a `NULL` value is not
hypothetical: `pull_out()` and `update()` preserve the method they read off the
object they operate on, which is `NULL` for an object that records none, and
`match.arg()` reads that as stacking. `weights_controls()` builds an empty
`weights` list for an unknown method so the default fills it, and is used by
`amend()`, `pull_out()`, `update()` and `c()`. `amend_model_set()` was reading
the name rather than the value for the same decision and now matches.

`c.bnecfit()` reads the method off its inputs with the existing
`fit_weights_method()` helper, whose documentation in `R/bnec_group.R` is
updated to record its second set of callers. `unique()` over the inputs that
record one: length 1 inherits, length 0 defaults, length greater than 1 defaults
with a message.

`update.bnecfit()` reads the method before `recover_prebayesnecfit()` replaces
`object`, for the same reason the `bnec_record` attribute is read there.
`loo_controls` names its `fitting` and `weights` arguments separately, so the
method is filled in whenever the caller names none, not only where the argument
is absent altogether: without that, changing a LOO fitting argument reweighted
the set as a side effect of asking for something else. `amend_model_set()`
already worked this way and the two now agree.

`bnec_group()` and `bnec_hurdle()` were checked. Neither reaches
`expand_manec()` directly; both fit through `bnec()`, which supplies the
default. `bnec_group()`'s own `wt_method` already defaults to `"pseudobma"`, so
the crossed-weight identity `crossed_group_weights()` checks is unaffected.

Tests assert on the recorded method, not on the weights, which are not
deterministic under pseudo-BMA. `test-bnecfit-methods.R` covers the `c()`, `+`
and `update()` routes, the inheritance of an explicit method, the disagreement
message, and the fitting-only `loo_controls` call; its `update()` tests
substitute `update.brmsfit` with one that returns the stored fit, so they reach
the weighting code without sampling, and restore it through the S3 methods
table for the reason `capture_brms_update()` records. `test-helpers.R` covers
`define_loo_controls()`, `weights_controls()` and `fit_weights_method()`
directly, including the named-but-`NULL` case. `test-amend.R` covers a set that
records no method, which is the shape of any set assembled by `c()` before this
change, and of one whose attribute was dropped by row-subsetting of
`mod_stats`. In `test-expand_classes.R` the existing `expect_null()` on the
missing-argument branch pinned the defect and is the reason it survived; it now
asserts the default.

None of these tests sample. Run with `NOT_CRAN=true`, `test-bnecfit-methods.R`,
`test-expand_classes.R`, `test-amend.R`, `test-helpers.R` and `test-pull_out.R`
pass with no failures.

</details>

https://claude.ai/code/session_01GAm5fjR1ZMJ2WRnvVEuu1e

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: bayesnec
 Title: A Bayesian No-Effect- Concentration (NEC) Algorithm
-Version: 2.1.3.33
+Version: 2.1.3.34
 Authors@R: c(person("Rebecca", "Fisher", email = "r.fisher@aims.gov.au", role = c("aut", "cre"),comment = c(ORCID = "0000-0001-5148-6731")), person("Diego R.","Barneche",role="aut",comment = c(ORCID = "0000-0002-4568-2362")), person("Gerard F.","Ricardo",role="aut",comment = c(ORCID = "0000-0002-2220-3971")), person("David R.","Fox",role="aut",comment = c(ORCID = "0000-0002-3178-7243")))
 Description: Implementation of No-Effect-Concentration estimation that uses 'brms' (see Burkner (2017)<doi:10.18637/jss.v080.i01>; Burkner (2018)<doi:10.32614/RJ-2018-017>; Carpenter 'et al.' (2017)<doi:10.18637/jss.v076.i01> to fit concentration(dose)-response data using Bayesian methods for the purpose of estimating 'ECx' values, but more particularly 'NEC' (see Fox (2010)<doi:10.1016/j.ecoenv.2009.09.012>), 'NSEC' (see Fisher and Fox (2023)<doi:10.1002/etc.5610>), and 'N(S)EC (see Fisher et al. 2023<doi:10.1002/ieam.4809>). A full description of this package can be found in Fisher 'et al.' (2024)<doi:10.18637/jss.v110.i05>. This package expands and supersedes an original version implemented in 'R2jags' (see Su and Yajima (2020)<https://CRAN.R-project.org/package=R2jags>; Fisher et al. (2020)<doi:10.5281/ZENODO.3966864>).
 Depends:
--- a/NEWS.md
+++ b/NEWS.md
@@ -642,12 +642,12 @@
   method at all and reports and ignores one given to it. `amend()` and
   `pull_out()` already preserved the recorded method, but passed an unknown one
   on as `method = NULL`, which `loo` resolves to stacking; an unknown method is
-  now left for the default to fill. `update()`
-  did not preserve it at all and now does, since refitting a set is not a
-  request to reweight it. Because `loo_controls` names its `fitting` and
-  `weights` arguments separately, both `amend()` and `update()` read a call
-  that names only a LOO fitting argument as naming no method, so changing one
-  does not reweight the set as a side effect. `c()` and `+` take no
+  now left for the default to fill. `update()` did not preserve it at all and
+  now does, since refitting a set is not a request to reweight it. Because
+  `loo_controls` names its `fitting` and `weights` arguments separately, both
+  `amend()` and `update()` read a call that names only a LOO fitting argument
+  as naming no method, so changing one does not reweight the set as a side
+  effect. `c()` and `+` take no
   `loo_controls` argument, so they inherit the method where every object being
   combined that records one names the same method, and report the fallback to
   the default where two disagree.
--- a/NEWS.md
+++ b/NEWS.md
@@ -618,6 +618,35 @@
 
 ## Bug fixes
 
+- A model set assembled by `c()`, `+`, `amend()` or `update()` is now weighted
+  by pseudo-BMA, the documented default, rather than by stacking.
+  `expand_manec()` validated the `loo_controls` it was given but supplied no
+  default where the caller named no method, so `method` reached
+  `loo::loo_model_weights()` as `NULL`, `loo` applied its own default of
+  `"stacking"`, and `attr(mod_stats$wi, "method")` recorded nothing. The same
+  set fitted in a single `bnec()` call was weighted by pseudo-BMA, so the
+  weighting method --- and with it the model-averaged NEC, NSEC and ECx
+  estimates --- depended on how the set was assembled rather than on what was
+  requested. Combining single fits with `c()` is a documented workflow and is
+  how the training material introduces model averaging, so this was a common
+  path. Measured on the two packaged `manec_example` fits pulled out and
+  recombined (R 4.6.1, `loo` 2.8.0): stacking placed 0.892 of the weight on
+  `nec4param`, while pseudo-BMA placed between 0.821 and 0.862 over twenty
+  repeats --- it uses a Bayesian bootstrap and so is not deterministic ---
+  against the 0.827 recorded by the `bnec()` call that fitted them. The default
+  is supplied in `expand_manec()`, which is the one point every assembly route
+  reaches (#320).
+
+  Where the set being operated on records a method, that method is kept rather
+  than replaced by the default. `amend()` and `pull_out()` already preserved
+  it, but passed an unknown method on as `method = NULL`, which `loo` resolves
+  to stacking; an unknown method is now left for the default to fill.
+  `update()` did not preserve it at all and now does, since refitting a set is
+  not a request to reweight it. `c()` and `+` take no `loo_controls` argument,
+  so they inherit the method where every object being combined that records one
+  names the same method, and report the fallback to the default where two
+  disagree.
+
 - `dispersion()` no longer discards the statistic where a single observation is
   reproduced exactly. The Pearson denominator is the fitted standard deviation,
   which underflows to exactly zero for a curve that decays fast enough --- `mu
--- a/NEWS.md
+++ b/NEWS.md
@@ -638,9 +638,11 @@
   reaches (#320).
 
   Where the set being operated on records a method, that method is kept unless
-  the caller names another. `amend()` and `pull_out()` already preserved it,
-  but passed an unknown method on as `method = NULL`, which `loo` resolves to
-  stacking; an unknown method is now left for the default to fill. `update()`
+  the caller names another --- `pull_out()` excepted, which takes no weighting
+  method at all and reports and ignores one given to it. `amend()` and
+  `pull_out()` already preserved the recorded method, but passed an unknown one
+  on as `method = NULL`, which `loo` resolves to stacking; an unknown method is
+  now left for the default to fill. `update()`
   did not preserve it at all and now does, since refitting a set is not a
   request to reweight it. Because `loo_controls` names its `fitting` and
   `weights` arguments separately, both `amend()` and `update()` read a call
--- a/NEWS.md
+++ b/NEWS.md
@@ -637,15 +637,18 @@
   is supplied in `expand_manec()`, which is the one point every assembly route
   reaches (#320).
 
-  Where the set being operated on records a method, that method is kept rather
-  than replaced by the default. `amend()` and `pull_out()` already preserved
-  it, but passed an unknown method on as `method = NULL`, which `loo` resolves
-  to stacking; an unknown method is now left for the default to fill.
-  `update()` did not preserve it at all and now does, since refitting a set is
-  not a request to reweight it. `c()` and `+` take no `loo_controls` argument,
-  so they inherit the method where every object being combined that records one
-  names the same method, and report the fallback to the default where two
-  disagree.
+  Where the set being operated on records a method, that method is kept unless
+  the caller names another. `amend()` and `pull_out()` already preserved it,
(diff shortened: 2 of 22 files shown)


beckyfisher:
## Review findings that were defects, and what changed

An independent review of this branch ran over three rounds. Five findings were
raised against the first commit, all of them real, and a sixth defect was
introduced by the documentation fix for one of them and caught in the second
round. This comment records them; the review exchange itself is not reproduced.

**`update()` reweighted a set as a side effect of a LOO fitting argument.**
The first commit applied the recorded weighting method only where
`loo_controls` was absent altogether, so `update(fit, loo_controls =
list(fitting = list(reloo = FALSE)))` discarded it. Measured on a
stacking-weighted set: `dev` returned stacking, the first commit returned
pseudo-BMA. That is a regression introduced by this branch, not merely an
undocumented limit, and it contradicted the PR body, `NEWS.md` and the
`@param` text, all three of which said the method was preserved.
`loo_controls` names its `fitting` and `weights` elements separately, so a call
naming only the first names no method. Fixed in 10c0720f: `update.bnecfit()`
validates a supplied `loo_controls` and fills the recorded method in where none
is named, assigning into the existing `weights` list so another
`loo_model_weights()` argument the caller set survives. `amend_model_set()`
already worked this way and the two now agree.

**`amend_model_set()` and `define_loo_controls()` disagreed about
`method = NULL`.** `amend_model_set()` tested whether the name was present
while `define_loo_controls()` had been changed to test the value, so
`amend(fit, loo_controls = list(weights = list(method = NULL)))` treated the
name as a request two lines before the other function treated the same value as
unspecified. Fixed in 10c0720f; both now test `is.null()`.

**The shared `@param loo_controls` was inaccurate on pages that inherit it.**
This is the one defect the review cycle created rather than found, and it took
two rounds to settle. The parameter text speaks for seven functions at once,
and each attempt to tighten it for one mis-stated another. The first version
said `bnec()` sets the default, which had become false for every other
assembly route. The second said a function operating on an existing set keeps
the method unless the caller names another, which is false for `pull_out()` —
it reports and ignores a method given to it and always keeps the set's own.
The third named `c()` alongside `amend()` and `update()`, where the clause is
vacuous because `c()` takes no `loo_controls` at all. Settled in 644b19f5 and
32267496: the parameter states the default, names `amend()` and `update()` for
preservation, states the `pull_out()` exception, and points at
`?c.bnecfit` for how `c()` and `+` decide. `expand_nec()`, which reads only the
`fitting` element, now has its own parameter rather than inheriting a statement
about weighting it does not do.

**Two documentation sentences described behaviour the code does not have.**
The `c.bnecfit()` details read as though a message is emitted when no input
records a method; it is emitted only where two inputs name different ones. A
test comment described the `dev`-era failure rather than the one its assertion
measures. Both corrected.

**Register, against the repository's prose conventions.** The PR heading
`## What changes for a user` was a second-person what-clause; seven bold
lead-ins were used inside the `<details>` block; the decision not to migrate
saved objects was stated only inside the fold, which the conventions rule out
for a decision; `actually` was used as an intensifier in a roxygen title. All
corrected, and the no-migration decision now sits above the fold with the
remedy stated.

### Findings answered rather than changed

The review noted that the `#216` weighted draw shifts position in the RNG
stream, because pseudo-BMA draws a Bayesian bootstrap where stacking consumed
none. That is a consequence of the change rather than a defect —
`draw_seed` is drawn from the caller's stream by design, so a `set.seed()` in
the session still governs it — and the PR body already states that repeated
calls no longer agree without a seed. No change made.

The review also proposed moving the enumeration of routes out of the shared
`@param` and into `@details` on `bnec()`. Not taken: it would leave `amend`,
`pull_out`, `update` and `expand_manec` with no statement at all in order to
fix one page that does not weight anything. Scoping the sentences and giving
`expand_nec()` its own parameter addresses the same inaccuracy without removing
information from the pages that need it.

### Corrections to my own earlier statements

Three assertions were added to the `update()` test block, not two. Of the
three, `defaulted_fitting` discriminates against `dev` but passes on the first
commit of this branch, since there a supplied `loo_controls` reached
`define_loo_controls()` directly. It is kept as a guard that the new branch
does not pass `method = NULL` down, not as a test of the second commit.

### Verification at the head of the branch

`NOT_CRAN=true` over `test-bnecfit-methods.R`, `test-expand_classes.R`,
`test-amend.R`, `test-helpers.R` and `test-pull_out.R`: 245 assertions, 0
failures, 0 errors, no sampling. All five GitHub Actions jobs passed on the
first commit; the later commits change roxygen comments, generated `man/`
pages, `NEWS.md` and one test comment, with no executable change between
10c0720f and the head.

https://claude.ai/code/session_01GAm5fjR1ZMJ2WRnvVEuu1e

beckyfisher (commented):
This all looks great and resolves a serious existing problem.

PR #323, opened 2026-09-12
example2: a section on fitting the model set in parallel (#322)

### LINKED ISSUE #322: example2: a section on fitting the model set in parallel
## What

A new section at the end of `example2`, *Multi model usage*, on fitting the
model set in parallel: how to turn it on, how many workers to ask for and why
memory rather than cores decides that, what reproduces between a sequential and
a parallel run and what does not, and what the speed-up is on a realistic set.

It goes after *Putting it all together* and before *References*, as a `##`
sibling of *Fitting multiple models and model averaging using the `bnec`
function*.

Most of the section #318 added to `?bnec` moves into it. `?bnec` keeps the
minimum a reader needs at the point of use — set a plan, chains sample in
sequence inside each worker, supply a seed — and points at the vignette for the
rest.

## Why it matters

`example2` is the vignette about fitting a model set, which is exactly the
situation in which parallelism applies. A reader arrives there because they have
more than one model to fit; that is the moment to say the models can be fitted
at the same time.

It already takes a position on parallelism, silently and only for chains.
Line 62 is `options(mc.cores = parallel::detectCores())`, in an `echo = FALSE`
chunk, so the vignette runs its chains in parallel and never says so. The new
section is where that becomes explicit and where the choice between the two
levels — chains in parallel, or models in parallel — gets made in front of the
reader.

And `?bnec` is being asked to do a tutorial's job. The parallel section #318
added is **1,238 words across 138 lines**, in a help file now at **728 lines**
against 151 for `?bnec_hurdle` and 92 for `?amend`. It explains a mechanism,
states a measurement, warns about a development trap and works through process
counts. Someone looking up an argument scrolls past all of it.

There is a second reason to build this, and it is the substantive one. #184's
acceptance criteria include *a measured, documented speed-up on a realistic
multi-model fit, including the compilation cost*. That is the one criterion #318
does not meet, which is why it says "Addresses" rather than "Closes". The
measurement needs a realistic model set, a quiet machine and a rendered record —
which is what a precompiled vignette is, and `example2` already fits the set it
would be measured on.

## Evidence

Measured on `issue-184-future-apply` at `ff22b311`.

`example2` has room. It is the shortest of the worked-example vignettes bar
`example4`:

| vignette | lines | words |
|---|---|---|
| example6 | 1287 | 9233 |
| example1 | 637 | 7510 |
| example2b | 433 | 3845 |
| example3 | 366 | 7454 |
| example5 | 230 | 1052 |
| **example2** | **239** | **2386** |
| example4 | 177 | 1786 |

What would move out of the help page:

| | lines | words |
|---|---|---|
| `man/bnec.Rd`, whole file | 728 | — |
| its *Fitting a model set in parallel* section | 138 | 1,238 |

The roxygen in `R/parallel_models.R` is 188 lines against 110 of code. That
stays: it records why the RNG kind is restored and why the environment is
narrowed, which are the two things a future maintainer would otherwise undo. It
is noted here only because it is the other place these explanations live, and
the vignette should describe the behaviour rather than restate the internals.

<details>

### Where it goes

After *Putting it all together*, before *References*. A `##` heading under
`# Examples`, sibling to *Fitting multiple models and model averaging using the
`bnec` function*. A noun phrase, per the heading convention: *Fitting the model
set in parallel*.

### What moves out of `?bnec`

- the mechanism of the model-averaging draw and why it differs between plans;
- the `mc.cores` measurement and the "belt and braces" reasoning for
  `cores = 1`;
- the single-worker case and which plans reach it;
- the development trap: a `multisession` worker loads the *installed* bayesnec,
  not a `pkgload::load_all()` one;
- worker start-up loading brms, and the OOM-kills-the-batch note;
- the formula-environment and `...`-function export note.

### What stays in `?bnec`

Four sentences and a link: set a plan before the call; there is no `cores` or
`parallel` argument; chains sample in sequence inside each worker unless you
pass `cores`; supply a seed if the run has to be reproducible, and note that the
model-averaged quantities do not reproduce across plans. A reader who never
opens the vignette should still not be surprised by anything.

### Line 62

`options(mc.cores = parallel::detectCores())` in the `echo = FALSE` chunk is the
thing to build the section around rather than to work around. It is the
chains-in-parallel choice, made invisibly. Either surface it where the section
starts, or leave it and explain that the fits above it used it — but do not
leave the vignette demonstrating models-in-parallel while an unmentioned line
sets up chains-in-parallel for everything before it.

Note also that `vignettes/precompile.R` sets `mc.cores` from the allocation and
records that `example2`, `example3` and `example6` override it here. #190 is
named there as the place those overrides get removed; this section is a reason
to look at it sooner for `example2`.

### The measurement

A realistic set — the `decline` group, or the full 23 — fitted sequentially and
then under `plan(multisession, workers = n)` for two or three values of `n`, on
the same seed, reporting wall clock **including compilation** and confirming the
fits are identical. Needs a quiet machine: every timing attempted during #318
was taken while other work was running and none was worth reporting.

Report alongside it the two things that decide whether a plan is worth setting,
both of which were surprises: worker start-up is about three seconds per worker
because each loads brms, and memory rather than cores binds, since a worker
killed by the OS ends the whole call rather than one model of it.

### Precompilation

Under a multi-worker plan bayesnec passes `cores = 1`, so this section's fits
sample chains serially inside each worker while everything above them in the
vignette does not. That is the behaviour being documented and is correct, but it
means this section's contribution to the precompile is not comparable to the
rest and the job should be sized for it.

A timing table is only honest if the render that produced it ran on a quiet
machine, so transcribe the numbers with their conditions stated rather than
recomputing them at every build — the pattern `example7` uses for its case
studies.

### Source material

Nothing needs writing from nothing: the `?bnec` section as it stands at
`ff22b311`, the roxygen in `R/parallel_models.R`, and the `future`/WSL findings
comment on #184, which has the backend-by-backend results and the RNG
measurement.

### Definition of done

`example2` builds and is precompiled; the new section is present with a timing
table stating its machine and conditions; `?bnec`'s parallel section is reduced
to the four sentences plus a link and the help page is shorter than it is now;
and #184 can be closed on it, or the reason it still cannot is written down.

### Not in scope

No behaviour change. #318 is merged and this is documentation only.

</details>

Follows #318. Would close the remaining part of #184.

beckyfisher:
## What

A section at the end of `example2`, *Multi model usage*, on fitting the model
set in parallel: how to turn it on, how many workers to ask for, how to nest
model and chain parallelism, what reproduces between plans and what does not,
and why memory binds before cores do.

Most of the parallel treatment `?bnec` acquired in #318 moves into it. The help
page keeps what a reader needs at the point of use and points at the vignette
for the rest.

## Why it matters

`?bnec` was doing a tutorial's job. Its parallel section was 1,238 words across
138 lines, in a help file of 728 lines against 151 for `?bnec_hurdle`. It is now
262 words in a file of 626, and someone looking up an argument no longer scrolls
past a mechanism, a measurement and a worked process count.

`example2` is where the material belongs because it is the vignette about
fitting a model set, which is the only situation in which any of this applies.
It was also already parallelising, silently: `options(mc.cores =
parallel::detectCores())` sat in an `echo = FALSE` chunk, so every fit in it
sampled chains in parallel without the reader being told. That is now `cores = 4`
on the call the reader already sees, which is what the hidden line did — `brms`
caps chain parallelism at `chains`, and `chains` is 4 here — so no fit changes.

Running the real workload rather than a fixture changed what the section says,
twice over, and both corrections are the point of the PR rather than incidental
to it. The draft asserted that the speed-up begins above four workers; it does
not, for a set this size. And neither backend/plan combination I would have
recommended actually completes this workload on this machine.

## Evidence

Thirteen of the fourteen `"decline"` models, `iter = 2e3`, `chains = 4`,
`cmdstanr`, `plan(multicore)`, each run in a fresh session with the compiled
program cache emptied first, on an otherwise idle 22-core workstation. Runs
during which another R process appeared were discarded and repeated.

| plan | cores in use | runs | median s | range | change |
|---|---|---|---|---|---|
| no plan | 4, one per chain | 2 | 300 | 294-305 | -- |
| `workers = 2` | 2, one per model | 3 | 226 | 216-240 | -24% |
| `workers = 4` | 4, one per model | 2 | 174 | 168-181 | -42% |
| `workers = 8` | 8, one per model | 2 | 150 | 147-153 | -50% |
| `workers = 5`, `cores = 4` | 20, four per model | 2 | 154 | 150-157 | -49% |

Run again with the programs already built, the same five take 99, 114, 108, 101
and 94 s: every plan level with the sequential fit or behind it. Compilation is
what the plan overlaps, and these models sample in seconds.

So the section turns on which run is being timed. A first fit is halved at eight
workers; a repeat gains nothing. That is the useful thing to tell a reader and
it is not what either measurement says on its own.

**rstan cannot compile in forked workers.** They share the parent's `tempdir()`;
thirteen models compiling at once in eight forked workers all failed with
`cannot open the connection` from `stanc`. Two in two workers did not, so a small
trial does not reveal it. A `multisession` worker has its own temporary
directory and is unaffected. Stated as consistent with the shared directory
rather than as an isolated mechanism, and as seen on one machine.

**`multisession` did not complete a fit on this machine**: 0 of 4 against 5 of 5
for `multicore`, three killed at seven minutes with no output and one returning
`error writing to connection` from both workers. The same code passes under
`multisession` on three platforms in CI, so the section gives a check to run
rather than a backend rule. Recorded on #184, where my earlier "not reproduced"
finding is corrected.

<details>

### No refit

The section is added to both `example2.Rmd.orig` and the committed
`example2.Rmd` without re-running the vignette. Every chunk in it is
`eval = FALSE`, so knitting it is a no-op apart from the fence, which knitr
turns from ```` ```{r name, eval = FALSE} ```` into a blank line and ```` ```r ````.
That equivalence was checked by knitting the section standalone and diffing
against the source: the only differences were the two fences.

The two files are generated from one source by a script rather than edited
twice, so they cannot drift.

This matters because a real re-render is not a docs change. `example2.Rmd` was
last rendered on 2026-08-27 and there have been **81 commits to `R/`** since —
the regularizing prior set (#305, #307, #314), the initial-value search (#312),
the estimation resolution (#292). Re-rendering pulls all of it in at once, which
is what #190 exists to do across the whole set. Two consequences are visible
here and are worth knowing:

- the committed weights table lists 8 models; the benchmark run below fits 13;
- `ecxll5` appears in the committed table carrying weight 0.22, and now fails to
  initialise.

The timing table is therefore presented as a separate measured run with its own
seed and conditions, not as something to reconcile against the output above it,
and the section names no model count that the rendered vignette would contradict.

### The one hand-edit to generated output

`example2.Rmd`'s echoed call is hand-synced to match the source. It had already
diverged — it still showed `open_progress = FALSE`, which e721efb9 removed from
the source on 2026-09-10 and left for #190 — and neither `open_progress` nor
`cores` affects the draws, so the committed output remains what this code
produces. The alternative was to leave the new prose referring to a `cores = 4`
the reader cannot see.

### The benchmark

Five configurations, each in a fresh R session, with a discarded warm-up pass
first so that no configuration carries the whole compile cost. Each records the
load and the count of foreign R processes at both ends of its own fit; a run
with anything else on the machine is discarded rather than caveated. Three
earlier attempts were discarded on that basis.

One analysis was computed and thrown away. A per-model extrapolation of where
the time goes gives a negative remainder -- one model alone takes 9.4 s, thirteen
take 101 s -- because a lone fit carries setup a set pays once. What the table
supports is a bound, twenty cores returning six per cent, and that is what the
section claims.

### What stayed in `?bnec`

Set a plan; there is no `cores` or `parallel` argument; under a plan of more than
one worker chains sample in sequence, so a plan pays only where the models are
slow enough for fitting to dominate; supply a seed; and the model-averaged
quantities do not reproduce across plans. A reader who never opens the vignette
should still not be surprised.

That sentence was corrected here too. The trimmed `?bnec` first said "the
speed-up starts above that", carried over from the draft, which the measurements
contradict: eight workers was still behind the sequential fit. It now states the
two figures and points at the vignette.

### Not in scope

No behaviour change. `R/parallel_models.R` is untouched — its roxygen is 188
lines against 110 of code, and it stays there because it records why the RNG
kind is restored and why the applied function's environment is narrowed, which
are the two things a maintainer would otherwise undo.

</details>

Closes #322.


beckyfisher:
## What review changed in this PR

Three rounds. Recording what was actually wrong, since the commits alone do not
say which changes were found rather than intended.

### The benchmark measured the wrong thing, and its conclusion was backwards

The first table warmed the Stan cache once and timed five configurations against
it. That removes compilation --- the cost #184's criterion names, and the largest
thing a plan overlaps --- so it measured parallelism using only the phase
parallelism cannot help. The conclusion drawn from it, that a plan is not worth
setting for a set this size, was the opposite of what a first run does.

Re-measured cold, with the compiled-program cache emptied before every run:

| plan | cores sampling | runs | seconds | range | change |
|---|---|---|---|---|---|
| no plan | 4, one per chain | 2 | 299.5 | 293.7-305.3 | -- |
| `workers = 2` | 2, one per model | 3 | 226.4 | 216.3-240.5 | -24% |
| `workers = 4` | 4, one per model | 2 | 174.2 | 167.8-180.6 | -42% |
| `workers = 8` | 8, one per model | 2 | 149.9 | 146.7-153.2 | -50% |
| `workers = 5`, `cores = 4` | 20, four per model | 2 | 153.5 | 149.7-157.3 | -49% |

A first fit is halved at eight workers. Run again with the programs already
built, the same five take 99, 114, 108, 101 and 94 s and no plan improves on the
sequential fit. The section now turns on which run is being timed, which is more
useful than either half and is not what either measurement says alone.

### One number per configuration, with differences inside the noise

The first table reported n = 1 per row, and every difference in it except one was
smaller than the 6.7 per cent spread I had myself published on #184 for five
replicates of a single configuration. Each row now has two or three clean runs
with its range. Effects of 24 to 50 per cent stand against ranges of 4 to 11. The
one comparison the data cannot support --- eight workers against five-by-four,
whose ranges overlap --- is not attempted.

### A heading that would not have rendered

The reordering left no blank line before `### The Stan backend`. Pandoc's
markdown reader has `blank_before_header` on, so the heading would have rendered
as literal text at the end of the previous paragraph and the section would have
vanished from the document and its table of contents. Found by rendering it, not
by reading it. Every heading in both files is now checked programmatically.

### A sentence that promised the opposite of what the code does

`example2` hid `options(mc.cores = parallel::detectCores())` in an `echo = FALSE`
chunk, so every fit sampled its chains in parallel without saying so. Making that
visible, the first draft hard-coded `cores = 4` --- which `precompile.R` exists to
prevent, since it sets the chain count from the HPC allocation and names example2
as one of three vignettes that override it. Now `cores = getOption("mc.cores", 1)`,
and `precompile.R`'s comment is corrected.

The prose was then left describing the old call. Worse than stale: on a reader's
machine the option is unset, so the fit samples its chains one after another
while the sentence promised they ran together --- and that promise is the premise
the section is built on.

### Claims that were wrong, and one that was absent

- The vignette's rendered output shows eight models with `ecxll5` at the
  second-largest weight; the timing run fits thirteen and `ecxll5` fails to
  initialise. The caption now states that the two do not correspond. The PR body
  had asserted there was no such contradiction, which was the justification for
  not re-rendering.
- A measured disclaimer ("how the remainder divides was not measured") had been
  replaced by an unmeasured mechanism, asserted twice. The disclaimer and the
  parent-side work it referred to are restored.
- The percentages were computed from unrounded medians while the table displayed
  rounded ones, so a reader recomputing them got a different answer. Both now
  print to one decimal and every figure reproduces.
- The `cmdstanr` advice was given unconditionally though its own evidence is that
  the collision is a property of forking; a `multisession` worker has its own
  temporary directory and is unaffected. Now scoped, with cmdstanr's non-CRAN
  status stated.
- The three-second worker start-up was stated for all workers; a forked worker
  inherits the loaded namespace and does not pay it, and the benchmark ran forked.
- **Finding 4 was described as done in a commit message and was not in the tree.**
  A wholesale rewrite had replaced the region the sentence was added to. The
  omission is minor; asserting my own work was complete without checking is not.

### Prose

Eleven ruled words and constructions removed across the rounds: *cost*, *honest*,
*worth understanding*, *touch*, *repays itself*, *actually*, *that is not
hypothetical*, and a competition family --- *beat*, *level with*, *behind*,
*loses* --- that the earlier passes did not cover and which has been added to the
style rules rather than corrected only where it was caught. Two headings were
renamed: one contradicted its own closing paragraph, the other carried a count.
Two directional references pointed the wrong way after reordering and are now
named by section.

### Raised and not changed

A third replicate everywhere. The effects are two to ten times the largest
within-configuration range and the machine is shared; the reviewer agreed this
buys nothing the conclusions need. A pre-precompile check for the heading defect
belongs in its own issue rather than in a documentation PR.

beckyfisher:
I have reviewed this - it looks ok.

PR #326, opened 2026-09-12
Return the control for a draw at or below the NSEC reference (#325)

### LINKED ISSUE #325: nsec() discards the draws whose NSEC is the control concentration, biasing the lower bound of every fit
## What

`crossing_x()` (`R/helpers.R:1528`) locates the NSEC reference by requiring a sign
change in `y - target` along a draw's curve. The NSEC reference is the `sig_val`
quantile of the pooled control posterior, so by construction `sig_val` of the
draws have a control prediction at or below it. Those draws start below the target,
no sign change exists, and `crossing_x()` returns `NA`. The summary is then taken
with `na.rm = TRUE`, so they are dropped.

Fisher and Fox (2023) report those draws as the estimate, not as a failure. The
NSEC for such a draw is zero concentration, which in `bayesnec` is the control —
the lowest concentration in the data, which the package already assumes and reads
the reference at (`control_posterior()`, `R/bnec.R:23`, `R/define_prior.R:374`).
Because many designs add a small value to a zero control so the predictor can be
modelled on a log scale, the control concentration rather than the literal zero is
the value in every case, and a floored zero needs no special handling.

## Why it matters

Every `nsec()` call and every `ecx`-class equation's stored no-effect estimate
discards `sig_val × n_draws` draws from the lower tail, in every fit, whatever the
data are doing. The reported lower credible bound is therefore too high, which is
the non-conservative direction for a protective concentration, and the published
behaviour of the method cannot be reproduced: Fisher and Fox (2023) state that the
lower bound is zero for any `sig_val` above the quantile used for that bound
(typically 0.025), and `bayesnec` returns a positive bound at every `sig_val`.

The affected estimate reaches `summary()`, model averaging, and `nec()` on an
`ecx`-class fit, through `expand_nec()`. It is not confined to direct `nsec()`
calls.

The message a user currently sees is also the reverse of what has happened. On the
`example9` workflow fit it reads: "The fitted ecxll3 curve does not fall to the
control's 0.01 quantile within the predictor range for 80 of 8000 draws. Those
draws are excluded from the NSEC summary, which is therefore censored above the
highest concentration tested." Those 80 draws fall to zero. They reach the
reference at the control, which is the lowest concentration tested, not above the
highest.

## Evidence

Measured on `cache/logworkflow/fit_log_regularizing.rds`, the `example9` workflow
fit (`suc | trials(tot) ~ crf(log(dose_adj), model = "decline")`, 8000 draws),
equations pulled from the model-averaged set with `pull_out()`. Grid resolution
1000 for the first table and 200 for the second. The lowest predictor value in the
data is 0.005, the floored zero control.

Every dropped draw reaches the reference, and none fails to reach it:

| equation | draws returning `NA` | of those, at or below the reference at the control | of those, never reaching the reference |
|---|---|---|---|
| `ecxll3` | 80 | 80 | 0 |
| `ecxwb1` | 80 | 80 | 0 |
| `ecx4param` | 80 | 80 | 0 |
| `nec3param` | 79 | 79 | 0 |

For all 80 `ecxll3` draws the curve minimum is 0.9145 below the reference, the
reference being 0.9145 and the curve reaching zero.

The count is `sig_val × n_draws` at every significance level, and the lower bound
never reaches the control (`ecxll3`, resolution 200):

| `sig_val` | draws returning `NA` | reported 2.5% bound | 2.5% bound with those draws at the control |
|---|---|---|---|
| 0.01 | 80 (1.0%) | 0.983 | 0.918 |
| 0.05 | 400 (5.0%) | 0.883 | 0.005 |
| 0.10 | 800 (10.0%) | 0.829 | 0.005 |
| 0.20 | 1600 (20.0%) | 0.761 | 0.005 |

The second column of the right-hand table reproduces the structure of Fisher and
Fox (2023) Table 3, whose Bayesian MCMC row is 10.5 (6.15–13.6) at 0.01, 8.78
(0–12.1) at 0.05, 7.84 (0–11.5) at 0.10 and 6.51 (0–10.7) at 0.20. The zeros are
this class of draws. The paper states the reason twice: at p. 2026, that the lower
bound "will be 0 for any significance level greater than the 0.025 quantile"; and
in the discussion, that such an outcome "potentially validly reflects the
underlying fact that the confidence bounds of a true no-effect value may in fact
contain 0 because, for smooth curves, a decline in the response may occur at the
lowest concentration".

The paper obtains every draw's NSEC by backward interpolation, solving the inverse
of the fitted equation analytically, with no crossing search and no exclusion
step. Its printed code, `(-log(p/b0)/b1)^(1/b2)`, does not itself return zero for
such a draw: the base is negative where `b0 < p` and a fractional exponent of it
is `NaN` in R, exactly zero arising only at `b0 == p`. What the paper reports for
those draws is zero, and the text above says so twice.

<details>

### Where the exclusion occurs

- `R/helpers.R:1528`, `crossing_x()` — `min(zero_crossings(y - target))` is not
  finite when no sign change exists, and `NA_real_` is returned.
- `R/nsec.R:184`, `nsec.bayesnecfit()`; `R/nsec.R:371`, `:387`, `:502`, `:532`,
  the `bayesmanecfit` and `brmsfit` methods.
- `R/expand_classes.R:63`, `nsec_off_curve()` inside `expand_nec()`, which
  produces the stored `ne` posterior for every `ecx`-class equation. This is the
  path that reaches `summary()`, model averaging and `nec()`.
- `R/bayesnechurdlefit-methods.R:263`, the hurdle method.
- `ecx()` is affected only for `type = "direct"`. For `absolute`, `relative` and
  `range` the target is derived per draw from that draw's own control
  (`ecx_from_posterior()`, `R/ecx.R:466`), so the curve always starts above it.

### Provenance

The crossing search entered on `batch-1-ecx-definition` and reached `dev` in
PR #281, merged 2026-09-07 at `212ff649`. Two commits in that merge are
responsible: `ce2aa236`, which added the crossing search in `crossing_x()` and the
`nsec()` warning, and `027b9f9d`, which added `nsec_off_curve()` in `expand_nec()`
and the message quoted above. PR #281 closes #39, which asked for the precision
improvement that the crossing search provides. The other open pull requests whose
history contains those commits — #228, #238, #243 and #323 — branched from `dev`
after the merge and did not introduce the behaviour.

In 2.1.3 `nsec_fct()` (`R/nsec.R:221`) used `min_abs(y - reference)`, the nearest
grid point, which for this class of draws is the control. The regression is the
change from nearest-grid-point to crossing detection, not the crossing detection
itself. Measured in the first comment below.

### D15 ruling 3

Ruling 3 reads: "A target the curve never reaches within the predictor range
returns `NA` with a warning. Today the nearest grid point is returned, which for a
curve that never declines to the target can report the control concentration
itself as the ECx."

That is correct for ECx, where the target is per draw and below that draw's own
control, so a returned control concentration is spurious. It does not transfer to
the NSEC, where the reference is a pooled quantile of the control posterior and the
control concentration is the correct answer for the draws in question. The ruling
needs restating so that it applies to ECx and to the right-hand case only.

### Proposed change for the measured defect

Where a draw's curve is at or below the reference at the control, return the
control concentration — the same predictor value the reference is read at, not
`x_vec[1]` and not zero. Tying it to the control's x rather than to the grid edge
keeps `x_range` from changing the estimate, consistent with D15 ruling 2. Emit
nothing for those draws: they are `sig_val × n_draws` of every fit by
construction, and a message reporting them restates the definition of the
quantile.

`crossing_x()` currently sees only `y`, `target` and `x_vec`, so either it is
given the control's x or the clamp is applied in the callers that hold it.

### The draws that do not reach the reference within the range

This case is unresolved. A draw whose curve stays above the reference across the whole range would also
return `NA`, and this is the case the present message was written for. Whether it
occurs in a fit anyone would report is not established. The only demonstration
obtained was artificial — the prediction grid of the `example9` fit was truncated,
which is not a design that would have fitted, and RF notes that with so short a
range there would be no curve to draw. In the four equations measured over the
full range the count is zero.

The arithmetic condition is that the decline over the tested range is smaller than
the gap between the control posterior's centre and its lower `sig_val` tail; on the
`ecxll3` fit that gap is 0.0433, against a decline of 0.958 over the full range.
The realistic route to it is a non-responsive test rather than a short range, and a
real example is needed before deciding anything.

Two questions to settle if it is real. Whether the value should be the highest
tested concentration, making the estimate a lower bound and matching the
conventional report of a result greater than the highest concentration tested, or
`NA`; the paper's analytic inversion is unbounded above and would return an
extrapolated concentration, which a grid search cannot. And what the summary should
do: `quantile(..., na.rm = TRUE)` currently reports a value computed from whichever
draws crossed, with a message rather than a refusal, and a quantile over a subset
selected that way is not an NSEC.

### Tests and documentation

- A test pinning the structure of Fisher and Fox (2023) Table 3: the lower bound
  sits at the control for `sig_val` of 0.05, 0.10 and 0.20, and above it at 0.01.
- A test that the count of draws at the control is `sig_val × n_draws`.
- `tests/testthat/test-helpers.R:204` pins only the right-hand case for
  `crossing_x()` and needs an assertion for the left-hand one.
- `tests/testthat/test-nsec.R:137` produces the right-hand case by restricting
  `x_range`; it stays valid as a test of the mechanism but should not be read as
  evidence that the case arises in practice.
- Wording corrections at `R/nsec.R:47`, `R/expand_classes.R:73` and
  `R/nsec.R:197`, all of which describe the affected draws as not reaching the
  reference.
- `ecnsec` is computed over all draws while the NSEC is computed over the
  surviving subset, so the two are drawn from different sets of draws. This
  resolves itself once no draw is dropped, but it should be checked.

</details>

beckyfisher:
Closes #325.

## What

A draw whose curve is at or below the NSEC reference where the search begins now
returns the control concentration, which is its estimate, instead of `NA`. The
reference is the `sig_val` quantile of the control posterior, so `sig_val` of the
draws are in that class in every fit; they were discarded, and the message
reporting them described the opposite case.

## Why it matters

Every `nsec()` call and every `ecx`-class equation's stored no-effect estimate
dropped `sig_val × n_draws` draws from the lower tail, whatever the data were
doing, so the reported lower credible bound was too high — the non-conservative
direction for a protective concentration. Fisher and Fox (2023) report those
draws as the estimate: their Table 3 gives a lower credible bound of zero at every
significance level above the 0.025 quantile the bound is read at, and `bayesnec`
could not produce that at any `sig_val`.

`bayesnec` treats the lowest observed concentration as the control, so the
paper's zero concentration is that value, and a design that adds a small value to
a zero control so the predictor can be modelled on a log scale needs no separate
case: the flooring value is the control.

## Evidence

Measured on `cache/logworkflow/fit_log_regularizing.rds`, the `example9` workflow
fit (`suc | trials(tot) ~ crf(log(dose_adj), model = "decline")`, 8000 draws),
equations pulled from the model-averaged set. Every discarded draw reached the
reference and none failed to — 80 of 80 on `ecxll3`, `ecxwb1` and `ecx4param`, 79
of 79 on `nec3param` — and the 2.5% bound could not reach the control:

| `sig_val` | draws discarded | 2.5% bound before | 2.5% bound after |
|---|---|---|---|
| 0.01 | 80 (1.0%) | 0.983 | 0.918 |
| 0.05 | 400 (5.0%) | 0.883 | 0.005 |
| 0.10 | 800 (10.0%) | 0.829 | 0.005 |
| 0.20 | 1600 (20.0%) | 0.761 | 0.005 |

The control is 0.005. The right-hand column reproduces the structure of Fisher and
Fox (2023) Table 3, whose Bayesian MCMC row is 10.5 (6.15–13.6) at 0.01, 8.78
(0–12.1) at 0.05, 7.84 (0–11.5) at 0.10 and 6.51 (0–10.7) at 0.20.

Master's estimator, `x_vec[min_abs(y - reference)]`, applied to the same
posterior, places exactly `sig_val × n_draws` draws at the control and discards
none, so the defect was introduced in this development cycle and never released.
It entered in #281 at `ce2aa236` and `027b9f9d`. Master is not correct either: it
snaps to the nearer grid point instead of interpolating, and a draw that never
reaches the reference returns the highest tested concentration with nothing said.
This pull request keeps the interpolated crossing from #39 and restores master's
treatment of the boundaries.

<details>

### The change

- **`crossing_x()` takes `x_start`** (`R/helpers.R`), the value returned where the
  curve has already reached the target at the first grid point. A search for a
  sign change cannot tell that case from a curve that never reaches the target,
  and the two are opposite statements about the estimate, so the caller decides
  rather than `zero_crossings()`. It defaults to `NA_real_`, which leaves **every
  ECx that a user can reach through `ecx()`'s default types unchanged**: an ECx
  target is derived from the draw's own control and cannot be met before the grid
  begins.

  One ECx case does change: a curve that has already reached its target where the
  grid begins. Three routes reach it, all degenerate. `type = "direct"` with a
  supplied target above the curve at the control; `type = "range"` where the
  curve's lowest predicted response is at the control, which makes the target
  equal to it exactly; and `type = "relative"` on a hormetic equation for a draw
  whose `bot` exceeds its control. Each now returns `NA`. Where the curve is
  hormetic the value returned before was the crossing on the *rising* limb, the
  concentration at which the response reaches the target on the way up, which
  estimates nothing; otherwise it was the lowest concentration in the series.
  `test-ecx.R` pins the hormetic case through the internal estimator, which is
  where a mismatched control can be supplied.

- **`nsec_from_posterior()` is added** and used by `nsec.bayesnecfit()`,
  both branches of `nsec.brmsfit()`, `nsec_off_curve()` in `expand_nec()` and
  `nsec.bayesnechurdlefit()`. Which draws take the control is read from the
  **control posterior** — the draw's own control against the reference, the same
  vector the reference is a quantile of — rather than from the curve at the first
  grid point. That is the definition, and it makes the value independent of
  `x_range`, which D15 ruling 2 requires. A draw whose control is above the
  reference but whose curve is already below it where a supplied `x_range` begins
  reached the reference below the range asked for: that is not identified within
  the range, so it is `NA` and is reported with the draws that never reach the
  reference, and is reported by a warning of its own rather than by the one about
  curves that never reach it. The search runs from the control upward, the
  control itself being made the first point of the searched grid — with the
  control posterior supplying the predictions there — so no estimate is placed
  below the lowest tested concentration and none is lost between the control and
  the first grid point above it. Where a grid reaches no concentration above the
  control the call is refused by name — an `x_range` at or below the lowest
  observed value, or `resolution = 1` — and `bnec()` refuses a `resolution`
  below 2 up front rather than letting it arrive from `expand_nec()` after every
  model has sampled. The two counted attributes the warnings are built from are
  stripped before the posterior is returned.

- **`nsec.drc()` passes `min(x_vec)`.** Its reference is the control's lower
  confidence limit at `level = 1 - 2 * sig_val` while the curves are drawn at
  `prob_vals[3] - prob_vals[2]`, so the lower curve begins at or below the
  reference for any `sig_val` at or above half the excluded probability — 0.025
  under the default `prob_vals` — and returned `NA` there. Measured on
  `drm(y ~ x, nec_data, LL.3())`: reference 0.88331 against a lower curve
  beginning at 0.88557 at `sig_val = 0.01`, and equality at 0.025. `drc` is not a
  dependency, so no test covers this method, and the interval this construction
  builds is not from Fisher and Fox (2023), whose frequentist NSEC is the
  fitted-mean crossing alone.

- **`control_x()` is added and `hurdle_control_x()` removed.** The latter was the
  same computation for a hurdle fit; `control_x()` resolves a hurdle object to its
  survival component, matching `hurdle_component_preds()`, and its two call sites
  now use it.

- **Nothing is reported for the draws at the control.** They are `sig_val` of the
  draws of every fit by construction, so a message about them restates the
  definition of the quantile. The warning in `nsec()` and the message in
  `expand_nec()` are unchanged in wording and now fire only for a draw that does
  not reach the reference at any tested concentration.

### Scope of the defect

`nsec()` on every class, and `expand_nec()`, which writes the stored no-effect
posterior of every `ecx`-class equation — the estimate `summary()`, the model
weights and `nec()` report, so a fit `nsec()` was never called on was affected.

### D15 ruling 3, corrected in D17

Recorded in `notes/implementation/03_decisions.md`. The ruling's rationale states
that the nearest-grid-point search "for a curve that never declines to the target
can report the control concentration itself as the ECx". Measured on
`x = seq(0, 10, length.out = 11)` and `y = 10 - x`,
`which.min(abs(y - target))` for a target below the whole curve returns the
*highest* concentration. The nearest grid point is the control only where the
target lies above the curve at the first grid point, which for an ECx requires
one of the three degenerate routes above and for an NSEC is the lower `sig_val`
tail — where the control is the correct answer rather than a defect. So the
ruling stands for the ECx and is corrected for the NSEC. A related wording error
is corrected in `NEWS.md` and in a test comment, which called the
nearest-grid-point value an interpolated estimate.

### Tests

- `test-nsec.R`: that an `x_range` beginning above the control reports the control
  only for the `sig_val` tail and `NA` for the rest, under a warning naming that
  case rather than the other; that extending `x_range` below the data or far
  above it leaves the estimates and the count at the control unchanged; and that
  a range reaching no concentration above the control, and a `resolution` below
  2, are both refused by name.
- `test-nsec.R`: the Table 3 structure on `ecx4param` — the 2.5% bound at the
  control for `sig_val` of 0.05, 0.10 and 0.20 and above it at 0.01; the count at
  the control is `sig_val × n_draws`; no warning and no message for them.
- `test-nsec.R`: `nsec_from_posterior()` on a constructed matrix of 100 declining
  curves with a known answer, including that a grid extended below the control
  and a grid beginning above it both leave the same five draws at the control,
  and that a reference no curve reaches is still `NA`.
- `test-helpers.R`: `x_start` returned where the series begins at or below the
  target, including exact equality; the default `NA` preserved; `x_start` not
  reaching the never-crossing case; a one-point series returning `NA` rather than
  stopping, and a two-point series still interpolating.
- The existing `test-nsec.R` case for a curve that never reaches the reference is
  unchanged and still warns.

### Not addressed here

`ecx()` reports a draw that reached its target below a supplied `x_range` with
the warning written for curves that never reach it, which is the conflation this
splits for `nsec()`. `warn_censored_draws()` already takes the arguments to fix
it. Left out because #325 is about the NSEC; recorded on the issue.

Whether a draw that never reaches the reference should instead return the highest
tested concentration, making the estimate a lower bound in the conventional sense,
and whether a summary should be reported at all when most draws are in that class
— `quantile(..., na.rm = TRUE)` reports a value from whichever draws crossed. No
fit in which that class is non-empty over the full tested range has been measured.
Recorded as open in #325 and in D17.

`expand_nec()` reads its reference at the first column of the prediction grid
rather than at the lowest observed concentration, so it does not honour D15
ruling 2 when `bnec()` is given an `x_range`. The reference and the start of the
search are the same point there, which is what this fix requires; the ruling-2
gap is pre-existing and separate.

The `example9` vignette output in #238 quotes the message this changes, and the
precompiled vignettes are stale for other reasons already tracked in #190.

</details>

https://claude.ai/code/session_01NrmuPSnzqBnGpAcU2qbHsQ


beckyfisher:
## Defects found in review, and what was changed

The pull request was reviewed over four rounds by an independent agent working
from the branch rather than the diff, running the estimators and extracting the
paper. Fourteen defects were found: ten in the first version of the change and
four in the fixes for those. All are closed. The items below are the ones that
were real; everything else the review raised was a correction to the record or
to wording, listed at the end for completeness.

### Two defects that would have shipped a wrong estimate

**A draw below the reference at the first grid point was reported as the
control.** The first version tested the curve where the search began instead of
the draw's own control against the reference. Where `x_range` begins above the
lowest observed concentration those are different sets, and much larger:
`nsec(ecx4param, sig_val = 0.01, resolution = 200, x_range = c(1, 3))` reported
65 of 100 draws at 0.0323, a concentration below the whole grid, where 64 of them
have a median NSEC of 0.877. The summary changed from 0.405/0.934/1.141 to
0.0323/0.0323/1.141, with no warning. The test is now on the control posterior,
the vector the reference is a quantile of, so exactly the `sig_val` tail takes
the control and the value does not depend on `x_range`. The other draws are `NA`
and are reported.

**Restricting the search to concentrations at or above the control lost
identified draws.** Dropping the grid points below the control began the search
one grid step above it, so a draw crossing inside that gap was discarded.
`nsec(ecx4param, sig_val = 0.05, resolution = 200, x_range = c(0, 100))` returned
11 `NA` where the observed range returns none; their estimates over that range
are 0.064 to 0.484 against a first searched point of 0.503. That is `x_range`
changing the estimate, which D15 ruling 2 forbids. The control is now made the
first point of the searched grid, its predictions taken from the control
posterior, so the gap does not exist. Verified across `x_range` of `NA`,
`c(0, 3.22)`, `c(0, 10)`, `c(0, 50)` and `c(0, 100)`: no `NA`, and the same draws
at the control in each.

### Three defects in reporting and in refusals

**The warning stated the opposite of what happened to the draws it counted.**
A draw that reached the reference below a supplied `x_range` was reported by the
message written for curves that never reach it — the same conflation this issue
was opened for. The two classes are now counted separately and have a message
each, and a call producing both raises them in the same order on the single-fit
and hurdle paths.

**A grid with no concentration above the control crashed or was summarised from
the remainder.** A single retained column stopped inside
`modelbased::zero_crossings()` with "`lower` is not smaller than `upper`"; none
returned a vector of `NA` under a warning naming the wrong cause; and
`resolution = 1` produced a summary computed from whichever draw needed no
search. Such a grid is now refused by name, and `bnec()`, `amend()` and
`update()` refuse a `resolution` below 2 before any fitting — `update()` refits,
so from `expand_nec()` the refusal arrived after the work was done.

**Internal counters reached the caller.** The two attributes the warnings are
built from survived `sub_x_transformation()` onto the returned posterior, and
only for a `bayesnecfit`, so `posterior = TRUE` carried different attributes
depending on the class of its argument.

### Corrections to the record

Four statements in the first version were wrong and are corrected in `NEWS.md`,
the roxygen and D17: the measurement supporting the correction to D15 ruling 3
was printed as `seq(0, 10, 11)`, which is `by = 11` and reproduces the claim it
corrects; the justification given for the `nsec.drc()` change was false at the
default `sig_val`, the lower confidence curve beginning on the reference only
from 0.025; Fisher and Fox (2023) were said to exclude no draw, where their
printed `NSEC_vals` returns `NaN` for exactly the draws at issue and it is the
reported result and the text at p. 2026 that support the behaviour; and the
guarantee about `x_range` was stated over the estimates when it holds over which
draws are identified, their values moving with the grid spacing as any
interpolated estimate does.

Two further departures from the paper are now documented rather than implicit:
no extrapolation above the tested range, and the lowest observed concentration
rather than zero below it.

### Tests

A test asserting that the comparison posteriors contain censored draws was
passing on a three-draw margin that a regenerated fixture would break; it now
measures its precondition and skips with a reason. A test pinning the old ECx
value reached the case through an internal call rather than the documented route,
which its comment now states. Regression tests were added for each defect above.

### Left for later

`ecx()` still reports a draw that reached its target below a supplied `x_range`
with the warning written for curves that never reach it. Splitting it means
deciding what an ECx below the requested range should return, which is the
question D17 leaves open at the other end, so it is recorded here rather than
answered in this pull request.

beckyfisher:
This fixes a serious issue introduced when the interpolation was added.

Issue #324, opened 2026-09-12
Add an accessor for the highest-weighted model of a bayesmanecfit


beckyfisher:
## What

Add an accessor that returns the highest-weighted model of a `bayesmanecfit` as a
`bayesnecfit`, without the caller having to reach into `mod_stats` or to handle
the case where the object is already a `bayesnecfit`.

## Why it matters

Extracting the best-weighted model is a routine step: it is the candidate that
contributes most to the model-averaged estimate, so it is the one whose fit is
worth checking with `pp_check()` and `check_fit()`. There is no accessor for it,
so every caller writes the same block, and every caller has to know two things
that are not obvious.

The first is that the weights live in `mod_stats` and `pull_out()` takes a model
name, so the name has to be found before it can be passed. The second is that
`screen_models()` returns a `bayesnecfit` rather than a one-model
`bayesmanecfit` when the screen leaves a single equation, so `mod_stats` is
absent and code that reads it errors. That second case is documented behaviour
rather than an edge case, but it means the obvious one-line version is wrong.

`vignette("example9")` on #238 currently carries this, with a comment explaining
the class check:

```r
best_fit <- if (inherits(fit_screened, "bayesmanecfit")) {
  ms <- fit_screened$mod_stats
  pull_out(fit_screened, model = ms$model[which.max(ms$wi)])
} else {
  fit_screened
}
```

Nine lines and a comment, in a vignette whose subject is the workflow rather than
the object model.

## Evidence

`pull_out()` (`R/pull_out.R`) takes `model` as a name and has no option to select
on weight. `mod_stats` is the only route to the weights, and it exists on
`bayesmanecfit` alone. `?screen_models` records that a screened set reduced to
one model is returned as a `bayesnecfit`.

<details>
<summary>Implementation detail</summary>

Two shapes are possible and the choice is open:

- a new exported function, `pull_best(object)`, or
- an argument on the existing one, `pull_out(object, best = TRUE)`, mutually
  exclusive with `model`.

Either way the behaviour needed is: on a `bayesmanecfit`, return
`pull_out(object, model = mod_stats$model[which.max(mod_stats$wi)])`; on a
`bayesnecfit`, return the object unchanged, so that a caller need not test the
class. That second branch is what makes it worth having.

Points to settle before writing it:

- what to do when two equations tie on weight, which is unlikely but not
  impossible on a small set;
- whether to warn when the highest weight is small in absolute terms, since the
[...]

PR #330, opened 2026-09-13
Add pull_best(), an accessor for the highest-weighted model of a set

beckyfisher:
## What

Adds `pull_best()`, which returns the highest-weighted candidate of a
`bayesmanecfit` as a `bayesnecfit`, and returns a `bayesnecfit` unchanged.

## Why it matters

The highest-weighted candidate contributes most to the model-averaged estimate,
so its fit is the one normally inspected with `pp_check()` and
`check_fit()`. Selecting it required reading the weights out of `mod_stats` and
passing the name to `pull_out()`, under a class test, because `screen_models()`
returns a `bayesnecfit` whenever the screen leaves one equation and such an
object has no `mod_stats` to read. Nine lines, written out in every caller:

```r
best_fit <- if (inherits(fit_screened, "bayesmanecfit")) {
  ms <- fit_screened$mod_stats
  pull_out(fit_screened, model = ms$model[which.max(ms$wi)])
} else {
  fit_screened
}
```

becomes `best_fit <- pull_best(fit_screened)`.

The vignette half of #324 is not done here and the issue stays open for it:
`vignettes/example9.Rmd.orig` does not exist on `dev` and arrives with #238,
whose branch holds the block above and the paragraph explaining the class test.
Both should be cut there once this merges.

## Evidence

`pull_out()` takes `model` as a name and has no option to select on weight
(`R/pull_out.R`). `mod_stats` is the only route to the weights and exists on
`bayesmanecfit` alone; `?screen_models` records that a screened set reduced to
one model is returned as a `bayesnecfit`.

The new tests pass under `NOT_CRAN=true` and add roughly five seconds to the
suite: nothing is fitted, the refusal and weights-table branches are tested on
fabricated input, and three calls to `pull_out()` on the stored `manec_example`
account for all of the time. `R CMD check --no-tests` reports no new WARNING or NOTE; the
one NOTE is the pre-existing `modifyList` import.

<details>

### Shape

A new exported generic, not `pull_out(object, best = TRUE)`. `pull_out()`'s
first argument is named `manec` and is documented as a `bayesmanecfit`; the
argument form would have required it to accept a `bayesnecfit` as well, and
would have added a pair of mutually exclusive arguments to a function that
already has four branches of name-matching messages. The separate function
leaves `pull_out()` untouched apart from a `@seealso` line.

`R/pull_best.R` holds the generic and four methods:

- `bayesmanecfit`: names the model with the internal `best_weighted_model()`
  and delegates to `pull_out()`, so the returned object is rebuilt exactly as a
  hand-written `pull_out()` call would build it, and the record of failed models is
  copied onto it by `pull_out()`'s existing code rather than by a second
  implementation.
- `bayesnecfit`: returns the object, and reports which single model it holds.
  The pass-through is the reason the function exists, so it is not a usage
  error; it is reported on `screen_models()`'s principle that the record is the
  point, since a workflow that names the selected equation on one branch and
  says nothing on the other records the choice only when the set happened to
  hold more than one equation.
- `bayesnechurdlefit`: selects within each component and rewraps with
  `hurdle_rewrap()`, as `screen_models()` and `amend.bayesnechurdlefit()`
  already do. The two components hold different candidate sets --- survival is
  0-1 bounded and growth need not be --- so they may end on different equations.
- `default`: an error naming the three classes, and naming `bayesnecgroupfit`
  explicitly, since that class holds one fit per level in `fits` and neither
  `amend()` nor `screen_models()` has a method for it either.

The object is the only argument. Anything in `...` is refused, in the generic
and before dispatch, so that the refusal cannot itself depend on the class and
cannot arrive after the selection has been reported. `x_range`, `resolution`,
`sig_val` and `loo_controls` re-specify a fit rather than select one: honouring
them on the `bayesmanecfit` branch, where `pull_out()` rebuilds the fit, and
ignoring them on the `bayesnecfit` branch, where nothing is rebuilt, would make
the returned object depend on the class the caller was told not to test for.
`pull_out()` takes them alongside a model name and `amend()` rebuilds an
existing fit. `model` is refused by the same guard, which otherwise reached
`pull_out()` as a second value for an argument this function sets itself and was
reported by base R, several frames away, as "matched by multiple actual
arguments".

#### Ties

Detected by exact equality, reported by `message()`, and resolved to
the first candidate in the order of the set, which is the order `expand_manec()`
built `mod_stats` in and is stable across calls. Only exact ties are detected:
two weights differing in the last decimal place are not a tie and the larger is
taken.

#### A small highest weight

Reported, not warned against. Every absolute
threshold is wrong for some set size --- the highest weight of two candidates is
at least 0.5 and of twenty at least 0.05 --- so the message gives both numbers,
`"Highest-weighted model is nec4param, holding 0.827 of the weight of 2
candidate models."`, and the user compares them against the equal share. A user
who needs more has `summary()`, which prints the whole weights table.

#### Hurdle fits

Given a route, per component, as above.

### Further refusals

`best_weighted_model()` raises an error where the weights table is absent, empty
or without either of the two columns it reads, and where no weight is finite;
the second names `amend()` and `summary(x)$mod_weights`. None is reachable
through the public API today, and the alternative is a failure several frames
from its cause: `which.max()` on an all-`NA` vector returns `integer(0)`, and a
missing `model` column reports `NA` as the name and then reaches `pull_out()`,
which answers that `NA` is not in the set and returns the object it was given.

The model name is read from the `model` column of `mod_stats` rather than from
its row names. Only the column is set by name:
`data.frame(model = success_models)` in `expand_manec()` takes automatic row
names, since `success_models` is unnamed, and the model names reach the row
names only at `cbind(mod_stats, disp)`, from the row names of the dispersion
matrix. A change to the shape of that matrix, or to how those columns are
attached, would drop them without touching the column or raising anything.
`fit_best_waic()` in `R/bnec_group.R` reads the row names for the same purpose
and is left alone: it reads a stored per-level fit's own `mod_stats`, which has
that `cbind()` applied, so it is fragile rather than wrong, and correcting it
belongs with the group path rather than with this accessor.

### Tests

`tests/testthat/test-pull_best.R`, following the `test-<filename>.R` convention
rather than the issue's suggestion of `test-pull_out.R`, since the code is in
`R/pull_best.R`. Covers: selection and its message; an exact tie and a near-tie;
a weights table that is absent, empty, or missing either column it reads;
weights that are not finite; the `bayesmanecfit` case against
`pull_out(manec_example, model = "nec4param")` computed in `setup.R`; the
`bayesnecfit` pass-through and the model it names; the refusal of a further
argument on both branches, on an unusable object, and before the selection is
reported; the `default` method on two classes; and a hurdle fit whose two
components are both model sets, weighted opposite ways, so that each has to be
selected from on its own to reach the equation asserted.

The set screened to one model, which the issue asks for, is the pass-through
test: `screen_models()` returns that object as a `bayesnecfit`, and running the
screen itself would test `screen_models()` rather than this function.

</details>

https://claude.ai/code/session_0192FF4nsjtuGufrtat1bygq


beckyfisher:
## Review pass

An independent review ran over this branch in three rounds. Five findings were
real and are fixed in the commits above; the rest were answered without a
change. The findings, in the order they matter:

1. **`...` was honoured on one branch and dropped on the other.**
   `pull_best(fit, resolution = 50)` rebuilt the fit at that resolution where
   `fit` was a `bayesmanecfit`, since the argument reached `pull_out()`, and
   ignored it where `fit` was a `bayesnecfit`, since nothing is rebuilt there.
   The returned object therefore depended on the class the function exists to
   stop the caller testing for. The object is now the only argument, and
   anything in `...` is refused in the generic, before dispatch. `model = ` is
   refused by the same guard: it previously reached `pull_out()` as a second
   value for an argument this function sets itself and failed as base R's
   "matched by multiple actual arguments", several frames away and after the
   selection had been reported.

2. **The pass-through was silent while the model-set branch reported its
   selection.** A workflow recorded which equation was used only where the set
   held more than one, which is the asymmetry the function is meant to remove.
   Both branches now report, and the hurdle method labels its two components so
   the two reports can be told apart.

3. **The hurdle test's survival half was vacuous.** The component was a
   `bayesnecfit` to begin with, so the assertion on it passed whether or not the
   method selected within it. Both components are now model sets weighted
   opposite ways, and each has to be selected from on its own to reach the
   equation asserted.

4. **`best_weighted_model()` did not check the column it reads.** A weights
   table without `model` reported `NA` as the name, which then reached
   `pull_out()` and came back as the unchanged object with a "non-existent in
   current set of models" message. The guard now refuses a table missing either
   column.

5. **The stated reason for reading the `model` column rather than the row names
   was wrong, in two successive versions.** Row names do survive row-subsetting
   of the assembled `mod_stats`; the frame whose row names are positions is the
   intermediate inside `expand_manec()`, which no caller holds. The comment now
   states the mechanism: the model names reach the row names only at
   `cbind(mod_stats, disp)`, from the dispersion matrix, so a change to how
   those columns are attached would drop them silently, while the column is set
   by name.

One mechanism found on the way, recorded because the idiom is used elsewhere in
this suite: under testthat's third edition `expect_message()` returns the
condition it matched rather than the value of the expression, so a chain that
binds a fit from it binds a `simpleMessage`. The expression does still run to
completion, and an expectation nested inside such a chain still bites. The
hurdle test uses `capture_messages()` instead.

Answered without a change: a `quiet` argument, since `suppressMessages()`
covers it and neither `pull_out()` nor this accessor has anything else to
silence; sharing the selection with `fit_best_waic()` in `R/bnec_group.R`, which
reads a `mod_stats` that nothing subsets and is fragile rather than wrong;
`@seealso` on `amend()`, which has no such section to add to; and the
missing-argument message, where the `default` method's error names the three
classes and where a `bayesnecgroupfit` should go instead.

https://claude.ai/code/session_0192FF4nsjtuGufrtat1bygq

Issue #316, opened 2026-09-11
The test suite runs at its parallel floor: one file sets 16 minutes of every CI job


beckyfisher:
## What

`tests/testthat/test-check_fit.R` takes 16 minutes on its own. `testthat`'s
parallel runner distributes work by file, so elapsed time for the whole suite
cannot fall below the longest single file. That 16 minutes is therefore a floor
on every `R CMD check` job, whatever the runner's core count. Splitting the file
removes the floor.

## Why it matters

CI is the bottleneck on `dev`. A full check matrix is 20 to 38 minutes per job
and the test suite is essentially all of it: dependencies are cached and install
in under a minute, vignettes rebuild in 11 to 13 seconds because they are
precompiled, and examples are 20 seconds plus 45 to 72 seconds with
`--run-donttest`.

The suite already sets `Config/testthat/parallel: true`, and the measured
elapsed test time on `ubuntu-latest (release)` is 16 minutes against 29 minutes
of CPU. Those two figures agreeing with the longest file is the finding: **the
suite is already running at its parallel floor, so adding workers or a larger
runner buys nothing.** Splitting the longest file is what makes any further
parallelism useful.

## Evidence

R 4.6.1, brms 2.23.0, `issue-314-regularizing-predictor` at `7c8b0562`,
2026-09-11. Each test file was run in its own process, which is how the parallel
runner distributes them, with `NOT_CRAN=true`. `load_all()` is excluded and
averaged 7.0 s per process. Nothing was measured on the runners themselves.

58 files, 3,968 assertions, **73.2 minutes of serial test time**.

| file | seconds | assertions | s/assertion |
|---|---|---|---|
| `test-check_fit.R` | 958 | 71 | 13.5 |
| `test-inits_functions.R` | 688 | 484 | 1.4 |
| `test-dispersion.R` | 386 | 36 | 10.7 |
| `test-disp_model.R` | 352 | 122 | 2.9 |
| `test-get_priors.R` | 303 | 68 | 4.5 |
| `test-component-args.R` | 258 | 34 | 7.6 |
| `test-bnec_hurdle.R` | 235 | 56 | 4.2 |
| `test-amend.R` | 197 | 21 | 9.4 |

Eight of 58 files are 77% of the total; the top two are 37.5%.

Implied bounds on the test phase, from these measurements:

| | elapsed |
|---|---|
| floor set by the longest file | 16.0 min |
| total work over 2 workers | 36.6 min |
[...]

PR #331, opened 2026-09-13
Build each repeated test fixture once: fourteen Stan compilations to five (#328)

### LINKED ISSUE #328: Ten of the suite's twenty-nine Stan compilations rebuild a program already built
## What

The test suite is almost the whole of every `R CMD check` job. Two properties of
it are costing time without buying any coverage: it runs over two worker
processes whatever the runner provides, and it compiles ten Stan programs that
have already been compiled in the same process. Neither is fixed by removing a
test.

## Why it matters

A check job is 25 to 31 minutes and the test phase is 21 to 26 minutes of it.
Six pull requests are open against `dev`, so a correction propagated down the
stack is read three quarters of an hour after it is pushed. Nothing else in the
check is large: dependencies are cached and install in under a minute, examples
are 12 to 21 seconds plus 44 to 83 seconds under `--run-donttest`, and the
precompiled vignettes rebuild in 8 to 20 seconds.

The failures seen on pull requests over the past week are not caused by the
length of the suite. That distinction sets what the remedy is allowed to do. Four `R-CMD-check` runs have failed since 2026-09-09. Each
failed on a deterministic assertion, on all four platforms at once, naming the
defect: `as_one_character(prior)` on `issue-302-predictor-prior` and
`issue-309-init-search`, and `test-nechormepwr-bounded.R:90` on
`issue-302-predictor-prior`. No job has timed out, and the twelve jobs of the
three runs started on 2026-09-13 all ran concurrently with no queueing. What the
length adds is latency, not incorrect results, so every assertion has to be
left standing.

## Evidence

R 4.6.1, brms 2.23.0, rstan 2.32.7, `dev` at `127acf15`. CI figures are from run
34696507417 (2026-09-12), the last full green matrix. Local figures are from
Debian on WSL2 with the package checkout on a Windows drive, which compiles more
slowly than the runners; they are used for ratios and counts, not as predictions
of runner time.

The test phase against the rest of the check:

The test phase against the rest of the check, on run 34696507417 of 2026-09-12:

| job | test phase, CPU | test phase, elapsed | whole job |
|---|---|---|---|
| `ubuntu-latest` (release) | 51 min | 26 min | 31 min |
| `ubuntu-latest` (devel) | 42 min | 24 min | 29 min |
| `macOS-latest` (release) | 33 min | 21 min | 25 min |
| `windows-latest` (release) | not reported | 24 min | 31 min |

One run, deliberately, and not a range over several. Quoting a range here would be
wrong, and the reason is worth recording because it will recur: **the suite is
growing fast enough that two runs days apart do not measure the same thing.** The
`PASS` counts of the four most recent green matrices are 3,835, 3,923, 4,000 and
4,050 — 215 assertions added in thirty-six hours — and `macOS-latest` CPU is
monotone in that count at 17.2, 22, 30 and 33 minutes. A "17 to 33 minute range"
read off those four is four measurements of four different suites, not the spread
of one.

Any timing comparison across runs on this repository has to quote the `PASS` count
beside the time, and a before-and-after pair has to come from runs whose counts
agree. 34696507417 is the last green matrix at 4,050.

#### The worker count

`testthat:::default_num_cpus()` reads
`getOption("Ncpus")`, then the `TESTTHAT_CPUS` environment variable, and returns
`2L` when neither is set. Neither is set here, and the CI log confirms it with
`Starting 2 test processes.` The elapsed-to-CPU ratios above are 1.96, 1.75 and
1.57, which is two workers scaling close to perfectly, so elapsed time is bound
by total work divided by the worker count. `--as-cran` does not constrain this:
`_R_CHECK_LIMIT_CORES_` is read by `parallelly` and `future`, not by base R or
`testthat`.

#### The compilation inventory

Ten of 29 Stan compilations rebuild a program already built in the same process.
Every `rstan::stan_model()` call in the suite was recorded with the MD5 of the
Stan code it was given:

| | |
|---|---|
| compilations | 29 |
| distinct Stan programs | 19 |
| mean seconds per compilation | 79.3 |
| programs compiled 5 times | 2 |
| programs compiled twice | 2 |

That is 10 redundant compilations, 793 seconds of local CPU, in a run that
stopped part way through the suite, so 10 is a lower bound. The two programs
compiled five times are the two blocks of `bnec_hurdle()` in
`tests/testthat/test-check_fit.R`, where `hurdle_fixture()` is called by five
separate `test_that()` blocks and refits from scratch in each. That file was
already the worst in the suite at 958 seconds for 71 assertions (#316).
`poisson_fit()` in `tests/testthat/test-dispersion.R` is called by two blocks
the same way, in a file that already uses the memoised form for
`degenerate_fit()` with the comment "Fitted once and reused".

Nothing is reused between calls on its own: two identical `bnec()` calls in one
session took 88.8 and 89.2 seconds, and setting `rstan_options(auto_write =
TRUE)` does not change that, because rstan's cached object is rejected by its own
`avoid_crash()` check once the previous model's shared object has been unloaded.
Reuse has to be arranged by holding the fitted object, which is what the
memoised form does.

#### Repeated work outside the sampler

`check_fit()` defaults to `seed = 10` and is therefore deterministic. `test-check_fit.R` calls `check_fit(manec_example)`
four times with identical arguments and `check_fit(n4, group = 4, ndraws = 50)`
three times, and rebuilds `pull_out(manec_example, model = "nec4param")` seven
times although `setup.R` already provides that object as `nec4param`.

<details>

### The order the work should be done in

Each step is measured against the `Running 'testthat.R' [CPU/elapsed]` line of
the check log, which is the only figure that matters, and each is reported
separately so the gains are attributed rather than assumed.

1. **Remove the repeated work, which lowers CPU.** Give `hurdle_fixture()` and
   `poisson_fit()` the memoised form `test-dispersion.R` already uses for
   `degenerate_fit()`:

   ```r
   hurdle_fixture <- local({
     cached <- NULL
     function() {
       if (is.null(cached)) {
         cached <<- ...
       }
       cached
     }
   })
   ```

   The returned object is identical to the one each call built before, so no
   assertion changes. Two conditions make that safe and both hold: the fixture
   is seeded (`set.seed(17)`), and no test modifies the fit it is given --- every
   one of the five reads it. Do the same for the repeated `check_fit()` calls,
   and use `setup.R`'s `nec4param` in place of the seven `pull_out()` calls.

   Leave the one `check_fit()` call whose warning is asserted
   (`expect_warning(check_fit(n4, ndraws = 50), "not a design point")`) calling
   directly, because a memoised accessor emits its warning only on the first
   call.

2. **Raise the worker count, which lowers elapsed time.** Set `TESTTHAT_CPUS`
   from `parallel::detectCores()` in `R-CMD-check.yaml`, clamped at four. Set it
   in the workflow and not in `DESCRIPTION`: CRAN runs its checks within two
   cores and testthat's default already gives that, so raising the count is a
   property of these runners rather than of the package. Clamped because one
   Stan program compiling peaks at about 1.2 GB resident and the smallest runner
   in this matrix has 7 GB. Confirm it took effect by reading
   `Starting N test processes.` in the check log --- `rcmdcheck` runs the tests
   in a subprocess and an environment variable that did not reach it would
   otherwise fail silently.

3. **Then re-measure before doing anything further.** Step 1 lowers the CPU
   figure and step 2 divides it; whether any file is then long enough to set a
   floor of its own is a question to answer from the new numbers, not from these.

### Changes not to make

- Do not reduce the check matrix. `windows-latest` is usually the longest job, at
  31 minutes on the run above and 37 to 39 on the three before it, and #275 was a
  Windows-only failure that the `StanHeaders` pin exists to prevent. Dropping or
  deferring that job trades coverage for latency and is a separate decision.
- Do not skip vignette building. It takes 8 to 20 seconds because the vignettes
  are precompiled, and `R-CMD-check.yaml` records why they are checked.
- Do not lower `iter` across the suite as the first step. 35 of the 56
  `bnec()`-family calls in the tests inherit `iter = 1e4`, `warmup = 8000` and
  `chains = 4` from `add_brm_defaults()`, which looks like the obvious target,
  but most of those calls are `expect_error()` cases that never reach the
  sampler. Which of them reach the sampler has not been measured, and lowering
  `iter` on a fit whose test reads a parameter changes what the test asserts.
- Do not add `skip_on_ci()` anywhere. Three tests already have it ---
  `test-component-args.R:67` and `test-disp_model.R:397,423` --- so three
  regression tests for real defects do not run on the runners at all. That is
  the failure mode this issue is trying not to widen.

### An unrelated deadlock

A `testthat::test_local()` run of the whole suite deadlocked with both workers
idle and the parent waiting, after `test-timeout.R`'s output was the last
relayed. `test-timeout.R:41` interrupts a live `brm()` call with
`timeout = 1e-3`, which is a plausible mechanism for leaving a worker
unresponsive, but it was not isolated and it has never been seen under
`R CMD check`, which has completed every run. Recorded so that it is not
rediscovered as a symptom of the changes above.

### Definition of done

The `Running 'testthat.R'` CPU and elapsed figures are reported for all four
jobs against the table above; the suite passes with the same number of
assertions as before the change, read from `[ FAIL 0 | WARN n | SKIP 3 | PASS
4050 ]`; and no test is skipped, deleted or given a shorter chain that was not
shortened before.

</details>

---

## Addendum

Written 2026-09-13, after #331 measured the worker count.

The compilation finding stands and #331 fixes it: fourteen compilations to five
across the two files, with every assertion unchanged.

**The worker-count half is closed rather than deferred, and the measurement that
appeared to support it is withdrawn.** It was raised to one per core, then to one
fewer, then removed. The reason for removing it is that the quantity it targets
cannot be measured on this infrastructure at the precision required.

Two paired comparisons of two workers against three, on the same tree in the same
window, disagree in direction:

| platform | 2 workers | 3 workers | ratio |
|---|---|---|---|
| `ubuntu-latest` (release) | 37 CPU / 20 elapsed | 55 / 21 | elapsed 1.05, CPU 1.49 |
| `windows-latest` (release) | 25 elapsed | 18 elapsed | elapsed 1.39, the other way |

Both sit beside a null cell measured with **no configuration difference at all** —
two `macOS-latest` cells of the same runs, identical worker count on the identical
tree minutes apart, at 1.27 on elapsed and 1.33 on CPU. Neither the unfavourable
ubuntu figure nor the favourable Windows one is distinguishable from that. An
effect of this size against this variance needs roughly sixteen pairs per arm,
about thirty-two matrices, which is why the question is closed here rather than
left open for someone to rediscover at that price.

### The withdrawn CPU figure

This issue, and commit `6cfba6a2`, stated that raising the worker count to one per
core took the `ubuntu-latest (release)` test phase from a 41-to-51 minute CPU
range to 62. The 62 is real; the range is not a
reference interval. At two workers with `PASS 4050` in every case, that phase was
**31, 41 and 51 minutes of CPU and 16, 22 and 26 elapsed** on three runs of `dev`.
The 62 is one sample from a distribution spanning 31 to 51. Commit messages cannot
be edited, so `6cfba6a2` still states it as fact; this addendum is the record.

### The withdrawn stall

An earlier version of this issue suggested that `ubuntu-latest (devel)` stalls on
the #331 branch. It was terminated there four
times, but of the last 36 completed `pull_request` runs on this repository the
devel job was cancelled 23 times and succeeded 11. The branch is unremarkable
against a 68 per cent base rate. Filed separately as #333.

### Rules for any future timing work here

1. **Quote the `PASS` count beside any time**, and pair only runs whose counts
   agree. The suite gained 215 assertions in thirty-six hours, so two runs days
   apart do not measure the same thing.
2. **A single pair resolves nothing below about 30 per cent.** The null cells above
   are the yardstick, and they were measured between identical configurations.
3. **Check the base rate of any event before reasoning from it.** A 2×2 design, a
   pre-registered decision rule and several rounds of review were spent here on an
   event that happens to two thirds of runs.
4. **Vary one setting through `workflow_dispatch`, not by editing the workflow.**
   #331 keeps an input for this. While the worker count was computed in the
   workflow it could not be varied for measurement, because varying it meant
   editing the file and editing the file changed the tree being measured.

beckyfisher:
Closes #328. Raises #333.

## What

`tests/testthat/test-check_fit.R` and `tests/testthat/test-dispersion.R` build
their repeated fixtures once instead of once per `test_that()` block. The suite
compiles **five Stan programs where it compiled fourteen**, and runs every
assertion it ran before, against the same objects.

The workflow gains a `workflow_dispatch` input that can hold the testthat worker
count, a `timeout-minutes` bound and a dump of the test log that runs whatever
happened. It sets no worker count on an ordinary run, so CI is configured exactly
as `dev` is.

## Why it matters

`hurdle_fixture()` in `test-check_fit.R` was called by five separate
`test_that()` blocks, and `bnec_hurdle()` fits two brms models, so the same two
Stan programs were compiled five times each. `poisson_fit()` in
`test-dispersion.R` was called by two blocks the same way. Stan compilation is
what the suite spends its time on: 29 compilations over 19 distinct programs when
#328 instrumented it, at 79 seconds each locally.

`test-dispersion.R` has no `skip_on_cran()`, so one of the removed
compilations is removed from CRAN's own check as well as from ours. That is the
one saving here that reaches anyone outside this repository.

#### The compilation count

This is the claim. Every `rstan::stan_model()` call was counted by replacing it in
its namespace, on `dev` and on this branch:

| file | compilations | assertions |
|---|---|---|
| `test-check_fit.R` | **10 to 2** | 71, unchanged |
| `test-dispersion.R` | **4 to 3** | 36, unchanged |

A count, deterministic, reproduced. Running each file in its own process, the same
pair takes 1,177 seconds to 415 — but that figure is a single-process local
measurement on one machine, and is reported as such rather than as a prediction of
runner time.

**On CI the timing effect is below the noise floor, which was predicted before it
was measured.** Nine fewer compilations at 30 to 40 runner-seconds is about five
minutes against a 52-minute test phase, under ten per cent. The noise is larger
than that, so there was nothing to detect, and there was not:

| platform | `dev` at 3 workers | this branch at 3 workers | ratio |
|---|---|---|---|
| `ubuntu-latest` (release) | 52 min CPU / 19 elapsed | 55 / 21 | 1.06 **against** |
| `windows-latest` (release) | 23 min elapsed | 18 | 1.28 for |

Same worker count, same window, trees differing only by this change and a
documentation delta. The two platforms disagree in direction, which is what a
difference below the noise floor looks like. Neither figure supports the change
and neither undermines it; the compilation count does not depend on either.

#### The noise floor, measured rather than assumed

Two cells of the same run ran the identical configuration on the identical tree
minutes apart, `macOS-latest` at two workers in both arms:

| | CPU | elapsed |
|---|---|---|
| one cell | 32 min | 19 min |
| the other | 24 min | 15 min |
| ratio | **1.33** | **1.27** |

A third on the same platform at fixed configuration, days apart, spans 17 to 33
minutes of CPU. On `ubuntu-latest (release)`, at two workers with `PASS 4050` in
every case, the test phase was 31, 41 and 51 minutes of CPU and 16, 22 and 26
elapsed. Any cross-run timing comparison here has to quote the `PASS` count beside
the time and pair only runs whose counts agree — the suite gained 215 assertions
in thirty-six hours — and even then a single pair resolves nothing below about 30
per cent.

## Correctness of the reuse

Each accessor returns the object its callers used to build for themselves.

1. **The fixtures are deterministic.** `hurdle_fixture()` and `poisson_fit()` seed
   their simulated data and pass `seed =` to the sampler. `check_fit()` and
   `dispersion()` take `seed = 10` by default and call `set.seed(seed)` internally
   before the only stochastic step they run (`R/check_fit.R:168`,
   `R/dispersion.R:127`), so neither depends on the RNG state it is entered with.
   Removing the repeat calls changes no drawn value, even though it no longer
   advances the stream.
2. **No test modifies what it is handed.** All five blocks reading
   `hurdle_fixture()` read from it, as do both reading `poisson_fit()` and all
   four reading `manec_checkfit()`. A comment on each accessor says so, and says
   that a test needing to modify a fixture must build its own.
3. **`setup.R` already provided the pulled-out fit.** Seven rebuilds of
   `pull_out(manec_example, model = "nec4param")` are replaced by `setup.R`'s
   `nec4param`, which is that call.

The one place reuse would have changed an assertion is the block asserting
`check_fit()`'s warning: a memoised accessor emits its warning on the first call
only, so `expect_warning(check_fit(nec4param, ndraws = 50), "not a design point")`
keeps calling `check_fit()` directly.

<details>

### The memoised form

The form `degenerate_fit()` in `test-dispersion.R` already used, with the comment
"Fitted once and reused":

```r
poisson_fit <- local({
  cached <- NULL
  function() {
    if (is.null(cached)) {
      cached <<- ...
    }
    cached
  }
})
```

Applied to `hurdle_fixture()`, to `poisson_fit()`, and to two accessors in
`test-check_fit.R`: `manec_checkfit()` for the four identical
`suppressWarnings(check_fit(manec_example))` calls and `nec4param_checkfit_g4()`
for the three identical `check_fit(nec4param, group = 4, ndraws = 50)` calls. Each
sits immediately above the block that first calls it.

The invariant the comments record is that every block reading an accessor has the
same skip guard, so if one can skip they all can and none is left needing a
table that was never built. Calling the accessor first after `skip_on_cran()` is
the convention that makes this visible; it is not the invariant.

### The absent worker count

An earlier version of this pull request raised the testthat worker count from its
default of two. `testthat:::default_num_cpus()` reads `getOption("Ncpus")`, then
`TESTTHAT_CPUS`, and returns `2L` when neither is set, and neither was set.

It was measured and dropped. Two paired comparisons, same tree and same window,
disagree in direction: `ubuntu-latest (release)` elapsed 20 to 21 minutes, a ratio
of 1.05, and `windows-latest (release)` 25 to 18, a ratio of 1.39 the other way,
with ubuntu CPU 37 to 55 at 1.49. Set against the 1.27 and 1.33 null cells above,
measured with no configuration difference at all, none of those is distinguishable
from noise — the favourable Windows figure no more than the unfavourable ubuntu
one. Detecting an effect of that size against this variance needs roughly sixteen
pairs per arm.

#328 holds that measurement and closes the question rather than deferring it, so
that nobody spends thirty matrices rediscovering it. It also withdraws the CPU
figure that commit `6cfba6a2` gives as its rationale; commit messages cannot be
edited, and that one states 62 minutes against a 41-to-51 range as fact, which
later turned out to be one sample from a distribution spanning 31 to 51.

### The workflow additions

- **The `workflow_dispatch` input.** It does nothing unless supplied, so every
  ordinary run is configured as `dev` is. It exists because while the count was
  computed in the workflow it could not be varied for measurement: varying it
  meant editing the file, and editing the file changed the tree being measured.
  Holding one setting is what made the paired comparisons above possible.
- **`timeout-minutes: 90`.** The default is six hours. Terminations on this
  repository arrive as a bare exit 143 partway through the test phase with no
  timing line, and neither of the two seen here was a timeout, so neither could be
  told from a job that was merely slow. A bound well above the longest observed
  job makes a hang a deterministic event at a known boundary.
- **An `if: always()` dump of `testthat.Rout`.** The action's own "Show testthat
  output" step reported `outcome=skipped` on both terminations, so neither left
  any record of which file the suite had reached. This is also the gap that left
  #328's own deadlock note unable to name its last file.

### Considered and left out

- **`rstan_options(auto_write = TRUE)`.** Tested, because a one-line fix would
  have been preferable to editing tests. `rstan::stan_model()` on identical code
  takes 78.9 seconds then 0.1 while the earlier model object is still reachable,
  so the mechanism works in isolation. It shortens nothing here: three identical
  `bnec()` calls in one session took 91.2, 90.6 and 85.4 seconds with the option
  set, because rstan's `avoid_crash()` rejects the cached object once the previous
  model's shared object has been unloaded. Holding the fitted object is what
  achieves the reuse, which is what this change does.
- **Lowering `iter`.** 35 of the 56 `bnec()`-family calls in the tests inherit
  `iter = 1e4`, `warmup = 8000` and `chains = 4` from `add_brm_defaults()`. Most
  are `expect_error()` cases that never reach the sampler, which of them sample
  has not been measured, and lowering `iter` on a fit whose test reads a parameter
  changes what that test asserts.
- **Splitting `test-check_fit.R`**, which is #316's first step. The repeated fits
  were why that file was long; it is now shorter than `test-dispersion.R`.
- **`Config/testthat/start-first`.** Needs per-file figures from the current tree,
  and those are now known to be worth little on CI at a single observation.
- **Reducing the check matrix.** `windows-latest` is the longest job, 31 to 39
  minutes across four green runs, and #275 was a Windows-only failure that the
  `StanHeaders` pin exists to prevent.
- **Re-enabling the three `skip_on_ci()` tests**, at `test-component-args.R:67`
  and `test-disp_model.R:397,423`. They guard real fixed defects and do not run on
  the runners, which is worth changing — but it raises coverage rather than
  lowering runtime, and has its own flake risk, so it belongs in its own
  change.

### Risks checked for shared state

Not needed for the shipped configuration, which is testthat's default two workers
as `dev` runs today, but checked while more workers were under consideration and
recorded because it bears on #333 and on `start-first`.

`setup.R` is sourced once per worker by `testthat:::queue_process_setup()` and
each file runs in a child of that worker's environment, so every worker holds its
own `nec4param` and the accessors are per file. Snapshots are per worker through
`SubprocessSnapshotReporter`, and `_snaps` is empty. The one genuinely shared path
is the default graphics device: `test-bayesnec_methods.R:17`,
`test-bayesmanec_methods.R:23`, `test-expand_classes.R:298` and
`test-check_priors.R:6` each draw to it, opening `Rplots.pdf` in the shared check
directory. No platform here denies the second `fopen`, and `Rplots.pdf` is matched
by `.Rbuildignore`, so the symptom would be a corrupt file nothing reads;
`test-plot.R:32` shows the fix, `pdf(NULL)`, if it ever binds.

### Not changed

No `NEWS.md` entry. This touches two test files and the workflow, and the shipped
workflow sets no worker count, so nothing a user can reach behaves differently.
No test is skipped, deleted, or given fewer chains or iterations than before.

</details>

https://claude.ai/code/session_01H6p5XE9Uo4EbDanK5QXtmp

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -111,6 +111,39 @@ jobs:
           cmdstanr::check_cmdstan_toolchain(fix = TRUE)
         shell: Rscript {0}
 
+      # testthat runs its files over two worker processes unless it is told
+      # otherwise: testthat:::default_num_cpus() reads getOption("Ncpus"), then
+      # TESTTHAT_CPUS, and returns 2 if neither is set. The test suite is very
+      # nearly the whole check -- dependencies are cached and install in under a
+      # minute, examples are 12 to 21 seconds plus 44 to 83 seconds with
+      # --run-donttest, and the precompiled vignettes rebuild in 8 to 20 seconds
+      # -- so the worker count sets the length of the job.
+      #
+      # On run 34696507417 the test phase reported 51, 42 and 33 minutes of CPU
+      # against 26, 24 and 21 minutes elapsed on the three jobs that report
+      # both. Those ratios, 1.96, 1.75 and 1.57, are two workers scaling close
+      # to perfectly, which means elapsed time is bound by total work divided by
+      # the worker count and not by the longest single file.
+      #
+      # The count is read from the runner rather than written here as a
+      # constant, so that a change to the hosted runner sizes is followed
+      # instead of ignored, and it is clamped at four because the memory each
+      # worker needs is not small: one Stan program compiling peaks at about
+      # 1.2 GB resident, measured on R 4.6.1 with brms 2.23.0, and the smallest
+      # runner in this matrix has 7 GB.
+      #
+      # It is set here and not in DESCRIPTION on purpose. CRAN runs its checks
+      # within two cores and testthat's default already gives that; raising the
+      # count is a property of these runners, not of the package.
+      - name: Set the testthat worker count from the runner
+        run: |
+          cores <- parallel::detectCores(logical = TRUE)
+          workers <- max(1L, min(as.integer(cores), 4L))
+          cat("detected", cores, "cores; using", workers, "testthat workers\n")
+          cat(sprintf("TESTTHAT_CPUS=%d\n", workers),
+              file = Sys.getenv("GITHUB_ENV"), append = TRUE)
+        shell: Rscript {0}
+
       # Vignettes ARE built and checked. They are precompiled -- precompile.R
       # renders every .Rmd.orig to a .Rmd whose chunks are display-only -- so
       # building them is markdown to HTML and costs ~16s, not a refit. Skipping
--- a/tests/testthat/test-check_fit.R
+++ b/tests/testthat/test-check_fit.R
@@ -3,10 +3,42 @@
 # variability in one region, because a free dispersion parameter absorbs exactly
 # the discrepancy the global statistic measures.
 
+# The fixtures below are built once and reused, following degenerate_fit() in
+# test-dispersion.R. Each was previously rebuilt in every test_that() block that
+# needed it, and each is deterministic, so the reused object is the one those
+# blocks built for themselves: setup.R's `nec4param` is pull_out() of the
+# packaged model set, and check_fit() takes `seed = 10` by default and draws
+# from a stored posterior. Safe here only because no test modifies the object it
+# is handed --- every one of them reads it. A test that needs to modify a
+# fixture must build its own.
+#
+# check_fit() on the packaged example warns that the predictor is not a design
+# point. The accessor suppresses it, as every call site here already did; the
+# block that asserts that warning calls check_fit() directly and must keep
+# doing so, because a memoised accessor emits it on the first call only.
+manec_check <- local({
+  cached <- NULL
+  function() {
+    if (is.null(cached)) {
+      cached <<- suppressWarnings(check_fit(manec_example))
+    }
+    cached
+  }
+})
+
+check_fit_g4 <- local({
+  cached <- NULL
+  function() {
+    if (is.null(cached)) {
+      cached <<- check_fit(nec4param, group = 4, ndraws = 50)
+    }
+    cached
+  }
+})
+
 test_that("check_fit returns a row per group with both statistics", {
   skip_on_cran()
-  n4 <- suppressMessages(pull_out(manec_example, model = "nec4param"))
-  out <- check_fit(n4, group = 4, ndraws = 50)
+  out <- check_fit_g4()
   expect_s3_class(out, "checkfit")
   expect_equal(nrow(out), 4)
   for (nm in c("group", "n", "obs_mean", "sim_mean", "mean_ratio", "ppp_mean",
@@ -14,7 +46,7 @@ test_that("check_fit returns a row per group with both statistics", {
     expect_true(nm %in% names(out), info = nm)
   }
   # every observation lands in exactly one group
-  expect_equal(sum(out$n), nrow(n4$fit$data))
+  expect_equal(sum(out$n), nrow(nec4param$fit$data))
 })
 
 test_that("exactly one group is flagged as the control, and it is the lowest", {
@@ -22,16 +54,14 @@ test_that("exactly one group is flagged as the control, and it is the lowest", {
   # observed control is y[x == min(x)] -- the package's own convention. The
   # flag has to agree with that or it points at the wrong row.
   skip_on_cran()
-  n4 <- suppressMessages(pull_out(manec_example, model = "nec4param"))
-  out <- check_fit(n4, group = 4, ndraws = 50)
+  out <- check_fit_g4()
   expect_equal(sum(out$control), 1)
   expect_true(out$control[1])
 })
 
 test_that("posterior predictive p-values are probabilities", {
   skip_on_cran()
-  n4 <- suppressMessages(pull_out(manec_example, model = "nec4param"))
-  out <- check_fit(n4, group = 4, ndraws = 50)
+  out <- check_fit_g4()
   for (nm in c("ppp_mean", "ppp_sd")) {
     expect_true(all(out[[nm]] >= 0 & out[[nm]] <= 1, na.rm = TRUE), info = nm)
   }
@@ -42,8 +72,7 @@ test_that("ndraws is reduced to what the fit holds rather than erroring", {
   # the default of 1000 would fail on the package's own example object -- which
   # is exactly what someone runs a diagnostic on first.
   skip_on_cran()
-  n4 <- suppressMessages(pull_out(manec_example, model = "nec4param"))
-  expect_no_error(check_fit(n4, group = 3, ndraws = 1e6))
+  expect_no_error(check_fit(nec4param, group = 3, ndraws = 1e6))
 })
 
 test_that("an unreplicated predictor is binned, with a warning", {
@@ -52,8 +81,7 @@ test_that("an unreplicated predictor is binned, with a warning", {
   # always returns something, including where the answer is meaningless, so it
   # has to say so.
   skip_on_cran()
-  n4 <- suppressMessages(pull_out(manec_example, model = "nec4param"))
-  expect_warning(check_fit(n4, ndraws = 50), "not a design point")
+  expect_warning(check_fit(nec4param, ndraws = 50), "not a design point")
 })
 
 test_that("a replicated predictor is grouped by its distinct values", {
@@ -104,8 +132,7 @@ test_that("check_fit reproduces the local finding a global statistic misses", {
   # than the data show in the control region. If check_fit cannot see that, it
   # does not do the job it was written for.
   skip_on_cran()
-  n4 <- suppressMessages(pull_out(manec_example, model = "nec4param"))
-  out <- suppressWarnings(check_fit(n4, ndraws = 200, seed = 10))
+  out <- suppressWarnings(check_fit(nec4param, ndraws = 200, seed = 10))
   ctrl <- out[out$control, ]
   expect_lt(ctrl$sd_ratio, 0.95)
   # and the steep tail fails the other way, which a single global number
@@ -119,7 +146,7 @@ test_that("check_fit reproduces the local finding a global statistic misses", {
 
 test_that("plot.checkfit returns a ggplot with both statistics panelled", {
   skip_on_cran()
-  cf <- suppressWarnings(check_fit(manec_example))
+  cf <- manec_check()
   p <- plot(cf)
   expect_s3_class(p, "ggplot")
   # both panels present -- location and scale fail independently, so a single
@@ -132,7 +159,7 @@ test_that("plot.checkfit returns a ggplot with both statistics panelled", {
 
 test_that("the control is distinguished in the plot data", {
   skip_on_cran()
-  cf <- suppressWarnings(check_fit(manec_example))
+  cf <- manec_check()
   p <- plot(cf)
   expect_true("control" %in% p$data$role)
   expect_true("exposed" %in% p$data$role)
@@ -145,7 +172,7 @@ test_that("the control is distinguished in the plot data", {
 
 test_that("the simulated intervals are on the object but not printed", {
   skip_on_cran()
-  cf <- suppressWarnings(check_fit(manec_example))
+  cf <- manec_check()
   expect_true(all(c("sim_mean_lo", "sim_mean_hi", "sim_sd_lo", "sim_sd_hi")
                   %in% names(as.data.frame(cf))))
   # print() drops them: they are for plot(), and including them takes the
@@ -156,7 +183,7 @@ test_that("the simulated intervals are on the object but not printed", {
 
 test_that("the interval brackets the simulated median", {
   skip_on_cran()
-  d <- as.data.frame(suppressWarnings(check_fit(manec_example)))
+  d <- as.data.frame(manec_check())
   expect_true(all(d$sim_mean_lo <= d$sim_mean & d$sim_mean <= d$sim_mean_hi))
   expect_true(all(d$sim_sd_lo <= d$sim_sd & d$sim_sd <= d$sim_sd_hi))
 })
@@ -165,19 +192,29 @@ test_that("the interval brackets the simulated median", {
 # each half against its own subset; neither asks whether the fit reproduces the
 # observed response, which contains the zeros and is what was measured.
 
-hurdle_fixture <- function() {
-  set.seed(17)
-  x <- rep(seq(0, 5, length.out = 15), each = 6)
-  p_alive <- 1 / (1 + exp(-(2.5 - 0.9 * x)))
-  alive <- rbinom(length(x), 1, p_alive)
-  g <- rgamma(length(x), shape = 4, rate = 4 / pmax(3 - 0.35 * x, 0.3))
-  dat <- data.frame(x = x, y = ifelse(alive == 1, g, 0))
-  fit <- suppressWarnings(suppressMessages(
-    bnec_hurdle(y ~ crf(x, "nec3param"), data = dat, iter = 400, warmup = 200,
-                chains = 2, seed = 17, refresh = 0, open_progress = FALSE)
-  ))
-  list(dat = dat, fit = fit)
-}
+# Fitted once and reused across the five blocks below. bnec_hurdle() fits two
+# brms models, so a call per block compiled the same two Stan programs five
+# times each: 10 of the 29 compilations the suite ran. See #328.
+hurdle_fixture <- local({
+  cached <- NULL
+  function() {
+    if (is.null(cached)) {
+      set.seed(17)
+      x <- rep(seq(0, 5, length.out = 15), each = 6)
+      p_alive <- 1 / (1 + exp(-(2.5 - 0.9 * x)))
+      alive <- rbinom(length(x), 1, p_alive)
+      g <- rgamma(length(x), shape = 4, rate = 4 / pmax(3 - 0.35 * x, 0.3))
+      dat <- data.frame(x = x, y = ifelse(alive == 1, g, 0))
+      fit <- suppressWarnings(suppressMessages(
+        bnec_hurdle(y ~ crf(x, "nec3param"), data = dat, iter = 400,
+                    warmup = 200, chains = 2, seed = 17, refresh = 0,
+                    open_progress = FALSE)
+      ))
+      cached <<- list(dat = dat, fit = fit)
+    }
+    cached
+  }
+})
 
 test_that("the observed response is reconstructed exactly, zeros included", {
   # The load-bearing assumption: bnec_hurdle() subsets rather than reorders, so
--- a/tests/testthat/test-dispersion.R
+++ b/tests/testthat/test-dispersion.R
@@ -9,17 +9,26 @@ test_that("dispersion fails because family is gaussian", {
 # bnec() forces link = "identity", the linear predictor is already on the
 # response scale, so the value was transformed a second time.
 
-poisson_fit <- function() {
-  set.seed(247)
-  x <- runif(60, 0, 3.2)
-  mu <- 5 + (85 - 5) * exp(-exp(0.3) * (x - 1.5) * (x > 1.5))
-  d <- data.frame(x = x, y = as.integer(rpois(length(mu), mu)))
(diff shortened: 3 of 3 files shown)


beckyfisher:
The title and body have been rewritten, and the scope is now smaller than what any
earlier review was written against. A record of the change follows, since the body
describes the current state rather than its history.

The testthat worker count is dropped. This pull request previously raised it from
its default of two, first to one per core and then to one fewer. It now sets no
worker count at all, so CI is configured exactly as `dev` is. It was measured, and
the effect cannot be told from a null cell measured between identical
configurations; #328 holds the figures and closes the question rather than
deferring it.

The workflow step survives only as a `workflow_dispatch` input that does nothing
unless supplied. It is the instrument that made the measurement possible, not a
policy.

Two additions: `timeout-minutes: 90`, and an `if: always()` dump of
`testthat.Rout`, because terminations here arrive as a bare exit 143 with no record
of where the suite had reached.

The memoisation is unchanged, and is what this pull request now consists of:
fourteen Stan compilations to five, every assertion unchanged, verified by
instrumenting `rstan::stan_model`.

Two numbers that earlier versions of this description relied on are withdrawn
rather than revised, and both are in #328's addendum. The first is the claim that
raising the worker count put 62 minutes of CPU against a 41-to-51 minute baseline,
where that baseline was three samples of a distribution spanning 31 to 51. The
second is the suggestion that `ubuntu-latest (devel)` stalls on this branch, where
23 of the last 36 pull-request runs on this repository have a cancelled devel job
whatever branch they ran on. That one is now #333.

beckyfisher:
## The review findings that survived checking

An independent review ran over this branch across eleven rounds. What follows is
the set of findings that survived checking, each with the commit that fixes it.
Findings that did not survive are not listed, and the exchange itself is not
reproduced.

### A defect in the code

`parallel::detectCores()` returns `NA` where it cannot determine the count, and
the guard did not exclude it. `max(1L, min(NA_integer_, 4L))` is `NA`,
`sprintf("%d", NA_integer_)` writes `TESTTHAT_CPUS=NA`, and testthat then aborts
the **entire test phase** with "`TESTTHAT_CPUS` must be an integer" — naming the
variable rather than the step that set it. Fixed in `6cfba6a2`, floored at two so
the step could never set fewer workers than the default it replaced. The worker
count has since been dropped altogether, and the same guard now protects the
dispatch input.

### Assertions the evidence did not support

A figure was wrong in three places: `windows-latest` was described as "the longest
job at 42 minutes", where it was 31 minutes on the run cited and 37 to 39 on the
three before it. Corrected here and in #328, and the conclusion it supported — that
the matrix should not be reduced — stands.

Two mechanisms were asserted without evidence, both corrected in `664e7147`. The
workflow comment claimed each testthat worker spawns its own compiler, so four
workers on four cores give eight runnable processes; the worker blocks in
`system()` while `R CMD SHLIB` runs, so it does not. And the accessor comment gave
the placement of a call as the reason the memoisation is safe, where the invariant
is that every block reading an accessor has the same skip guard.

A rewritten section left the fold contradicting it. The evidence section was
corrected while the `<details>` block still held the superseded code, the
retracted claim and the old accessor names, in six places. Since email
notifications send a body expanded, a fold cannot hold a retracted version of what
is above it.

### The finding that changed the conclusion

**CI wall-clock time on these runners cannot resolve an effect of the size this
change was claimed to have.** `ubuntu-latest (release)` on `dev`, two workers,
`PASS 4050` in every case:

| run | CPU | elapsed |
|---|---|---|
| 34739135545 | 31 min | 16 min |
| 34741511558 | 41 min | 22 min |
| 34696507417 | 51 min | 26 min |

Two cells of a single run, same tree, same configuration, minutes apart, differ by
1.33 on CPU and 1.27 on elapsed. Every timing comparison in the first version of
this pull request was one sample from that distribution, including the one that
motivated lowering the worker count. The ranges quoted as baselines were partly
this variance and partly four different suites: the four runs used had `PASS`
counts of 3,835, 3,923, 4,000 and 4,050.

That is why this pull request is now the memoisation alone, and why its claim is a
compilation count rather than a time. Fourteen compilations to five is a count.
Its effect on CI timing is below the noise floor, which was predicted from the
arithmetic before it was measured — nine compilations at 30 to 40 runner-seconds
is about five minutes against a 52-minute phase — and the measurement came back at
1.06 in the unfavourable direction on one platform and 1.28 in the favourable
direction on another, which is what a difference below the floor looks like. Both
are in the description.

### Reasoning the review stopped from shipping

An apparent R-devel stall. `ubuntu-latest (devel)` was terminated four times on
this branch and passes on `dev`, which read as a branch-specific hang and was
treated as a merge blocker. Of the last 36 completed `pull_request` runs on this
repository, the devel job was cancelled 23 times and succeeded 11. The base rate of
the event being reasoned about was never checked until the review insisted on it.
Filed as #333, which is a larger finding than anything in this pull request.

A decision rule that could not decide. A threshold was pre-registered for the
measurement, and the review showed twice that it could not separate the two halves
of the change — the memoisation's own expected CPU saving was the same width as the
band that had been called vindication. The second version was underpowered by about
an order of magnitude against variance that had not yet been characterised.

### Corrections to the review itself

Three of its findings needed correcting, which is recorded because a review where
only one side is checked finds less. Its proposed `TESTTHAT_CPUS=1` discriminator
would not have worked, since `testthat:::find_parallel()` consults
`Config/testthat/parallel` and `TESTTHAT_PARALLEL` and never the worker count. Its
reading of one cell as a paired measurement of the memoisation was wrong, because
the dispatch input overrides the computed count and that cell ran three workers,
not two. And its arithmetic making one termination "1.8× the worst normal case"
used a maximum taken from runs with different `PASS` counts; against devel's own
like-for-like maximum of 30 minutes the figure is 1.62, inside the noise, which is
what removed the last reason to think it was a stall at all.

PR #332, opened 2026-09-13
report the curve parameters of a fitted set

### LINKED ISSUE #297: No way to report curve parameter estimates from a model-averaged fit
## What

There is no way to obtain the curve parameters --- `top`, `bot`, `beta`, `nec`,
`ec50`, `slope`, `d` --- with credible intervals from a `bayesmanecfit`.
`summary()` reports the model weights, the per-model dispersion, the weighted
*N(S)EC* and the per-model Bayesian R-squared, but no parameter estimates. For a
single `bayesnecfit` the parameters are reachable through the underlying
`brmsfit`; for a model average nothing reports them at all.

## Why it matters

The curve parameters are what a methods section states alongside the threshold
estimates. `top` is the control response the whole curve is referenced to,
`bot` is the asymptote a `"relative"` ECx is measured against, and `beta`
is the decay rate. A reader who wants to know what curve was fitted, or to
reproduce it, cannot get those numbers out of the object that was fitted.

The gap is easy to miss because `summary()` looks comprehensive. It reports
seven blocks and none of them is the parameters, so the absence reads as a
deliberate omission rather than a missing feature.

## Evidence

`vignette("example9")`'s reporting section lists "the estimates with their
credible intervals" as something a methods section has to state. Working through
that vignette against `dev` at `0976caca` on 2026-09-08, the threshold estimates
are reachable --- `nec()`, `nsec()`, `ecx()` and the weighted *N(S)EC* row of
`summary()` --- and the parameters are not. `NAMESPACE` exports 40 functions and
none returns them.

The reporting checklist in that vignette is currently ambiguous on the point and
is being narrowed to "the threshold estimates", because the wider claim cannot
be met.

<details>

### The question to settle first

**What does a model-averaged `top` mean when the set mixes equations?** Not every
equation has every parameter: `ecxexp` has no `bot`, the three-parameter
equations have no `d`, and only `mod_groups$nec` has `nec`. A weighted average
over the equations that happen to have a parameter is an average over a
different subset for each parameter, weighted by stacking weights that were
computed for the whole set. That is not obviously a meaningful quantity, and it
is the reason this has not simply been done already.

Three options, in increasing order of what they claim:

1. **Per-equation only.** Report a table of parameter estimates by equation,
   with the stacking weight beside each, and no averaging. Honest, needs no new
   theory, and is what a reader reproducing one curve actually needs.
2. **Averaged within the equations that share a parameter**, with the weights
   renormalised over that subset, and the subset named in the output. Requires
   stating in the documentation that the weights are not the set's weights.
3. **Averaged over the whole set**, treating an absent parameter as structurally
   zero where that is defensible (`bot` for a three-parameter equation is zero
   by construction, so this one is defensible; `nec` for an ECx equation is not,
   since there is no threshold at all).

Option 1 is the recommendation. It is a reporting function rather than an
inferential one, it cannot be wrong, and it does not foreclose 2 or 3.

### Shape

`parameters(x)` or `estimates(x)`, returning a `data.frame` with columns
`model`, `wi`, `parameter`, `Estimate`, `Q2.5`, `Q97.5`, one row per parameter
per equation, ordered by weight. `summary = FALSE` returns the draws, as
`dispersion()` does. `show_params()` already names the parameters of each
equation and is the natural source for which rows to expect.

For a `bayesnecfit` the same function returns the one-equation case. For a
`bayesnechurdlefit` it returns one table per component, as `rhat()` and
`check_sampling()` already do for that class.

### What to check

- The parameters are on the response scale, since `bnec()` assigns the identity
  link; a fit where the caller supplied a link is on the link scale and the
  output has to say so.
- A predictor transformed inline by `crf()` puts `nec` and `ec50` on the
  transformed scale, the same trap `ecx()` has, so the function needs `xform`
  and it has to be applied to those two and not to `top`, `bot` or `beta`.
- `bnec_group()` fits per level, so the group case is one table per level.

### Definition of done

`parameters()` exported and documented, tests over a `bayesnecfit`, a
`bayesmanecfit` and a `bayesnechurdlefit`, a NEWS entry, and the sentence in
`vignette("example9")`'s reporting checklist widened back to name it.

Whether to go beyond option 1. Averaging parameters across equations is a
modelling decision, not an implementation detail.

</details>

beckyfisher:
Closes #297.

## What

`curve_params()` reports the parameters of the fitted concentration-response
equation --- `top`, `bot`, `beta`, `nec`, `ec50`, `slope`, `d` and `f` --- with
their credible intervals, one row per parameter per equation, with the model
weight beside each row. There are methods for `bayesnecfit`, `bayesmanecfit`,
`bayesnechurdlefit` and `bayesnecgroupfit`. `summary = FALSE` returns the draws
the table was computed from.

```
> curve_params(manec_example)
      model        wi dpar     link parameter   Estimate        Q2.5      Q97.5
1 nec4param 0.8274318   mu identity       top  2.1683084   2.0515539  2.2620465
2 nec4param 0.8274318   mu identity       bot -8.4184951 -13.5780477 -5.6830274
3 nec4param 0.8274318   mu identity      beta -0.6709559  -1.1783599 -0.1954755
4 nec4param 0.8274318   mu identity       nec  1.4610359   1.3608295  1.5277839
5 ecx4param 0.1725682   mu identity       top  2.3077962   2.1421523  2.5009383
6 ecx4param 0.1725682   mu identity       bot -5.6942272  -7.1930136 -4.0032952
7 ecx4param 0.1725682   mu identity      beta  0.9580802   0.7073639  1.1674993
8 ecx4param 0.1725682   mu identity      ec50  2.3535255   2.1455780  2.5540610
```

## Why it matters

`summary()` reports the model weights, the per-equation dispersion, the weighted
no-effect estimate and the per-equation Bayesian R-squared, and no parameter
estimates. For a single `bayesnecfit` the parameters were reachable through the
underlying `brmsfit`; for a `bayesmanecfit` nothing returned them at all. They
are what a methods section states alongside the threshold estimates: `top` is
the control level the whole curve is referenced to, `bot` is the asymptote a
`"relative"` ECx is measured against, and `beta` is the decay rate.

## The decision #297 asked to be settled

Option 1 of the three the issue sets out, which is the issue's own
recommendation: report per equation, with the weight beside each row, and
average nothing. The equations of a set do not share a parameter list ---
`ecxexp` has no `bot`, the three-parameter equations have no `d`, and only the
equations of `mod_groups$nec` estimate `nec`. Averaging a parameter over
whichever equations estimate it would average over a different subset for each
parameter, under weights computed for the whole set, so the rows contributing to
one number would account for a different share of the model average from the
rows contributing to the next. Options 2 and 3 are not foreclosed by anything
here: both would add an argument to this function rather than replace it.

## Evidence

The table is the fit's own summary rather than a second computation of it. On
`manec_example`, every `Estimate`, `Q2.5` and `Q97.5` equals
`brms::fixef(fit, robust = TRUE)` for the same parameter, and the `nec` row
equals `nec(fit)[1]`; both are asserted in `tests/testthat/test-parameters.R`.

## Not done here

The `vignette("example9")` reporting checklist named in the issue's definition
of done is on the unmerged `issue-219-workflow-vignette` branch and does not
exist on `dev`, so the sentence cannot be widened in this pull request. It
should be widened when that branch lands.

<details>

### Files

- `R/parameters.R` --- new. The generic, four methods, and the internal
  helpers.
- `tests/testthat/test-parameters.R` --- new, 22 tests. Nothing in it samples:
  the packaged `manec_example`, the fits `setup.R` already pulls out of it, and
  a two-block `brmsformula` built without fitting cover every branch, so the
  file adds no measurable time to the suite.
- `R/expand_classes.R` --- `expand_nec()` and `parameters()` held two copies of
  the parameter list. The set now has one definition and the order `expand_nec()`
  appends in has another, because the extracted elements are appended to the
  `bayesnecfit` in that order and `test-expand_classes.R` pins it; a test
  asserts the two hold the same parameters.
- `NEWS.md`, `DESCRIPTION` (2.1.3.34 to 2.1.3.35), `NAMESPACE` and
  `man/parameters.Rd` (generated).

### Shape

`summary = TRUE` returns a `data.frame` of `model`, `wi`, `dpar`, `link`,
`parameter`, `Estimate`, `Q2.5`, `Q97.5`, ordered by decreasing `wi` and then
by a fixed parameter order. `Estimate` is the posterior median and the interval is
equal-tailed, which is what `fixef(robust = TRUE)` reports.

`dpar` and `link` are added to the six columns the issue sketches. A joint
two-block fit from `bnec(family = "hurdle_gamma")` holds a second set of curve
parameters as `b_hu<par>_Intercept`, and reporting only the `mu` block would
return half the model with nothing said; `link` is per block as well, and in a
column rather than an attribute so that a saved table records the scale its
numbers are on. Both are present for every fit rather than added only where a
second block or a named link exists, so the shape does not change between fits.

The two blocks need not use the same equation --- `model_survival` and
`bnec_joint()` are for selecting them separately --- so `model` names each
block's own. That name is not stored on the fitted object: `model_survival` is
passed to `fit_bayesnec()` and discarded, and `recover_prebayesnecfit()`
rebuilds a fit from four named elements, so a fifth would be dropped by `c()`,
`+`, `amend()` and `update()`. It is recovered from the fitted formula, which
`add_hu_block()` builds deterministically from the equation name, and reported
as `NA` where no equation matches.

`wi` is 1 for a `bayesnecfit`, which is the whole of the set it describes.
`pull_out()` returns a fit in its own right and does not record the weight the
equation held in the set it came from, so there is no other number to report.

`summary = FALSE` returns a named list of draws-by-parameter matrices, one per
equation, rather than the plain vector `dispersion()` returns, because there is
more than one parameter per equation. A two-block fit names the second block's
columns `hu_top` and so on.

### The link

`bnec()` assigns `link = "identity"`, so `top` and `bot` are on
the response scale for every fit that did not name a link. Where one was named,
brms applies its inverse to the whole non-linear expression and they are on the
link scale. The link is read off the fit for every distributional parameter, reported in
the `link` column, and a message names any that is not the identity. Reported
once per call, not once per equation or per level: a set shares one family, and
the default set is 23 equations.

### The transformed predictor

`nec` and `ec50` are the only two parameters
measured on the predictor axis, so where `crf()` transforms the predictor
inline they are on the transformed scale, exactly as `nec()` and `ecx()` are.
`xform` applies to those two and to nothing else; `top` and `bot` are response
levels and `beta`, `slope`, `d` and `f` are shape parameters, and a predictor
transformation applied to any of them would return a number in the wrong units
in silence. It is applied per draw rather than to the summary, which matters
for an `xform` that is not monotone. Where the predictor is transformed inline
and `xform` was left at `identity`, a message names the transformation and the
two parameters it applies to.

### Reading the draws

`curve_param_draws()` matches `b_<prefix><par>_Intercept` in full rather than by
prefix, for the reason `extract_pars()` records: unanchored matching makes
`top` of a hurdle fit also match `hutop`. A group-level term on a parameter
leaves the population-level entry in place and adds `sd_` and `r_` entries,
which are not curve parameters and are not matched.

The draws are read rather than `fixef()`, so that `summary = TRUE` and
`summary = FALSE` report the same numbers and `xform` acts on the draws.

### The name

Neither of the two the issue proposed, and the reason is a measurement across
the installed library. `parameters` is exported by `parameters`, `dials`,
`distributional`, `flexmix`, `modeltools`, `spatstat.model`, `urltools` and
`elliptic`; the first of those reaches every bayesnec install as a hard
dependency of `modelbased`, so `library(parameters); library(bayesnec)` would
make `parameters(lm_fit)` an error naming the missing method. `params` is
exported by `ssdtools` and `S4Vectors`. The `ssdtools` one is worse than a
plain collision: it is an `@inheritParams` documentation stub, literally
`function(...) NULL`, so `library(bayesnec); library(ssdtools)` would leave
`params(fit)` returning `NULL` in silence --- and `ssdtools` is the package a
bayesnec user is most likely to have attached.

`curve_params` is exported by nothing. It sits beside `show_params()`, which
names the parameters of each equation without fitting anything, and it says
what it returns: the parameters of the equation, and not the family's
dispersion parameter, a group-level term's standard deviation or the `ogl`
offset, all of which the documentation states are excluded.

`estimates()`, the issue's other candidate, sits one letter from
`average_estimates()` and `compare_estimates()`, which return toxicity
estimates rather than parameters.

### Definition of done

Exported and documented, under the name settled above; tests over a `bayesnecfit`, a `bayesmanecfit`, a
`bayesnechurdlefit` and a `bayesnecgroupfit`, plus the two-block and
non-identity-link paths; a NEWS entry. The vignette sentence is the one item
outstanding, for the reason given above the fold.

</details>

https://claude.ai/code/session_013cUHDfTSzeFknunjETZRoG


beckyfisher:
The pull request body has been updated to describe the change as it now stands.
This comment records what the review found, since the body says what the code
does and not what was wrong with it on the way there.

Three review passes were made against the branch. Thirteen findings were
raised. Eleven were real and are fixed; one was withdrawn by the reviewer as
wrong; one is a decision rather than a defect and is set out at the end. Every
fix was verified by running it rather than by reading it.

## Defects that changed a reported number

The survival block was labelled with the response block's equation. A joint
two-block fit need not use the same equation on both sides --- `model_survival`
and `bnec_joint()` are for selecting them separately --- and every row was
stamped with the response block's name. On
`bnec(y ~ crf(x, "nec3param"), family = "hurdle_gamma", model_survival = "ecx4param")`
that reported an `ec50` under `nec3param`, which does not estimate one, with
nothing in the table to show it. The survival equation is not stored on the
fitted object: `model_survival` is passed to `fit_bayesnec()` and discarded,
and `recover_prebayesnecfit()` rebuilds a fit from four named elements, so a
fifth would be dropped by `c()`, `+`, `amend()` and `update()`. It is recovered
from the fitted formula instead --- `add_hu_block()` writes the block
deterministically from the equation name --- and reported as `NA` where no
equation matches, so a row is labelled only where the equation is identified.
All 23 equations round-trip, and no two regenerate the same expression.

The weight was matched on the row names of `mod_stats`. Those arrive
incidentally, from the dispersion matrix `expand_manec()` binds, and a table
that reached the function without them reported every weight as `NA`. It is
matched on the `model` column, and a set holding no row for a success model is
now an error naming it.

## Defects that misdescribed a reported number

`wi` was documented as the stacking weight. The default is pseudo-BMA ---
`define_loo_controls()` sets it, and #320 exists to stop an accidental
fall-through to stacking. Corrected in four places.

The `nec` row was documented as what `nec()` returns. That holds for a
single-block threshold fit and for nothing else: `nec()` returns the combined
threshold for a two-block fit and the model-averaged estimate for a set,
neither of which is a row in this table. The claim is now scoped and the other
two cases are stated.

A transformed predictor was documented but never detected. `crf(sqrt(x))` puts
`nec` and `ec50` on the transformed scale, and the function said so in the Rd
while messaging about the link, which `bnec()` makes rare. It now messages
where the predictor is transformed inline and `xform` was left at its default.

The link was recorded only in an attribute nothing prints. Measured, the
attribute survived row subsetting and `dplyr::filter()` and was dropped by
column subsetting and by `as_tibble()`, and it was absent from the hurdle and
group returns although the Rd said otherwise. It is now a `link` column.

`summary = FALSE` then had no record of the scale at all, since a matrix of
draws has no column to put one in. The draws list keeps the links in an
attribute, and the Rd says so.

## Defects in the tests

Three documented behaviours would have survived being broken. The full-name
column matching --- the reason `b_top_Intercept` of a hurdle fit does not also
match `b_hutop_Intercept` --- was mocked out of the only test that reached it;
`curve_param_draws()` is now split so that `select_curve_cols()` is exercised
on a constructed frame. Applying `xform` per draw was pinned only on the
median, which for a monotone transformation equals the transform of the median
exactly; it is now pinned on a quantile and on a non-monotone `xform`. The
canonical parameter order was asserted nowhere, because the test sorted the
names before comparing.

The scale reports were raised once per level and once per component, so a
twelve-level group printed the same paragraph twelve times. They are now gated
separately, because the two scales belong to different things: the link belongs
to the family, which a set and a group share and the two components of a
`bayesnechurdlefit` do not, and the inline transformation belongs to the
formula, which all three share. A hurdle pair on two different links still
reports both, each paragraph naming its component, and the test asserts it.

## The rest

`expand_nec()` and `parameters()` held two copies of the parameter list; the
two are now held in step by a test, the order `expand_nec()` appends in being
pinned by four assertions in `test-expand_classes.R`. Both wrapper methods
validate their arguments before printing anything. A comment recorded the wrong
reason for applying `xform` per draw, and another claimed a coupling between
the two parameter lists that is not there; both are corrected. One ruled verb
each in the Rd and in a comment.

## Withdrawn

A parameter fixed by a `constant()` prior was reported as dropped from the
table in silence. It is not: `make_stancode()` on a non-linear model with
`prior(constant(1), nlpar = "nec")` declares the parameter under
`transformed parameters`, so it reaches the draws as a zero-variance column and
is reported at its fixed value --- which is why `screenable_pars()` excludes it
from the Rhat screen rather than the draws (#244). The guard added while
settling this is kept, and its comment says it is a guard and that no `bnec()`
route is known to reach it.

## The export name

This is the decision left open. `parameters()` is what this issue's definition
of done names, and it is what the branch exports. It collides with
`parameters::parameters()`, which reaches every bayesnec install as a hard
dependency of `modelbased`. Neither package attaches the other, so the two meet
only in a session that attaches both, and there the package attached later
supplies the definition: `library(parameters); library(bayesnec)` then makes
`parameters(lm_fit)` an error naming the missing method. Registering a method
on the other generic does not remove that, because it dispatches
`model_parameters()`, and a method there would still leave `parameters()`
unfindable after `library(bayesnec)` alone.

`curve_parameters()` collides with nothing, and is the more accurate name as
well: the function reports the parameters of the equation, and the
documentation states that the family's dispersion parameter, group-level terms
and `ogl` are excluded. Renaming before merge is a one-line change.
@beckyfisher this one is yours to settle; the branch stands as `parameters()`
until you say otherwise.

beckyfisher:
The open decision above is settled: the function is `curve_params()`, and the
body has been updated to describe the branch under that name. Commit 1a715e1a
does the rename, so the section of the earlier comment that says the branch
stands as `parameters()` no longer describes it.

The deciding evidence is a third candidate and a measurement neither of the
two in the issue survives. `params` is exported by `ssdtools`, where it is an
`@inheritParams` documentation stub --- literally `function(...) NULL` --- so
`library(bayesnec); library(ssdtools)` would leave `params(fit)` returning
`NULL` in silence. That is worse than the `parameters` collision, which at
least raises an error, and `ssdtools` is the package a bayesnec user is most
likely to have attached beside this one.

Scanning every installed package for each candidate:

| name | exported by |
| --- | --- |
| `parameters` | `parameters`, `dials`, `distributional`, `flexmix`, `modeltools`, `spatstat.model`, `urltools`, `elliptic` |
| `params` | `ssdtools`, `S4Vectors` |
| `curve_params` | nothing |

`curve_params()` also reads correctly beside `show_params()`, which names the
parameters of each equation without fitting anything, and it does not
over-claim: the documentation states that the family's dispersion parameter, a
group-level term's standard deviation and the `ogl` offset are excluded, which
`parameters()` would have promised.

Nothing else changed. The rename is mechanical --- the generic, the four
methods, `R/curve_params.R`, `man/curve_params.Rd` and
`tests/testthat/test-curve_params.R` --- and the paragraph warning about the
`parameters` package collision is gone from the documentation, since there is
no longer a collision to warn about.

PR #334, opened 2026-09-13
resolve crf symbols in the formula's own environment (#319)

### LINKED ISSUE #319: crf() resolves its model argument in the global environment only, so a variable model set fails inside a function
## What

`crf()` resolves its `model` argument in its own frame, so a variable holding the
equation set is found only when that variable is in the global environment. The
same call fails inside a function, and fails under `knitr`.

## Why it matters

Building a model set programmatically is ordinary use --- a wrapper function that
fits several datasets with one set, a loop over candidate sets, a vignette chunk.
All of them fail, and the message names neither `crf()` nor the reason:

```r
f <- function() {
  eqs <- c("nec3param", "ecxll3")
  bnec(y ~ crf(x, eqs), data = d)
}
f()
#> Error in eval(mf[[m]]) : object 'eqs' not found
```

At the console the same two lines work, because the variable happens to be in the
global environment. That difference is what makes it hard to diagnose: the code a
user develops interactively stops working the moment it is wrapped in a function.

It also blocks #190. The `example8` precompile errored at its first fit for this
reason, and the release re-run cannot produce that vignette until either the
defect is fixed or every vignette writes its model set out literally.

## Evidence

R 4.6.1, `issue-6-33-grouping-vignette` at `77444911`, 2026-09-11.

Observed on the AIMS HPC precompile of `example8` (job 896328): the vignette set
`eqs <- c("nec3param", "ecxll3", "ecxll4")` on the first line of a chunk and
called `bnec(... ~ crf(log(conc), eqs) + ...)` on the third. Every fit in the
vignette failed with `object 'eqs' not found`, and `precompile.R` correctly
refused to ship the output.

Reproduced three ways: inside a function, inside an explicit `new.env()`, and
under `knitr`. It succeeds only when the variable is in the global environment.

<details>

### Where it is

`R/bayesnecformula.R:1153`:

```r
crf <- function(x, model, arg_to_retrieve = "x") {
  mf <- match.call(expand.dots = FALSE)
  ...
  } else if (arg_to_retrieve == "model") {
    m <- match("model", names(mf), 0L)
    eval(mf[[m]])
```

`eval()` with no `envir` evaluates in the frame of `crf()`, whose lexical parent
is the package namespace, then imports, then base, then the global environment.
The caller's environment is never on that chain, which is why the global
environment is the only place a variable is found.

`crf()` is reached from `get_model_from_formula()` at `R/bayesnecformula.R:1077`,
which builds the call as text and evaluates it, also without an environment:

```r
x_str <- paste0(substr(x_str, 1, nchar(x_str) - 1), ", \"model\")")
expand_model_set(eval(parse(text = x_str)))
```

### The fix

Two lines, and both are needed --- fixing either alone leaves the chain broken.
`bayesnecformula()` returns a formula, which carries an environment, so the
caller's environment is already available:

- `get_model_from_formula()`: `eval(parse(text = x_str), envir = environment(formula))`
- `crf()`: `eval(mf[[m]], envir = parent.frame())`

Demonstrated standalone, without touching the package, by reproducing the
namespace and caller environments:

| `crf` variant | result |
|---|---|
| `eval(mf[[m]])` | `Error: object 'eqs' not found` |
| `eval(mf[[m]], parent.frame())` | `nec3param, ecxll3` |

A literal model set still resolves under the fix, so the change is additive.

### What else to check

There are seven further `eval(parse(text = ...))` sites with no environment
argument --- `R/bayesnecformula.R` lines 404, 618, 747, 750, 872 and 1032, and
`R/bnec_hurdle.R` lines 508 and 512. They parse the predictor, the aterms and the
group-level terms rather than the model set. Whether any of them can meet a
variable that has to resolve in the caller has not been tested, and should be,
rather than fixing line 1160 alone and assuming the rest are safe.

### Tests

A test that a model set held in a variable fits from inside a function is the
one that would have caught this; a test written at the top level of a test file
would not, because `testthat` evaluates there in an environment whose parent
chain reaches the global environment. The test must be inside a function, or use
an explicit `new.env(parent = baseenv())`.

</details>

beckyfisher:
## What

A symbol a user writes in a `bayesnecformula` now resolves in the environment
the formula was written in, rather than only in the global environment. That
covers the model set held in a variable, a function used to transform the
predictor, and the variance function named in a `disp()` term.

Closes #319.

## Why it matters

`bnec(y ~ crf(x, eqs), data = d)` worked at the console and failed the moment it
was wrapped in a function, inside a knitted chunk, or in any environment that
does not inherit from the global environment. The message named neither `crf()`
nor the cause:

```r
f <- function() {
  eqs <- c("nec3param", "ecxll3")
  bnec(y ~ crf(x, eqs), data = d)
}
f()
#> Error in eval(mf[[m]]) : object 'eqs' not found
```

This blocks #190: `vignettes/precompile.R` calls `knitr::knit()` from inside
`knit_one()`, so a variable assigned in a chunk lands in that frame and not in
the global environment, and every fit in `example8` failed there.

## Evidence

R 4.6.1, this branch against `dev` at `c765c3ae`.

The defect was reproduced four ways and all four now succeed: inside a
function; inside `new.env(parent = baseenv())`; from a chunk knitted by
`knitr::knit()` called from inside a function, which is the shape
`precompile.R` uses; and with a character formula in each of those. Every added
behavioural assertion fails against `dev` and passes on this branch, checked by
stashing the `R/` changes and running the files again.

The issue asked whether the other seven `eval(parse(text = ...))` sites can meet
a variable that has to resolve in the caller. Six of them evaluate `crf()` with
its default `arg_to_retrieve = "x"`, which returns the deparsed predictor and
never forces the `model` promise, so no variable is looked up there. The seventh
is the one that was wrong. Six further sites, found by widening the search from
`eval(parse(text = ...))` to every evaluation of a formula term, had the same
defect and are fixed with it; they are listed below.

<details>

### The model set

Two changes, and both are needed.

`crf()` (`R/bayesnecformula.R`) evaluated `mf[[m]]` with no `envir`, which
evaluates in `crf()`'s own frame. That frame's lexical parent is the package
namespace, then the imports, then base, then the global environment; the
caller's frame is on none of them. It now evaluates in `parent.frame()`.

`get_model_from_formula()` built the `crf()` call as text and evaluated it with
no `envir`, so `parent.frame()` inside `crf()` was a frame in the namespace. It
now evaluates in a frame that inherits from `environment(formula)`. The issue
proposed `envir = environment(formula)` directly; that does not work, because
`crf()` is internal and is not exported, so it cannot be found from a user
frame. The evaluation frame binds `crf` and takes the formula's environment as
its parent, which resolves both the function and the user's variable.

The resolved value is now checked to be a character vector. The symbol resolves
in more environments than it did, so a name bound to something else in the
caller reaches `expand_model_set()`, which reported `'match' requires vector
arguments` and named nothing.

### The predictor and the other formula terms

The same defect on the other half of the `crf()` term, and on four further
paths. Each was verified to fail inside a function and to succeed at the
console before the change.

| site | what failed |
|---|---|
| `simplify_formula()` | the reduced formula `model.frame()` is built from, so `crf(squared(x), ...)` gave `could not find function "squared"` |
| `sub_x_transformation()` (`R/helpers.R`) | the back-transform in `ecx()`, `nsec()` and `expand_nec()` |
| `hurdle_raw_data()` | the predictor read back off a stored hurdle fit |
| `swap_response()` (`R/bnec_hurdle.R`) | the component formulas, rebuilt from deparsed text |
| `parse_disp_term()` (`R/disp_model.R`) | a variance function named by a variable, `disp(vf)` |
| `wrangle_model_formula()` | the `brmsformula` handed to \pkg{brms}, so `disp(~cent(x))` gave `could not find function "cent"` from `make_stancode()` |
| `check_disp_finite()` (`R/check_data.R`) | the term could not be evaluated, and the `try()` around it then skipped the check silently |

`sub_x_transformation()` is the one that reached furthest. It runs from
`expand_nec()`, which is called for every ecx-class equation, so on `dev` the
model frame refused the formula up front, and with only the `simplify_formula()`
fix in place the same formula would have compiled and sampled every model in the
set and then failed. Both halves are needed together.

`check_formula()` already refuses a `crf()` term naming more than one variable,
and that variable must be a column of `data`, so a *variable* in the predictor
cannot reach these paths --- only a function can.

### Character formulas

A character string has no environment of its own, and `as.formula()`'s default
gave it the frame of `bayesnecformula()`, again inside the namespace.
`bayesnecformula()` and `bnf()` gain an `env` argument defaulting to
`parent.frame()`, used only for the character branch and refused if it is not an
environment --- `environment(formula) <- NULL` is legal, so an unchecked `NULL`
would have silently restored the old behaviour. `bnec()`, `bnec_group()`,
`bnec_hurdle()`, `bnec_joint()`, `get_priors.formula()`,
`get_priors.character()` and `make_brmsformula()` pass their own caller's frame.

A string is a value, so it is bound at conversion time: a character formula
passed down through a wrapper binds at the last call, not where the string was
written. The `env` documentation says so and names `bnf(string, env = ...)` as
the way to choose.

### Consequence for saved fits

Binding an environment to a character formula means the formula holds a
reference to the frame it was converted in, and the formula is stored once per
model. Measured on a frame holding a 2-million-element vector, the stored
formula went from 463 bytes to 16 MB. A formula object has always behaved this
way, and a console call binds the global environment, which serialises by
reference, so only a character formula converted inside a function is newly
affected. Recorded in NEWS with the same workaround.

### Sites left unchanged

`single_model_formula()`, `swap_crf_model()`, `add_formula_glef()` and
`parse_group_terms()` go through `stats::update()`, which preserves the
environment; checked rather than assumed. `R/bnec_hurdle.R:456` evaluates a
`cens()` argument only where `all.vars()` is empty, so no symbol is resolved.
`R/inits_functions.R:86` uses `baseenv()` deliberately.

### Tests

Every new test is written inside a function on purpose: at the top level of a
test file `testthat` evaluates in an environment whose parent chain reaches the
global environment, which is the one place the defect did not show, so a
top-level test would have passed against `dev`.

There is no end-to-end `bnec()` fit among them. The tests cover every step up to
`brm()` --- `get_model_from_formula()`, `model.frame()`, `single_model_formula()`,
`make_brmsformula()`, `get_priors()`, `check_disp_finite()` --- and, for the
`brmsformula` itself, `brms::make_stancode()`, which generates the Stan program
without compiling or sampling it. The back-transform is asserted through
`ecx()` on the packaged `manec_example` fit, which fits nothing. A new fit would
add a Stan compilation to the suite for no additional coverage.

</details>

https://claude.ai/code/session_01WL5k1EwmtywMZ1NJi74iq2


beckyfisher:
## Review findings that were real problems

The branch went through three review rounds. Two findings were defects in the
first version of the fix, both of which would have shipped; one further defect
was found by `R CMD check` and by neither reader. The rest were refinements.

### The defects in the first version

The first version fixed the predictor only where the
model frame is built, and `sub_x_transformation()` resolves the same expression
again when an estimate is put back on the recorded scale. That runs from
`expand_nec()`, `ecx()` and `nsec()`, so `crf(squared(x), ...)` with `squared()`
defined in the caller stopped erroring at the start of a fit and started erroring
after every model in the set had compiled and sampled --- a worse failure than
the one being fixed. `make_brmsformula()` was an exported entry point that the
first version missed, while its own documented example uses a character formula,
which is the affected form.

### The defect found by R CMD check

`pkgload::load_all()` exports internal functions by
default, so every local run resolved `trials()` from the user's environment where
an installed package cannot. The reduced formula keeps `trials(tr)` as a call and
`bayesnec` does not export it, so all three check jobs failed with `could not find
function "trials"`. `crf` and `trials` are now bound in a frame whose parent is
the formula's environment, and a test pins that from an environment parented on
`baseenv()`, so it holds whether or not the internals have been exported.

### A consequence accepted rather than removed

Binding an environment to a
character formula means the formula holds a reference to the frame it was
converted in, and the formula is stored once per model. Measured on a frame
holding a 2-million-element vector, the stored formula went from 463 bytes to
16 MB. A formula object has always behaved this way and a console call binds the
global environment, which serialises by reference, so only a character formula
converted inside a function is newly affected. It is recorded in NEWS with
`bnf(string, env = ...)` as the way to choose a smaller environment.

<details>
<summary>The full list</summary>

Fixed, each with a test that fails against `dev`:

| finding | change |
|---|---|
| `sub_x_transformation()` resolved off the namespace chain | `enclos = formula_env(formula)`, `R/helpers.R` |
| `make_brmsformula()` was not among the entry points fixed | `bnf(formula, env = parent.frame())` |
| `parse_disp_term()` resolved `disp(vf)` in its own frame | `eval(arg, envir = formula_env(formula))` |
| a `disp()` sub-model still failed at `brms::make_stancode()` | `wrangle_model_formula()` sets the top-level `brmsformula` environment, which is the one `brms` reads |
| `check_disp_finite()` skipped a term it could not evaluate, so the check passed silently | the evaluation takes the formula's environment |
| `env = NULL` was accepted and restored the old behaviour | refused unless `is.environment(env)` |
| a `model` argument resolving to a function gave `'match' requires vector arguments` | refused, naming the class and length it resolved to |
| `formula_env()` handled a `NULL` environment but not `emptyenv()` | both return the global environment |
| `formula_eval_env()` bound nothing for a positional argument | `list2env()`, which refuses an unnamed element |
| the six caller-frame call sites were untested | a character-formula assertion through `get_priors()` |

Answered without a change: the character formula binds at the last call rather
than where the string was written, which is unavoidable for a value and is now
documented with the workaround; and the retention above.

Corrected in the text: the PR body no longer states a test count, two loose NEWS
sentences are rewritten, and NEWS now says the `disp()` finiteness check fires
where it previously passed silently.

</details>

PR #335, opened 2026-09-13
A cancelled devel job records a superseded commit, not missing R-devel coverage (#333)

### LINKED ISSUE #333: Two thirds of pull-request check runs never complete their R-devel job
## What

Two thirds of `pull_request` check runs on this repository never complete their
`ubuntu-latest (devel)` job. It is cancelled, not failed, so nothing reports and
no pull request gets R-devel feedback before it merges.

## Why it matters

R-devel is in the check matrix and is effectively absent from review. `dev` is
covered — the job completes on every `push` to `dev`, which is where a merge
lands — so this is a gap in pull-request feedback rather than in coverage of the
integration branch. The consequence is that a defect R-devel would catch is found
after merging rather than before, on a branch that has already been reviewed and
approved.

It also consumes investigation time. A cancelled devel job looks like a branch
problem, and reasoning about it that way is the natural error: on #331 four
consecutive runs had their devel job cancelled, which read as a branch-specific
hang until the base rate was measured. Anyone who sees this next will reason the
same way unless the behaviour is written down.

## Evidence

Of the last 36 completed `pull_request` runs of `R-CMD-check`, the
`ubuntu-latest (devel)` job was **cancelled 23 times and succeeded 11** — 68 per
cent cancelled, across `issue-297-parameters`, `issue-324-pull-best`,
`issue-219-workflow-vignette`, `issue-ci-test-runtime` and others.

Two properties combine to produce it. `concurrency.cancel-in-progress: true`
supersedes a run whenever the same pull request is pushed again, which is frequent
on a stacked set of branches. And devel is the slowest job in the matrix by a wide
margin: 42 to 56 minutes of CPU against `ubuntu-latest (release)`'s 31 to 51 on
the same runs, and 24 to 30 minutes elapsed. So when a supersede arrives, devel is
the job still running, and the other three have usually finished.

A cancellation arrives as exit 143 with no timing line, and the action's own
"Show testthat output" step reports `outcome=skipped`, so it leaves no record of
where the suite had reached. #331 adds `timeout-minutes` and an `if: always()`
dump of `testthat.Rout`, which makes a future termination distinguishable from a
hang, but does not change how often this happens.

<details>

### The choice to make

This is a design decision between three things that cannot all hold, and it should
be made deliberately rather than inherited:

1. `cancel-in-progress: true`, which exists for a good reason — the comment in
   `R-CMD-check.yaml` records that pushing a correction to a branch whose check
   was still running made both run to completion and only the second was ever
   read;
2. devel's length, which is what makes it the job in the line of fire;
3. wanting R-devel feedback before a merge rather than after.

Three options, from least to most work:

- **Accept it and say so.** Document that PR-time devel is advisory and that
  `dev` is where R-devel is checked. This changes no configuration and stops the
  next person investigating it as a defect. It is the status quo, written down.
- **Exempt devel from cancellation**, with its own workflow and concurrency
  group, or a per-job group. It then runs to completion on the first push and is
  superseded only by a later push to the same job. This spends runner time on
  superseded runs, which is what `cancel-in-progress` was added to prevent.
- **Make devel fast enough not to be the last one standing.** It is 25 minutes of
  CPU slower than release on identical work, which is itself unexplained and worth
  a measurement before anything is built on it.

### The inference to avoid

Do not read a cancelled devel job as evidence about the branch it ran on. That is
the error this issue exists to prevent, and #331 records how far it can be taken:
a 2×2 design, a pre-registered decision rule and three rounds of review were all
spent on an event with a 68 per cent base rate that nobody had measured.

</details>

beckyfisher:
Closes #333, as not a defect.

## What

#333 reported that the `ubuntu-latest (devel)` job is cancelled on about two
thirds of pull-request runs, and read that as a gap in R-devel coverage before
merge. The rate is real for the days it was measured on. The gap is not.

`notes/ci_devel_cancellation.md` records the measurement and
`notes/scripts/ci_devel_cancellation.R` reproduces it. The workflow itself is
almost untouched: two statements in it turned out to be false, and only those are
corrected.

## Why it matters

Of the 30 most recently merged pull requests, 30 had a successful
`ubuntu-latest (devel)` check on the head commit, completed before the merge.
None merged without one. A cancellation records that a newer commit exists on the
pull request, and the newer commit is the one checked.

That is an observation rather than a configuration: `dev` has no branch
protection and the repository has no rulesets, so devel is not a required status
check and nothing prevented a merge without it.

#333 exists because four consecutive cancelled devel jobs on #331 were
investigated as a branch-specific hang before anyone measured how often it
happens. The measurement is committed so the next person can look it up instead
of repeating that.

## Evidence

From `notes/scripts/ci_devel_cancellation.R` on 2026-09-13, over the 200 most
recent `pull_request` runs, of which 197 had all four jobs finished. Elapsed time
is `completed_at` minus `started_at` from the jobs API.

About seven in ten cancellations stopped all four jobs a median 12.1 minutes in,
before any had finished. Those runs discard nothing a later run does not supply.
The asymmetry #333 describes, three jobs reported and devel lost, is 10 runs of
197.

The rate is a step change, not a level: no devel job was cancelled on any of the
eight sampled days to 2026-09-06, and from 2026-09-09 the daily rate runs 33 to
70 per cent. It tracks push frequency — median 14.1 minutes to the next run on
the same branch where devel was cancelled, against 100.0 where it succeeded. So
#333's 68 per cent reads its own days accurately; there is no single rate to
quote against it, and the case here rests on the coverage result instead.

Devel is not the slowest job. Paired within a run, it is 2.4 minutes longer than
`ubuntu-latest (release)` (n = 125, Hodges-Lehmann +2.57, 95 per cent interval
1.53 to 3.61) and `windows-latest` is 6.4 minutes longer than devel. #333's 25
minutes was CPU time, not elapsed; its own CPU ranges give 5 to 11.

## The corrections to the workflow

Both are statements this measurement falsified, not records of the finding.

The comment at `timeout-minutes: 90` justified the bound as "well above the
longest observed job -- windows-latest at 42 minutes". The longest devel job to
reach a conclusion took 72.7 minutes. The bound is left at 90 — it still clears
the routine distribution by about half again, nothing has reached it, and the
`if: always()` log dump makes a timeout diagnosable — but the justification is
corrected.

The `push` filter's comment justified itself partly on "a repository that runs
one job at a time". Up to 16 `R-CMD-check` jobs ran at once in the window, and a
run's four cells start a median 0.1 minutes apart. The clause is deleted; the
reason for the filter is unaffected.

<details>

#### The diff

`.github/workflows/R-CMD-check.yaml`: 5 changed lines, comment-only, no
configuration touched.

`notes/ci_devel_cancellation.md` and `notes/scripts/ci_devel_cancellation.R`:
new. The script is base R plus the `gh` CLI, reading TSV from `gh --jq`, so it
adds no package dependency.

```bash
Rscript notes/scripts/ci_devel_cancellation.R 200 30
```

About five minutes. It prints the window it sampled, which advances as runs
accumulate, so a re-run will not reproduce these figures exactly.

#### The rejected alternatives

Exempting devel from cancellation adds nothing to the coverage above, and takes
more than the obvious per-matrix-cell concurrency group: a job-level group does
not override the workflow-level `cancel-in-progress`, which stops the whole run
first. Shortening devel changes which job is lost, not whether one is — closing
its margin leaves `windows-latest` as the job still running. Both are recorded in
the notes document with their figures.

#### Limits of the instrument

Runs whose jobs have not all finished are excluded, so the denominator is 197 of
200. The jobs endpoint defaults to `filter=latest`, so a re-run's original
attempt is invisible; three runs in the window had `run_attempt` above 1. The API
reports a supersede and a manual stop identically, so a supersede is inferred
from the timing of a later run. The run listing is fetched as two pages of 100,
which have been observed to disagree — one fetch returned 113 distinct runs
spanning nine months — and the script stops rather than reporting the short
sample.

</details>


beckyfisher:
## Corrections made during review

Four rounds of independent review. The conclusion held throughout — document the
behaviour, change no configuration — but a good deal of the evidence it rested on
did not. The faults that were real, grouped:

### The longest devel job

The 72.4-minute devel job was described as one that
never completed. It completed and it succeeded, which inverts what it says about
the timeout: the bound is tight against that observation rather than loose. That
in turn exposed a justification at `timeout-minutes: 90` that this PR's own
measurement falsifies — it was set against a 42-minute maximum. The comment is
corrected. The bound is left at 90, and the reasoning now runs from the
distribution rather than the maximum, because the maximum is one afternoon: seven
of the eight longest devel jobs started on 2026-09-06 and six of those failed.

### The censoring check

The sensitivity check said admitting
failures recovers the censored pairs. It recovers none of them: a cancelled job
has no duration, so a pair containing one is absent from both sets, and the two
counts of twelve were equal by coincidence and have since diverged. The censored
pairs are now bounded instead. A second version of the same error said that
admitting them "could only enlarge the estimate", which a per-pair lower bound
does not establish; three of the twelve bounds are negative.

### The sole-casualty claim

"Devel is not the job that is
still running", against a table recording devel cancelled alone in 10 runs and
`windows-latest` in 5. Devel is that job, twice as often. The accurate argument is
the stronger one: closing devel's margin would leave windows still running.

### The rebuttals to #333

#333's 25-minute figure was CPU time
and was answered with elapsed time, without the substitution being declared;
#333's own CPU ranges give 5 to 11 minutes. And the per-matrix-cell concurrency
rebuttal answered a configuration nobody proposed. The real obstacle is that a
job-level group does not override the workflow-level `cancel-in-progress`.

### The cancellation rate

No devel job was
cancelled on any of the eight sampled days to 2026-09-06; from 2026-09-09 the
daily rate runs 33 to 70 per cent. So #333's 68 per cent reads its own days
accurately, and the pooled figure that appeared to correct it is two regimes
averaged together. Both regime rates are now named separately and the case rests
on the coverage result instead.

### The counting errors

138 runs "completed every job" included four that were
still running. The dev-push denominator came from a different set than the table
below it. "Superseded" was asserted where the API reports only "cancelled"; it is
now inferred from the timing of a later push and checked.

### Unattributed figures

Three figures quoted were not printed
by the committed script. The mechanism — that cancellation follows push frequency
— was asserted rather than measured; it now is, at 14.1 minutes to the next run
on the same branch where devel was cancelled against 100.0 where it succeeded.
The censoring caveat appeared in the script's output and in neither document.
Runner minutes were treated as money on a public repository where they are
unbilled, and a floored subtraction was called a lower bound, which it is not.

### Silent failure paths in the script

A non-zero `gh` exit
returned an empty frame and was tabulated. `check-runs` was unpaginated against a
30-per-page default. A deduplication added mid-review turned an inconsistent pair
of API pages into a plausible 113-run sample at exit 0 — caught only because the
reviewer ran the script three times — and the first guard written for it still
passed one presentation of the same anomaly. An assertion aborted on a run whose
jobs had not been created yet. And two volatile figures reached the workflow file,
where they had already drifted within a day.

### Placement of the decisions

The two rejections and the timeout decision were inside
the `<details>` fold, which is where precision goes, not where a decision is
stated.

## The disputed finding

The claim that flooring the runner-time subtraction inflates the estimate. A job
cancelled past the median cannot contribute negative saved runner time, so the
unfloored sum is not a conservative alternative. The reviewer accepted this and
was right on the other half: the floored figure is not a lower bound either, and
the estimator now conditions on what is known about each cancelled job instead.

## Judgement calls, for the record

`timeout-minutes: 90` is left as it is. The corrected margin is about 1.24 times
the longest job observed, which is tighter than its comment claimed, but that
observation is one day of mostly failing checks, nothing of any cell has reached
90 in 583 completed observations, and the `if: always()` log dump makes a timeout
diagnosable. Raising it would be a change made against a projected rather than an
observed failure.

Two comments elsewhere in the workflow file are corrected although #333 does not
name them, because the same measurement falsifies both and leaving them would
reproduce the stale-account problem this PR exists to prevent.

The figures are in `notes/ci_devel_cancellation.md` rather than the workflow
comment, dated, because they go stale and a comment gives a reader no reason to
re-run anything.

beckyfisher:
Scope reduced, and the body above is rewritten to match.

The twenty-line record at the `concurrency` block is removed. The issue was
wrong, closing it says so, and an account of a non-problem in a config file was a
third copy of reasoning the notes document and this pull request already hold. It
would also have gone stale: two of its figures drifted within a day during review
and had to be pulled back out.

The workflow now changes in two places only, and neither documents the finding.
Both are statements the measurement showed to be false. `timeout-minutes: 90` was
justified as "well above the longest observed job -- windows-latest at 42
minutes"; the longest devel job to reach a conclusion took 72.7. The `push`
filter was justified partly on "a repository that runs one job at a time"; up to
16 jobs run at once. Five changed lines, comment-only, and the setting itself is
untouched.

The measurement and the script are unchanged, and the review summary above still
describes what those rounds caught.

PR #336, opened 2026-09-13
Do not reseed the initial-value search from the clock (#310)

### LINKED ISSUE #310: example3 does not reproduce run to run despite set.seed() before every fit
## What

Precompiling `example3` three times, with no change to the source, the image or
the node, gives a different rendered `.Rmd` each time. The difference is small
and confined, but it is there, and `set.seed(333)` sits on the line above every
`bnec()` call in the vignette.

## Why it matters

The rendered `.Rmd` is committed and is the published record of what the package
does (#190). A reader who reruns the vignette expecting the numbers in it will
not get them, and a diff between two renders cannot be read as a change in the
package.

## Evidence

Three runs of `./hpc/precompile-hpc.sh example3` on 2026-09-10, all on
one compute node: runs 2 and 3 used the same container image, run 1 its immediate
predecessor. Jobs 896236, 896242 and 896244.

416 of the 434 lines of `vignettes/example3.Rmd` are identical across all three.
What differs:

- the `fixef()` summaries of `exmp_f` and `exmp_g`, the two `avoid_data` fits at
  the end of the vignette, in the third or fourth significant figure. For
  example `nec_Intercept` estimate 1.1445531, 1.1453314, 1.1396885 across the
  three runs;
- the `rbind(fixed = nec(exmp_f), free = nec(exmp_g))` table that follows from
  them;
- all three `check_priors` figures, which differ in content rather than in
  metadata: 45,057 against 42,920 bytes, with no `tIME` chunk in either.

Everything else reproduces byte for byte, including the two-model set `exmp_c`
and `exmp_d` and the `amend()` that adds `necsigm`.

That the figures differ is the more informative half. `check_priors` draws from
the priors using R's random number stream, and those chunks sit after a fit
whose own summary is identical across runs. So the fit reproduced its posterior
and still left the stream in a different state, which means a `bnec()` call
consumes a number of random draws that varies between runs.

<details>

### Where to start

The two fits that differ are the only ones in the vignette on `avoid_data` with
`family = Beta(link = "identity")`, and `exmp_f` is the one with
`constant(0.5)` on `top`. Whether either of those is the relevant property is
not established; they are simply what distinguishes the two that differ from the
nine that do not.

The figure result points at init generation. `R/inits_functions.R` is the place
where a variable number of draws could be taken, and the retry path in
`R/helpers.R` around line 945 is worth reading first.

### What would settle it

Record `.Random.seed` immediately before and immediately after a single
`bnec()` call, repeated across runs. If the state after differs while the fit is
identical, the number of draws consumed is the thing to find; if the state after
is the same, the divergence is inside Stan and the seed brms passes it is the
thing to check.

This bears on #216. A stored seed is not on its own enough to reproduce a fit,
and that is the same class of problem.

### What this is not

It is not caused by the container. Two runs on the same image disagree in the
same way as two runs on different images, and the difference is not between the
container and a workstation.

</details>

beckyfisher:
Closes #310.

## What

The initial-value search called `set.seed(seed)` whatever it was given, and on
the default path `seed` is `NULL`. `set.seed(NULL)` does not leave the random
number stream alone: it re-initialises it from the clock and the process id. So
every `bnec()` call discarded whatever `set.seed()` the user had run, and drew
its initial values fresh.

The search now seeds itself only where a seed was supplied. `NA` --- which is
how `brms` writes "no seed" --- is read the same way as `NULL`.

## Why it matters

Two things follow from one line.

A fit did not reproduce under a fixed seed, in the same session, on the same
machine. That is #310's own finding, and it is not confined to the two fits the
issue names: it holds for every `bnec()` call.

`vignette("example3")` runs `set.seed(333)` before each of nine fitting chunks,
which produce eleven individual fits, and passes no `seed`. The initial-value
search for each fit therefore discarded the stream it was handed and began
from a fresh draw. That alone accounts for both outputs that differed between
renders. The two `fixef()` tables are read off two of those fits. The three
`check_priors()` figures are pure functions of the fits they plot:
`check_priors()` calls `brms::hypothesis()`, whose only `set.seed()` is guarded
by `if (!is.null(seed))` and which bayesnec passes no seed, and then
`geom_density()`. It touches the stream nowhere else. So a figure differs exactly when its fit does, which is the
same statement as the issue's byte-identical PNGs from one saved fit.

Separately, how many proposals the search makes depends on the stream it starts
from, so a search begun from the clock also left the stream in an unpredictable
place, and any random operation after the fit differed for that reason as well.

`bnec(..., seed = NA)` also stopped with "supplied seed is not a valid integer",
because `set.seed(NA)` is an error. That is `brms`'s own documented way of
saying no seed.

## Evidence

R 4.6.1, `brms` 2.23.0, `rstan` 2.32.7, backend `rstan`, one workstation,
`nec3param` fitted to the packaged `nec_data` with `Beta(link = "identity")`,
`iter = 1000`, `chains = 2`. Two `bnec()` calls in one session, each preceded by
`set.seed(333)`. The released arm is the same script run against `origin/dev`'s
`R/inits_functions.R`, so the two arms differ in that one line and nothing else.

| | released | this branch |
|---|---|---|
| `fixef()` agrees between the two calls | no | yes, every digit |
| difference in `nec` | 9.1e-5 | 0 |
| difference in `top` | 5.1e-4 | 0 |
| difference in `beta` | 3.1e-3 | 0 |
| `sum(.Random.seed)` after the fit agrees | no | yes |

The same pair of calls for the two-model set `c("nec3param", "nec4param")`, on
this branch: the stacking weights agree at 0.8565476 and 0.1434524, the
`w_draw_seed` that realises the model-averaging draw agrees at 1343037611, and
`sum(.Random.seed)` after the fit agrees. That covers the LOO weighting and the
model-averaging draw of #216 as well as the fits.

The size of the disagreement in the released arm is a property of the sampler,
not a measure of how wrong the released estimates were: it says the two runs are
different fits.

`bnec(..., seed = NA)` on the same data raises "supplied seed is not a valid
integer" against `origin/dev` and completes on this branch.

<details>

### The change

`R/inits_functions.R`, in `make_good_inits()`:

```r
if (!is.null(seed) && !(length(seed) == 1 && is.na(seed))) {
  set.seed(seed)
}
```

Written on `length(seed) == 1` so that only `NULL` and a scalar `NA` are read
as "no seed". A seed of any other shape reaches `set.seed()` and is rejected
there with R's own message rather than by the condition.

No other site in the package can reach `set.seed(NULL)` on a default path.
`grep -rn "set.seed" R/` gives four others: `check_fit()` at two lines and
`dispersion()`, both defaulting to `seed = 10`, and `weighted_draw_index()`,
which supplies `216` for an object saved without one.

Two of those are worth naming rather than passing over. `check_fit()` does not
validate `seed`, so `check_fit(x, seed = NULL)` reaches `set.seed(NULL)` --- the
same defect, on a path the user has to ask for; `dispersion()` cannot, having
`chk_number(seed)`. And neither `check_fit()` nor `dispersion()` restores the
stream, so a diagnostic silently resets a user's simulation seed. Both are
left alone here: the guard is one line but the restore is a behaviour change
that needs its own NEWS entry, and neither is #310. Filed as #337.

### The stream after the search

`weighted_draw_index()` saves and restores the caller's `.Random.seed`, and the
search deliberately does not. The search is part of fitting rather than a
summary computed from a fit, and the Stan seed is drawn from the same stream
immediately afterwards --- `rstan:::check_seed()` and
`cmdstanr:::maybe_generate_seed()` both call `sample.int(.Machine$integer.max, 1)`
where no seed is given. A fit is meant to advance the stream. What it must not
do is advance it by an amount nobody can predict.

### The RNG under a future plan

`bnec_model_lapply()` recorded that `RNGkind()` re-initialises `.Random.seed`
from the clock and the process id, and used that to explain why workers do not
draw in step. It does so only where no seed exists yet --- which is what
`?Random` documents, and what makes `set.seed(NULL)` the cause of #310. Where a
seed is already present, which inside a `future.seed = TRUE` worker it always
is, the new state is derived from the current one:

```
set.seed(42); RNGkind(<same kind>)  ->  state -38435688197, first draw 0.8311705
three calls in one process, and again in a second process: identical
```

So a run under a plan reproduces as a sequential one does. Measured on the real
path --- `plan(multicore, workers = 3)`, three equations, the body being
`add_brm_defaults()` and so the initial-value search itself, no `seed` supplied
--- two runs at one `set.seed()` gave identical initial values and a third at
another seed gave different ones. What makes the workers differ from each other
is the per-element L'Ecuyer-CMRG stream `future.seed = TRUE` installs, not the
clock; the 2026-09-11 measurement recorded there is consistent with either
mechanism and did not distinguish them.

One backend and one R version, so this is a measurement rather than a
guarantee, and `seed` remains the way to fix a run that has to repeat across
either. The corrected reasoning is in `bnec_model_lapply()`, and
`test-parallel_models.R` pins the `RNGkind()` behaviour the argument rests on.

A parallel run matching a *sequential* one is a different question, and there
nothing changed. Same harness, run against the released code as well:

| | released | this branch |
|---|---|---|
| no seed, two sequential runs agree | no | **yes** |
| no seed, two parallel runs agree | no | **yes** |
| no seed, sequential agrees with parallel | no | no |
| `seed = 99`, sequential agrees with parallel | yes | yes |

So a parallel run still matches a sequential one only under a `seed`, as it
always did, and what this PR changed is that each run now repeats itself.
`?bnec` states both, and `bnec_model_lapply()`'s "the fitted models reproduce
exactly" --- written when a seed was assumed --- is qualified rather than left
to be read as covering the seedless case.

The model-averaged quantities are not reproduced between a sequential and a
parallel run either way, with a seed or without. That part is unchanged.

### Tests

`tests/testthat/test-inits_functions.R` gains four blocks.

"with no seed the search follows the caller's stream" asserts three properties
together, because a fix satisfying any two alone is wrong: the initial values
repeat under one session seed, the stream is left where a second run leaves it,
and a different session seed gives different values. The third is what stops a
hardcoded seed standing in for the caller's.

"seed = NA is read as no seed rather than erroring" covers the `brms` spelling
and asserts that a seed of any other shape, `integer(0)` included, still
reaches `set.seed()` and is rejected there.

"the seed reaches the search from brm_args, NA included" asserts the same at
the `add_brm_defaults()` level, which is the only route by which `NA` can
arrive, and so is the plumbing `bnec(..., seed = NA)` uses. It builds no Stan
program.

"both hurdle blocks follow the caller's stream" covers
`make_good_hurdle_inits()`, the one call site that passes the seed on twice.

All four restore `.Random.seed` on exit, through a helper that reads it via
`exists()` so a block run on its own does not error. Now that the search
follows the ambient stream, a block that advances it would otherwise decide
where the next block in the file starts.

`test-parallel_models.R` gains "RNGkind at the same kind derives the state, it
does not reseed", which pins the behaviour the parallel reasoning rests on.

The affected files pass (`NOT_CRAN=true`), as do `test-check_priors.R`,
`test-hurdle_family.R` and `test-helpers.R`.

### The rendered vignettes

The committed rendered `.Rmd` files are not regenerated here. Re-rendering them
is an HPC precompile run and the numbers in them will change --- they are drawn
from a stream that was being reset. That is a separate, deliberate step.

</details>

https://claude.ai/code/session_01NoaBZRw339tWq1oRRX9oAZ


beckyfisher:
Body updated for the review: the *example3* explanation in **Why it matters** was wrong and has been replaced, and the sentence about other `set.seed()` sites in the collapsed block now names what it found. A reviewer who read the earlier text is looking at different words.

beckyfisher:
Body updated again. The section on the `future` plan said a plan needs a `seed`; that was wrong and is now the opposite, with the measurement. The tests section changed with it.

beckyfisher:
## Findings of the review cycle

Five rounds with an independent reviewer. The fix itself --- the guard in
`make_good_inits()` --- is unchanged since the first commit and was never
contested. Everything below is what the review found around it.

### Claims withdrawn

The explanation of the `example3` figures. The first version said the three
`check_priors()` figures differed between renders because the fit left the
random number stream in a different place. That cannot be: `check_priors()` is
`brms::hypothesis()` then `geom_density()`, and `hypothesis.brmsfit()`'s only
`set.seed()` is guarded by `if (!is.null(seed))` with `bayesnec` passing none,
so the figure is a function of the fit alone. The figures also sit above both
`fixef()` calls rather than below them. The correct account, now in NEWS and in
the roxygen, is that all eleven `set.seed(333)` calls in the vignette were
discarded, every fit was a fresh draw, and both the tables and the figures
follow from that.

A warning that a run under a `future` plan would not reproduce, added in
round two and removed in round three. It rested on `bnec_model_lapply()`'s own
documented reason, that `RNGkind()` re-initialises `.Random.seed` from the clock
and the process id. It does that only where no seed exists yet, which inside a
`future.seed = TRUE` worker it never is. Measured on R 4.6.1, on the same-kind
call and on the L'Ecuyer-CMRG to Mersenne-Twister change a worker makes:
identical first draws over repeated calls in one process and in a second,
0.8311705 and 0.2046757. And on the real path, `plan(multicore, workers = 3)`,
three equations, the initial-value search as the body and no seed supplied, two
runs at one session seed gave identical initial values. The gap was not there.
The message is gone and the inherited reasoning is corrected in place, with a
test pinning the `RNGkind()` behaviour the argument rests on.

The claim that the sequential-parallel mismatch was new. It is not. Run against the
released code with the same harness: sequential disagrees with parallel without
a seed and agrees with one, before this change and after it. What changed is
that each run now repeats itself, where before neither did.

### Gaps in the change itself

`all(is.na(integer(0)))` is `TRUE`, so the first guard read an empty seed as "no
seed". It is now `!(length(seed) == 1 && is.na(seed))`, and a seed of any other
shape reaches `set.seed()` and is rejected there.

`seed = NA` can only arrive through `brm_args$seed`, and nothing exercised that
route, though the PR advertised `bnec(..., seed = NA)` as fixed. Covered at the
`add_brm_defaults()` level, building no Stan program.

`make_good_hurdle_inits()` forwards the seed to both blocks and was untested.
Covered.

The new tests advance the stream the search now follows, so they decided where
the next block in the file started. They save and restore `.Random.seed` through
`exists()`, and the generator kind as well where they change it.

### Matters of record

`check_fit()` and `dispersion()` reset the caller's stream without restoring it,
and `check_fit()` accepts `seed = NULL` and so reaches `set.seed(NULL)` --- this
issue's defect in a second place. Not fixed here, because the missing guard and
the restore are one defect in two functions and the restore needs its own NEWS
entry. Filed as #337, referenced from the sentence in `R/inits_functions.R` that
states the principle they break.

`?brms::threading` records that adaptive within-chain scheduling prevents exact
reproducibility even under a fixed seed. Noted in `?bnec` with the pointer.

R parses `NEWS.md` with plain commonmark, which has no table extension, so a GFM
table reaches `utils::news()` as a run of pipes. The comparison above is prose
in NEWS and a table only here.

### Not done, deliberately

The committed rendered vignettes are not regenerated. Re-rendering is an HPC
precompile run and the numbers in them will change, since they were drawn from a
stream that was being reset. That is a separate step.

beckyfisher:
## Independent follow-up review

A fresh independent review after merge found two genuine low-severity problems in #336. Both are corrected in #358:

- The record conflated nine fitting chunks in `example3` with the eleven individual fits they produce. NEWS, the internal documentation and this PR body now use the measured counts.
- The record said base R rejects every non-scalar seed, but R 4.6.1 silently accepts `set.seed(c(1, 2))` and uses the first value. `make_good_inits()` now validates explicit seeds before searching, and the test covers both empty and length-two values.

The reviewer independently reran the omitted, scalar-`NA`, changed ambient, empty, length-two and character seed cases after the corrections and found no remaining defect. The executable fix for #310 itself was not disputed.

Issue #313, opened 2026-09-11
The vignettes are precompiled with cmdstanr; CI still uses rstan


beckyfisher:
## What

As of #308 every vignette is precompiled with the `cmdstanr` backend rather than
`brms`'s `rstan` default. The GitHub Actions precompile workflow is held at
`rstan`, because its runner has neither `cmdstanr` nor a cmdstan installation. So
a vignette rebuilt on the HPC and one rebuilt in CI are sampled by different back
ends. This issue records that decision and holds the open question of whether to
close the gap.

## Why it matters

Two reasons to decide rather than leave it.

A render is the published record of what the package does (#190). Two renders
produced by different back ends are not comparable, so a diff between them cannot
be read as a change in the package, and which route produced a given render is not
visible in the file.

And the reason for using `cmdstanr` at all only partly applies in CI. `cmdstanr`
names a compiled Stan program after a hash of its source and keeps the executable
beside it, so the programs are reusable between runs; `rstan` compiles in process
and keeps nothing. That cache is the largest saving available on the HPC. In CI it
would also make the compile step cacheable, which #230 and #251 both wanted and
neither could have while the backend was `rstan`.

## Evidence

The decision was taken deliberately, not by omission, and it was two lines:
`options(brms.backend = Sys.getenv("BAYESNEC_BACKEND", "cmdstanr"))` in
`vignettes/precompile.R`. Everything else attributable to it is consequence:

- `open_progress = FALSE` is an `rstan::sampling()` argument. Under `cmdstanr`
  `brms` raises `unused argument (open_progress = ...)`, so **every** model failed
  and `bnec()` reported only `None of the models fit successfully`. It was removed
  from 20 `bnec()` calls across example1 to example4 on #308.
- `rstan::rstan_options(auto_write = TRUE)` is inert under `cmdstanr` and was
  removed from example2 and example6.
- `.github/workflows/precompile-vignettes.yaml` needed
  `BAYESNEC_BACKEND: rstan`, or every leg of it would have failed the same way.
- the committed renders of example1, example2 and example4 are now older than
  their sources and their numbers will change when they are next built. #190 is
  where the set is brought up to date together.

`example7` is unaffected: it already set `options(brms.backend = "cmdstanr")` in a
chunk of its own, which is why #306 needed no backend change to get the cache for
the vignette it was aimed at.

<details>
<summary>Implementation detail</summary>

### The question to decide

[...]

PR #339, opened 2026-09-14
Precompile with cmdstan in CI as well as on the HPC (#313)

beckyfisher:
## What

The precompile workflow installs `cmdstanr` and cmdstan and no longer sets
`BAYESNEC_BACKEND: rstan`, so a vignette rebuilt in CI is sampled by the same
back end as one rebuilt on the HPC. The compiled Stan programs are cached
between runs, which is what #230 and #251 both wanted and neither could have
while the back end was `rstan`.

Of the two options #313 puts, this is the first: close the gap rather than
accept it. The reason is the one #313 gives for opening it — a render is the
published record of what the package does (#190), and two renders produced by
different samplers cannot be diffed as a change in the package.

## Why it matters

A `.Rmd` in `vignettes/` does not say which route produced it. Before this
change, a reader comparing the rendered `example1` on `dev` with the one a
release run produced could not tell a change in `bayesnec` from a change of
sampler. The remaining route-dependent difference is the R package versions,
which `hpc/README.md` states.

One vignette changes what it does in CI. `example5` is a `cmdstanr`
installation guide whose chunks are evaluated; its leg used to fail at
`library(cmdstanr)` and upload nothing, and it now renders, substituting the
runner's cmdstan path for the contributor's in the committed output. Neither is
right for the Windows reader that vignette is written for, and the chunks that
do it are #342.

The second consequence is a cache. `rstan` compiles in process and keeps
nothing, so a leg recompiled every Stan program it needed on every run; under
`cmdstanr` a program is named after a hash of its Stan source and the compiled
executable is kept beside it, so a second run of the same vignette reuses what
the first one built. An Actions cache is scoped to the branch, so that reuse is
between runs on one ref and not between branches, which is where it differs from
the cache `hpc/run.precompile` keeps on shared scratch.

## Evidence

Measured for this change rather than asserted.

- **A cached program is about 2.9 MB.** Measured on a `bayesnec` Stan program
  cache of 267 compiled programs: 780 MB, which `du -sm` reports as 744 MiB, one
  executable 2.83–2.94 MB. A 40-file sample compressed 2.96:1 under
  `tar | zstd -3`, which is how `actions/cache` stores it. The 1500 MiB budget in
  the workflow is roughly twice that cache.
- **`CMDSTAN` pins the installation.** `cmdstanr:::cmdstanr_initialize()` reads
  the variable at load and calls `set_cmdstan_path()` with it; with it unset,
  `cmdstan_default_path()` takes the highest version found under `~/.cmdstan`.
  Run here with `CMDSTAN=~/.cmdstan/cmdstan-2.39.0`,
  `as.character(cmdstanr::cmdstan_version())` returns `"2.39.0"`, which is the
  string `hpc/image.lock` records and the string the workflow compares against.
- **The version is read from the lock, not written out.** `hpc/image.lock`
  records `cmdstan: 2.39.0`; the `plan` job parses that line and fails the run
  if it is absent, so the two routes cannot drift apart silently.
- **`actions/cache` declares `post-if: success()`.** Read from the action's own
  `action.yml`. A leg that failed part way through a vignette would therefore
  have saved none of the programs it had compiled, so restore and save are
  separate steps here and the save runs unless the run was cancelled.
- **A GitHub Actions cache is restorable from the current ref or the default
  branch only.** Quoted from the dependency-caching documentation. `master` is
  the default branch and this workflow is dispatched on `dev` or on a
  `precompile/**` branch, so entries warm repeated runs against one ref and
  nothing else.
- **The shell in the new steps was run.** The version parse, the toolchain key
  (`cmdstan2.39.0-gcc15.3.0` on this machine), the vignette selection, and the
  budget check in each of its four cases — no directory, empty directory, under
  budget, over budget — were executed outside a runner. The workflow YAML parses.

The workflow has now been run, on `example4`, in
[run 34800367969](https://github.com/open-AIMS/bayesnec/actions/runs/34800367969):
23 minutes end to end, every job successful. `plan` resolved `cmdstan 2.39.0,
g++ 13.3.0` from the lock. The `cmdstan` job missed its cache, as the first run
on a ref must, and installed in 3 m 09 s including the R setup. The leg restored
that installation, and `Check the backend` reported `cmdstan 2.39.0 at
/home/runner/.cmdstan/cmdstan-2.39.0` with `brms 2.23.0 | cmdstanr 0.9.0.9002`.
`example4` knitted in 15.6 minutes on four cores and left 40 MiB in 28 files in
the program cache, which is the 14 programs `hpc/README.md` records for that
vignette with its `.stan` source beside each one. The entry was saved as
`stan-Linux-cmdstan2.39.0-gcc13.3.0-example4-34800367969-1`.

Two things that run does not establish. The cache has not been shown to warm,
which needs a second run against the same ref. And `brms` was 2.23.0 on the
runner, which is the version `hpc/image.lock` pins, so the two routes happened
to agree on it here; that is a coincidence of timing rather than a guarantee,
and the reason for the paragraph above stands.

<details>

### The change

`.github/workflows/precompile-vignettes.yaml`

1. The `plan` job reads `cmdstan:` from `hpc/image.lock` and `g++
   -dumpfullversion -dumpversion`, and outputs the version, a toolchain key and
   a `dry_run` flag. Computing the key once and passing it to every job is what
   stops the fan-out and the install job disagreeing about it.
2. A new `cmdstan` job installs `cmdstanr` from r-universe and
   `install_cmdstan(version = <lock>)`, behind an `actions/cache` on
   `~/.cmdstan`. It runs before the fan-out because a cold install is 10–15
   minutes: every leg would otherwise run it and then contend to write one cache
   entry, and all but one would report `Unable to reserve cache` having done the
   work anyway. `timeout = 2400` because `install_cmdstan()` defaults to 1200 s
   and a cold build on four cores is close to that.
3. The job runs on a dry run and does nothing, its steps being conditioned
   rather than the job. A skipped job skips everything that needs it, so
   conditioning the job would skip the fan-out.
4. Each leg restores `~/.cmdstan` with `fail-on-cache-miss: true`, before the R
   setup because it is the cheapest hard failure in the job. Failing on a miss
   is deliberate: falling back to `rstan` would produce exactly the render this
   issue exists to prevent. `CMDSTAN` and `BAYESNEC_STAN_CACHE` are then set
   through `$GITHUB_ENV`.
5. The `cmdstanr` package itself comes from `setup-r-dependencies` as
   `stan-dev/cmdstanr`, which is the spelling `R-CMD-check.yaml` already uses,
   so the two workflows install one package from one source and this one is
   restored with the rest of the library rather than rebuilt every run.
6. A `Check the backend` step compares `cmdstan_version()` with the lock and
   prints the `brms` and `cmdstanr` versions, before the hold-back and the
   knit, so a toolchain fault is found in seconds rather than after twenty
   minutes of sampling.
7. `actions/cache/restore` and `actions/cache/save` on
   `$RUNNER_TEMP/stan-programs`, keyed
   `stan-<os>-<toolchain>-<vignette>-<run_id>-<run_attempt>` with the prefix as
   the fallback. They are separate actions rather than one `actions/cache`
   because that action declares `post-if: success()` and so saves nothing from a
   failed job, which is the job that has most to save. The run id makes a leg
   save at all, a cache entry being immutable; the attempt number makes a rerun
   of a failed leg save, since a rerun keeps the run id.
8. The fallback key does not reach across vignettes, although the vignettes do
   share programs. A leg saves whatever it restored plus whatever it compiled,
   so a shared fallback would have every leg save the union of every vignette's
   set.
9. A `Bound the Stan program cache` step decides what is saved and how large it
   may be, and the save also requires that the `Precompile` step ran, so a leg
   that failed before the knit does not re-upload the set it restored under a
   newer key. It reports nothing to save when the directory is absent or empty, so
   that a leg which failed before compiling anything cannot store an empty entry
   under a key newer than the good one — a prefix fallback resolves to the
   entry created most recently, so that one empty entry would leave the next run
   with none of that vignette's programs. Over 1500 MiB the directory is emptied
   and a stamp file left in it, which is what supersedes the oversized entry.
10. `open-pr` no longer runs when the `cmdstan` job failed. Every leg is then
    skipped and `no leg produced any output` would name the symptom.

`vignettes/precompile.R`, `hpc/README.md`, `NEWS.md` — the comment and the
section that stated the divergence, and an entry for the change.

#### The 1500 MiB budget

The figure is roughly twice the largest measured `bayesnec` program cache and is
arbitrary anywhere above about 800 MiB. What consumes the repository's 10 GB
cache quota is not one entry but their accumulation: the key includes the run
id, so no entry is ever replaced and each leg of each run adds one. A leg at the
budget adds about 530 MB compressed, so a whole-set run at the budget would add
several gigabytes. Stale entries are
never restored again, so they are the first the least-recently-used eviction
takes and they expire unread after seven days; the bound exists so that a single
runaway entry cannot displace the dependency caches `R-CMD-check` restores.
Nothing about the science depends on it: emptying the cache means compiling
again and changes no number, because a program is reused only when its Stan
source hashes the same and the sampling is re-run either way.

#### `example5` is left in the fan-out

A first version of this branch held `example5` out of a whole-set run, on the
grounds that a Linux render of a Windows guide is wrong for its readers. That
was withdrawn on review, for three reasons. The committed
`vignettes/example5.Rmd` already shows `/Users/dbarneche/.cmdstan/cmdstan-2.33.1`
and `Warning: Path not set. Can't find directory: C:/cmdstan`, so the render the
hold-back prevented is the render already shipped. `hpc/precompile-hpc.sh`
defaults to every `.Rmd.orig` with no exclusion, so the hold-back would have
made the two routes rebuild different sets, inside the change whose purpose is
to make them agree. And the dispatch form says "Leave blank to rebuild all of
them", which the hold-back falsified at the moment an operator reads it.

The defect is that an installation guide evaluates machine-specific chunks at
all. That is #342, with its own render for review.

#### The `cmdstanr` package version

The cmdstan version is pinned and the R package version is not. The image holds
`cmdstanr` 0.9.0; both `stan-dev/cmdstanr` and the r-universe build serve the
current state of that same source. `cmdstanr` does not sample — it writes the
`.stan` file, invokes `make` and reads the output — so the sampler the two
routes share is cmdstan 2.39.0. Pinning the wrapper would need a source install
from a tag and would change nothing about the output. The `cmdstan` job takes it
from r-universe rather than by the `stan-dev/cmdstanr` spelling because it has
no other reason to install pak or resolve `DESCRIPTION`.

#### The `brms` version

`brms` is still not pinned in CI, and that is the difference that remains after
this change. `brms` writes the Stan program, so a different `brms` is a
different program rather than a different sampler. Closing that means pinning
the whole R library to `hpc/image.lock`, which is a larger change than this
issue asks for. It is stated in `hpc/README.md` as the reason the HPC remains
the route for a render whose numbers will be quoted.

#### The cache does not cross branches

An Actions cache is restorable from the current ref or the default branch, and
this workflow runs on neither `master` nor anything that inherits from it, so
the program cache and the cmdstan installation warm repeated runs on one ref and
nothing else. The first run on a new branch pays the 10–15 minute cmdstan build
and compiles every program cold. Seeding the entries would need a run on
`master`, which is a change to when this workflow runs rather than to what it
does, and is not made here.

### Checks before merge

```
gh workflow run "precompile vignettes" --ref issue-313-ci-backend -f dry_run=true
gh workflow run "precompile vignettes" --ref issue-313-ci-backend -f vignettes=example4
```

The first exercises the plan and the conditioned-out toolchain job in about a
minute. The second is the real test: `example4` compiles 14 programs and took
13 m 58 s cold on the HPC, so it exercises the install, the program cache and
the knit without committing hours. Run it twice to see the cache warm — the
`Bound the Stan program cache` step prints the size and the file count.

</details>

https://claude.ai/code/session_011vWCqYtbT4by9mzPD8PC5D


beckyfisher:
The body has been updated and no longer says what it said when this pull request
was opened. Three review cycles ran against it and the branch is now at
`5b91946d`; what follows is what was found to be wrong, so that a reviewer
arriving at the body can see what it replaced.

## Defects in the code

- A failed leg stored an empty program cache, and the next run would have
  started cold. `!cancelled()` drops the implicit `success()`, so the save step
  ran after any earlier failure; a leg that failed at the cmdstan restore or the
  backend check had compiled nothing, saved an empty directory under a key newer
  than the good one, and a prefix fallback resolves to the entry created most
  recently. One failed leg would have left the following run with none of that
  vignette's programs, which is the saving this change exists to obtain. The
  save is now gated on there being something to save and on the knit having run.
  Where the directory is emptied for exceeding its budget a stamp file is left in
  it, so that the emptied state still supersedes the oversized entry rather than
  leaving it the newest match for the next run to restore and empty again.
- `actions/cache` declares `post-if: success()`, so a leg that failed part way
  through a vignette saved none of the programs it had compiled. Restore and save
  are separate steps now.
- The `example5` hold-back was justified against a defect already shipped. A
  first version of this branch kept `example5` out of a whole-set run, because a
  Linux render of a Windows installation guide misleads its readers. The
  committed `vignettes/example5.Rmd` already shows
  `/Users/dbarneche/.cmdstan/cmdstan-2.33.1` and `Warning: Path not set. Can't
  find directory: C:/cmdstan`, so the render the hold-back prevented is the
  render the repository ships today; and `hpc/precompile-hpc.sh` excludes
  nothing, so the hold-back would have made the two routes rebuild different sets
  inside the change whose purpose is to make them agree. Withdrawn. The evaluated
  chunks that cause it are #342.
- `max(1L, parallel::detectCores())` is `NA` when `detectCores()` is, which would
  have reached `make` as `-jNA`. Now with `na.rm = TRUE`.
- A rerun of a failed leg saved nothing: "Re-run failed jobs" keeps the run id,
  so the key already existed and the save was declined with a warning. The key
  includes `github.run_attempt` now.
- A corrupt `~/.cmdstan` entry would have failed every rerun identically, the key
  being deterministic and nothing evicting it. The check that finds it now names
  the remedy and the exact key.

## Claims that were wrong

- The Actions cache does not work the way `hpc/run.precompile`'s does, and both
  the body and `hpc/README.md` said it did. A run restores entries created on its
  own ref or on the default branch; this workflow is dispatched on `dev` or a
  `precompile/**` branch and `master` is the default, so entries warm repeated
  runs against one ref and nothing else, and the first run on a new branch is
  cold in every leg including the 10 to 15 minute cmdstan build.
- Three hours is this workflow's own `timeout-minutes` and not a runner limit,
  and the claim that the largest vignettes exceed it was unattributed and
  contradicted the comment three lines above it. The sentence is gone.
- The budget bounds one entry, and one entry is not what consumes the quota: the
  key includes the run id, so no entry is replaced and each leg of each run adds
  one. Stated per leg now, so that it does not go stale with the vignette count.
- `du -sm` reports MiB, and the measurement mixed MiB with MB and did not
  reconcile. It is 267 programs, 780 MB, which is 744 MiB, about 2.9 MB each.
- The toolchain key does not work the way `hpc/run.precompile`'s does either.
  That job keys on the image digest rather than a compiler version string, and
  the `g++` version here is read on the `plan` runner while the programs are
  built on others. The comment gives the analogy and the residual gap.

## Changed on review without having been wrong

`cmdstanr` comes from `setup-r-dependencies` as `stan-dev/cmdstanr`, the spelling
`R-CMD-check.yaml` already uses, so the two workflows install one package from
one source and this one is restored with the rest of the library. The cmdstan
restore now runs before the R setup, being the cheapest hard failure in the job.
Four comments said "nine" where seven `.Rmd.orig` files exist. The
`BAYESNEC_BACKEND` paragraph now sits above the `env:` mapping it explains,
not inside it.

## Left undone

The workflow has never run. The two dispatches named at the end of the body are
worth doing before merge, and `example4` twice is the only way to see the cache
warm. #342 should land before the next whole-set render, or that render commits
an `example5` with the runner's path in it.

A leg whose cache is already complete re-uploads it under the new key having
compiled nothing. That is deliberate, and the reasoning is in the `Bound the
Stan program cache` comment: one leg holds one vignette's programs, `example4`
being about 14 MB compressed, and the stale copy expires unread after seven
days. Detecting it would be justified if a vignette the size of `example7`
reaches this branch.

beckyfisher:
The workflow has been run and the body records the result. It passed on
`example4` in [run 34800367969](https://github.com/open-AIMS/bayesnec/actions/runs/34800367969),
23 minutes end to end, every job successful.

| stage | result |
|---|---|
| `plan` | `cmdstan 2.39.0, g++ 13.3.0` read from `hpc/image.lock`; `Rebuilding: ["example4"]` |
| `cmdstan` | cache miss, as the first run on a ref must be; installed in 3 m 09 s including the R setup |
| `Check the backend` | `cmdstan 2.39.0 at /home/runner/.cmdstan/cmdstan-2.39.0`, `brms 2.23.0 \| cmdstanr 0.9.0.9002` |
| knit | `example4.Rmd.orig knitted in 15.6 mins`, four cores |
| `Bound` | `Stan program cache: 40 MiB in 28 file(s)`, against a 1500 MiB budget |
| `Save` | `stan-Linux-cmdstan2.39.0-gcc13.3.0-example4-34800367969-1` |

The 28 files are the 14 Stan programs `hpc/README.md` records for `example4`
with the `.stan` source beside each executable, which is the closest thing this
run gives to an independent check that the cache holds what it should.

Two things the run does not establish. The cache has not been shown to warm: a
second run against the same ref would show it, and the entry is in place for
whoever does that. And `brms` on the runner was 2.23.0, which is what
`hpc/image.lock` pins, so the two routes agreed on it here by coincidence of
timing rather than by construction --- the paragraph in the body about the R
package versions stands.

The run opened #345 with the rebuilt `example4`, six files and 34 changed lines.
That should be closed rather than merged. Rendered vignettes are brought up to
date together under #190, and this render was made in CI, which the body and
`hpc/README.md` both give as the route not to use for numbers that will be
quoted.

PR #341, opened 2026-09-14
Restore the caller's random number stream in check_fit() and dispersion() (#337)

### LINKED ISSUE #337: check_fit() and dispersion() reset the caller's random number stream, and check_fit() accepts seed = NULL
## What

`check_fit()` and `dispersion()` call `set.seed()` and never put the caller's
random number stream back. `check_fit()` also does not validate its `seed`, so
`check_fit(x, seed = NULL)` reaches `set.seed(NULL)`, which is the defect #310
fixed in the initial-value search, in a second place.

## Why it matters

A diagnostic is a summary computed from a fit. Running one should not change
what the next random operation in the session returns, and today it does:

```r
set.seed(1); check_fit(fit); rnorm(1)   # not the same as
set.seed(1); rnorm(1)
```

The stream is left at whatever `set.seed(10)` --- the default --- reaches, so a
simulation that runs a diagnostic partway through silently continues from a
different place. `weighted_draw_index()` states the rule this breaks:
"model averaging must not silently reset a user's simulation seed"
(`R/helpers.R:99`), and it saves and restores for exactly that reason. #310's
own reasoning draws the same line: the initial-value search advances the stream
because it is part of fitting, and a summary computed from a fit is on the
other side of it.

`check_fit(x, seed = NULL)` is the narrower of the two. `set.seed(NULL)`
re-initialises the stream from the clock and the process id, so the diagnostic
is not reproducible and the caller's seed is discarded. `dispersion()` cannot
reach this: it has `chk_number(seed)` at `R/dispersion.R:100`. `check_fit()`
has no equivalent.

## Evidence

R 4.6.1, `bayesnec` at `dev` plus the #310 branch. Read from source rather than
measured; the call sites are `R/check_fit.R:169`, `R/check_fit.R:408` and
`R/dispersion.R:127`, none of which is wrapped in a save and restore, against
`R/helpers.R:128-147`, which is. `set.seed(NULL)` changing `.Random.seed` was
measured for #310.

Found while reviewing #336.

<details>

Two separate changes. The first is mechanical; the second needs a decision.

Add `chk_number(seed)` to the `check_fit()` methods, matching
`dispersion()`. One line each, and it closes the `set.seed(NULL)` route.

For the restore, save `.Random.seed` and `RNGkind()` on entry and restore on
exit, as `weighted_draw_index()` does at `R/helpers.R:128-147` --- kind first,
then the seed, because the generator is encoded in `.Random.seed[1]`, and
removing `.Random.seed` rather than assigning it where the session had not yet
used the RNG. That helper is the pattern to copy; whether to factor it out into
something both call, or repeat it, is open.

This is a behaviour change: code that today relies on a diagnostic having
advanced the stream would see different draws afterwards. No test does, and no
vignette calls `check_fit()` or `dispersion()` before a random operation, but
it is a change and belongs in NEWS.

Both `summary()` paths reach `dispersion()`, so the restore has to be correct
for a fit inside a model set as well as a single fit.

</details>

beckyfisher:
## What

`check_fit()` and `dispersion()` now put the caller's random number stream back
before they return, and `check_fit()` refuses a `seed` that is not a single
number.

## Why it matters

A diagnostic is a summary computed from a fit. Running one changed what the
next random operation in the session returned, because both functions called
`set.seed(seed)` and left the stream wherever that reached. A simulation that
printed a diagnostic partway through silently continued from a different place.
#310 drew the line this restores: the initial-value search advances the stream
because it is part of fitting, and a summary computed from a fit is on the
other side of it.

`check_fit(x, seed = NULL)` reached `set.seed(NULL)`, which re-initialises the
stream from the clock and the process id. The diagnostic then did not repeat
and the caller's seed was discarded without a word. That is #310's defect
reached through a second door; `dispersion()` already had `chk_number(seed)`.

Both functions still repeat under the same `seed`. Only the stream is restored.

## Evidence

Measured on R 4.6.1, one pair of runs per function: `set.seed(1)` then three
`rnorm()` draws, against the same three preceded by the diagnostic.

| arm | `check_fit()` | `dispersion()` |
|---|---|---|
| no diagnostic | -0.6264538, 0.1836433, -0.8356286 | -0.6264538, 0.1836433, -0.8356286 |
| `dev` (`cb045f93`) | -2.0177300, -0.7461739, -3.0760010 | 1.3750320, -0.7165580, 0.9482187 |
| this branch | -0.6264538, 0.1836433, -0.8356286 | -0.6264538, 0.1836433, -0.8356286 |

`check_fit()` was run on the packaged `manec_example`'s `nec4param` component;
`dispersion()` on a `nec4param` Poisson fit of 60 simulated observations, since
`dispersion()` returns an empty vector for a gaussian response. Both functions
returned the same result from two calls separated by an `rnorm(5)`, on `dev`
and here. `check_fit(nec4param, seed = NULL)` returned a table on `dev` and
stops here with ``` `seed` must be a number (non-missing numeric scalar) ```.

<details>

### The restore is factored out

`with_preserved_rng_state()` in `R/helpers.R`. The issue left open whether to
factor the block out or repeat it. It already existed twice ---
`weighted_draw_index()` and `bnec_model_lapply()` --- so repeating it at the
three diagnostic call sites would have made five copies of a block whose
correctness rests on two things that are easy to get wrong: the generator kind
has to be restored before the seed, because the kind is encoded in
`.Random.seed[1]`; and a session that has not yet used the RNG has no
`.Random.seed`, so the object is removed rather than assigned. Both existing
copies now call the helper, so the change removes a duplicate rather than
adding one.

The helper takes its argument lazily and evaluates it in the calling frame, so
assignments inside the braces reach the caller's variables:

```r
with_preserved_rng_state({
  set.seed(seed)
  yrep <- posterior_predict(fit, ndraws = ndraws)
  mu <- apply(posterior_epred(fit), 2, median)
})
```

The restore is registered with `on.exit()`, so it fires whether the wrapped
expression returns or errors.

### Call sites

- `check_fit_table()` (`R/check_fit.R`), reached by the `bayesnecfit` and
  `bayesmanecfit` methods.
- `check_fit_combined()` (`R/check_fit.R`), the hurdle path.
- `dispersion()` (`R/dispersion.R`). Both `summary()` paths reach it, so the
  restore holds for a fit inside a model set as well as for a single fit.

`posterior_epred()` is now called inside the wrapped block at both `check_fit()`
sites. It draws nothing from the stream, so the statistics are unchanged, and
the relative order of the calls that do draw from it is preserved.

### Seed validation

`chk_number(seed)` is called in each of the three exported `check_fit()`
methods rather than only in the `bayesnecfit` one. The `bayesmanecfit` method
loops over `pull_out()`, so validating only in the delegate would refuse the
seed after the first component had been extracted.

### Tests

`tests/testthat/test-helpers.R` asserts the helper itself: the stream is put
back, the generator kind is put back, the restore fires after an error, a
session with no `.Random.seed` is left without one, and the expression is
evaluated in the calling frame. `test-check_fit.R` and `test-dispersion.R`
assert the stream through the two functions, that each still repeats under its
own seed, and that `check_fit()` refuses `NULL`, a character, and a length-2
seed.

The existing `weighted_draw_index()` tests --- purity, stream restoration,
`sample.kind` pinning --- cover the refactor of that function.

### Behaviour change

Code relying on a diagnostic having advanced the stream would see different
draws afterwards. No test and no vignette calls either function before a random
operation. Recorded in NEWS.

</details>

## Also in the diff

`NEWS.md` records a second consequence found in review. `check_fit()` is not
only a function a user calls: `bnec()` runs it on both return paths through
`message_control_fit()`, and `summary()` runs it whenever `check_fit = TRUE`.
So the stream after a fit and after a summary is restored too. On the released
code the diagnostic's own `set.seed(10)` left the stream at a fixed point after
every `bnec()` call, whatever the fit had done; it now reflects the fitting
alone. Measured on `manec_example` under the protocol above: `summary()` and
`message_control_fit()` each returned three different numbers on `dev` and
return the reference three here.

`average_estimates()`, `compare_estimates()`, `compare_fitted()` and
`sample_priors()` draw from the stream with an unseeded `sample()` and were
left alone. They advance the stream rather than resetting it, which is the
ordinary behaviour of a function that uses the RNG and what their own
documentation already tells a user to handle with `set.seed()`. Opened as #343.

Closes #337

https://claude.ai/code/session_01B7q91E341PBW9CkzxVRTwe


beckyfisher:
`dev` was merged in to clear a conflict in `NEWS.md` and `tests/testthat/test-helpers.R`, both of which were additions at the same place rather than changes to the same text --- #319's NEWS entry against this one, and its two `sub_x_transformation()` tests against the `with_preserved_rng_state()` ones. Both sides were kept. The diff against `dev` is otherwise unchanged, and checks now run: a pull request GitHub cannot merge produces no run at all, which is why the branch showed none before.

beckyfisher:
## Review outcome

Three review rounds were run with an independent agent. Two comments were real problems and both are fixed on the branch. The rest were documentation and test refinements, also applied. One was declined, with reasoning, and recorded as its own issue.

### The NEWS entry understated the scope

The entry described `check_fit()` and `dispersion()` as called directly. `bnec()` reaches `check_fit()` internally through `message_control_fit()` (`R/bnec.R:830`, `:843`), and `summary()` through `control_fit_issues()` (`R/summary.R:221`), so on `dev` the stream also changes after every fit and after every summary taken with `check_fit = TRUE`. Measured on `manec_example`, `set.seed(1)` then three `rnorm()` draws, against the same three preceded by the call: on `dev` neither `bnec()` nor `summary(check_fit = TRUE)` returns the reference, and on the branch both do. A paragraph naming those two routes was added.

### The test's own cleanup restored the seed before the generator kind

A round-1 fix added an `RNGkind()` restore to `tests/testthat/test-helpers.R` as a second `on.exit(..., add = TRUE)`. `add = TRUE` defaults to `after = TRUE`, so the seed restore ran first and the `RNGkind()` call then re-initialised the generator from the clock: 624 of the 626 elements of `.Random.seed` differed from the value held before the test. Nothing failed, because every later test in that file seeds itself, but a test added after it would not have been reproducible. This is the ordering trap `with_preserved_rng_state()` documents, reproduced in the cleanup of the test whose subject is that trap. Both restores are now a single handler, kind first.

### Refinements applied

- `expect_error(check_fit(nec4param, seed = NA), "seed")` was also satisfied by `set.seed()`'s own refusal, so it did not assert that the argument check is what refuses. It now matches `"must be a number"`.
- The repeatability test for `check_fit()` compared only `$sim_sd`; it compares the whole table.
- Tests were added for `seed = NA`, and for the generator kind being restored in the branch where the session held no `.Random.seed`.
- `dispersion()`'s `seed` documentation now states the `NULL` refusal.
- `R/parallel_models.R` attributed the helper to `check_fit()`; it now names `check_fit_table()` and `check_fit_combined()`.
- The helper's roxygen now records that a `return()` inside `expr` returns from the caller, and that `sys.call()` sees one extra frame.
- NEWS called the diagnostic's `set.seed(10)` trailing; it is the first statement of `check_fit_table()`.

### Declined

`average_estimates()`, `compare_estimates()`, `compare_fitted()` and `sample_priors()` also call `sample()` without restoring the stream. They advance the stream rather than resetting it to a fixed point, their documentation already directs the user to `set.seed()`, and the thinning `sample()` in `average_estimates()` decides which draws are averaged, so a fix changes the reported estimates and needs its own decision about a `seed` argument. Measured: the same `average_estimates()` call under `set.seed(1)` and `set.seed(2)` returns 1.15730258, 0.77072572, 1.30099908 against 1.17447245, 0.75880967, 1.29249890. Recorded as #343.

A fourth copy of the save-and-restore block remains in `tests/testthat/test-inits_functions.R`. The observation is accurate and is left for a test-side pass.

PR #346, opened 2026-09-14
Rewrite example5 as installation and setup instructions (342)

### LINKED ISSUE #342: Rewrite example5 as installation and setup instructions, from the training course module
## What

Rewrite `vignettes/example5.Rmd.orig` as installation and setup instructions for
`bayesnec`, taking the technical content from the "Software setup" module of the
training course (`vignettes/0Software-setup.qmd` in
[open-AIMS/cr_modelling_training](https://github.com/open-AIMS/cr_modelling_training))
and putting it in the register of the other vignettes. Every chunk becomes
`eval = FALSE`.

## Why it matters

The vignette publishes a contributor's home directory. `vignettes/example5.Rmd`
names `/Users/dbarneche/.cmdstan/cmdstan-2.33.1` 21 times in 520 lines, because
seven of its chunks are evaluated when the vignette is precompiled and print the
paths of whoever built it. The last of them embeds 77 lines of compiler
diagnostics from that machine. A published vignette should record nothing about
the person who rendered it.

The instructions are also obsolete, and they contradict themselves at the step
that matters most. They direct the reader to clone CmdStan from GitHub, run
`mingw32-make build`, reboot, and hand-write a `make/local` file to supply two
compiler flags. `cmdstanr::install_cmdstan()` requires none of that. The rendered
output then shows

```
## Warning: Path not set. Can't find directory: C:/cmdstan
```

directly beneath the instruction telling the reader to set that path.

The course module is the maintained version of the same material. It was written
in September 2026 against R 4.5.x and Rtools45, verified end to end on Debian
with R 4.6.1 and CmdStan 2.39.0, and it records the failures seen on fresh
Windows machines this year: the download timeout, the directory that
`install_cmdstan()` will not create, OneDrive, and the `tbb.dll` failure after an
Rtools upgrade. None of those appear in `example5`.

## Evidence

In `vignettes/example5.Rmd`, the rendered file that ships: line 237 gives the
contributor's CmdStan path, lines 248 and 385 give the warning above, and the
final chunk's output runs to 77 lines of compiler diagnostics. The home
directory appears 21 times in a 520-line file.

In `vignettes/example5.Rmd.orig`, the seven evaluated chunks are `check cmdstan
path`, `set cmdstan path`, `toolchain check`, `fit model`, `mcmc`, `check model`
and `check brmsmodel`, at lines 146, 152, 166, 178, 186, 202 and 208.
`install_cmdstanr`, immediately above them, is already `eval = FALSE`; the
difference looks incidental rather than decided.

Lines 94 to 123 of the same file are the CmdStan build: `git clone`,
`mingw32-make clean-all`, `mingw32-make build`, a reboot, an addition to the
system `PATH`, and a `make/local` file written in Notepad with its extension
removed. The text names the version it expects to see built, CmdStan v2.23.0,
against the 2.39.0 that `hpc/image.lock` pins. The editor throughout is RStudio.

Raised from the review of #339. Before #313 this leg of the precompile workflow
failed at `library(cmdstanr)`, the runner having no CmdStan; with `cmdstanr`
installed it renders, and the render substitutes the runner's path for the
contributor's. That is one arbitrary machine for another, not a fix.

<details>

### Content taken from the course module

- The list of software required, and why a compiler is needed at all: each model
  is translated into C++, compiled, and run, which is also why the first fit of a
  session takes longer than the ones after it.
- Windows as the route described in full, Linux as the second tested route, and
  macOS as an outline pointing at the upstream instructions.
- Installing R first and then the matching Rtools, with the warning that a
  mismatch is the commonest single cause of failure and that the error it
  produces names no version.
- `install.packages("cmdstanr", repos = c("https://stan-dev.r-universe.dev",
  getOption("repos")))`.
- The three things that stop a fresh CmdStan installation on Windows, all of
  which the course module measured: `dir.create("C:/cmdstan")` first, because
  `install_cmdstan()` does not create the directory; `options(timeout = 1800)`,
  because R's default of 60 seconds does not cover a 49 MB download and the
  `timeout` argument of `install_cmdstan()` does not govern it; and a path free
  of OneDrive, spaces and non-English characters, which `C:/cmdstan` gives.
- The `options(cmdstanr_cmdstan_path = ...)` line for `.Rprofile`, needed because
  that path is not the default.
- The four recurring Windows problems: a failed install that needs
  `overwrite = TRUE`; "No chains finished successfully" after an Rtools upgrade,
  which is `tbb.dll` built against the old toolchain and needs CmdStan rebuilt;
  the `cmdstanr` version-to-Rtools mapping, which changed at 0.9.0; and Windows
  Defender making compilation slow.
- WSL named as the fallback for one situation only, compilation blocked by
  security policy, and what it asks of the reader.

### Material to omit

Everything specific to the workshop: the completion deadline, sending output to
organisers, the pinned course version of `bayesnec`, and Positron, Git and `drc`,
which the course needs and a reader of this package does not. The Quarto callout
blocks go as well --- this is an `html_vignette` built by `knitr`, which does not
render them.

Say once, near the top, that `brms` supports either back end, that `rstan` is
installed as a `brms` dependency whether or not it is used as the sampler, and
that `cmdstanr` is neither on CRAN nor a dependency of `bayesnec` but is what
this package's own vignettes are built with. `options(brms.backend = "cmdstanr")`
is the line that selects it.

### Register

Section 12 of the working-directory `CLAUDE.md` and the style section of
`notes/vignette_numbering.md` apply. Headings are noun phrases: *Software
requirements*, *Windows*, *The CmdStan installation*, *Recurring problems on
Windows*, *Verification*. Second person in the body is right for instructions and
is what the current vignette already uses; it is wrong in a heading.

### Mechanics

`eval = FALSE` on every chunk. The vignette then needs no Stan toolchain to
precompile, renders identically wherever it is built, and stops being the one
leg of the precompile workflow whose output depends on the machine. Where the
reader needs to see what a working installation reports, write the expected text
in a plain fenced block rather than evaluating a chunk to produce it --- for
example `The C++ toolchain required for CmdStan is setup properly!`.

A verification section at the end should give the reader something to run that
exercises the compiler: `cmdstanr::check_cmdstan_toolchain()`, then a short
`bnec()` fit on `nec_data` with `chains = 2`, and a description of what a
successful result looks like. The course's `vignettes/check_setup.R` is the model
for what to check, not a file to copy.

The title and `%\VignetteIndexEntry{}` are currently "Running `bayesnec`", which
does not say what the vignette is. Name it for what it becomes. Keep the file at
`example5`. Rule 3 of `notes/vignette_numbering.md` records why: a rendered
vignette's number reaches the `precompile.R` output, the `vignette-fig-` figure
names, every cross-reference and the published URL, so renumbering means
changing all of them.

The register row for 5 in `notes/vignette_numbering.md` reads *NEC*, NSEC and
ECx, which is not what the file contains. Correct it in the same pull request.

`vignettes/example5.Rmd` has to be rebuilt for any of this to reach a reader, and
with no chunk evaluated that rebuild takes seconds.

### Approaches rejected

Do not replace the machine-specific output with a fixed fake path. A guide that
shows output the reader cannot reproduce is the defect, not the cure.

Do not copy the course module across and edit it down. It addresses a workshop
participant with a deadline, and the parts of it that survive are the technical
ones; starting from the existing vignette's structure and bringing the content in
gives a vignette rather than a workshop handout with the logistics deleted.

</details>

beckyfisher:
Closes #342.

## What

`vignettes/example5.Rmd.orig` is rewritten as installation and setup
instructions, titled *Installation and setup*, and no chunk in it is evaluated.
The technical content comes from the software setup module of
[open-AIMS/cr_modelling_training](https://github.com/open-AIMS/cr_modelling_training)
(`vignettes/0Software-setup.qmd` at `9d05595`), put in the register of the other
vignettes. `vignettes/example5.Rmd` is rebuilt, row 5 of
`notes/vignette_numbering.md` is corrected, and `vignette("example1")`,
`vignette("example2")` and `README.md` now point at the new vignette, in place
of the separate installation text each of them held.

## Why it matters

The rendered vignette published a contributor's home directory. Seven chunks
were evaluated at precompile, so the file recorded whichever machine built it:
`/Users/dbarneche/.cmdstan/cmdstan-2.33.1` appeared 21 times in 520 lines, and
the final chunk's output ran to 77 lines of compiler diagnostics from that
machine. The rendered output also printed

```
## Warning: Path not set. Can't find directory: C:/cmdstan
```

directly beneath the instruction telling the reader to set that path.

The instructions were also obsolete. They directed the reader to clone CmdStan
from GitHub, run `mingw32-make build`, reboot, add two directories to the system
`PATH`, and hand-write a `make/local` file in Notepad with its extension
removed. `cmdstanr::install_cmdstan()` requires none of that. The text named
CmdStan v2.23.0 as the version it expected to see built, against the 2.39.0 that
`hpc/image.lock` pins.

## Evidence

The rendered vignette is 357 lines against 520, and evaluates nothing:

```
$ grep -c "dbarneche" vignettes/example5.Rmd        # was 21
0
$ grep -c '```{r' vignettes/example5.Rmd            # no executable chunks
0
$ grep -c "^## " vignettes/example5.Rmd             # was 162
4
```

The third of those is the discriminating count. `^#> Error`, which
`vignettes/precompile.R` greps for, returns 0 on the old file as well: the old
example5 had no setup chunk and so used knitr's default `## ` prefix rather than
`#> `, which means that guard could never have caught an error in it. The four
remaining `## ` lines are markdown headings. Ten occurrences of `C:/cmdstan`
remain and are intentional; what is gone is every path specific to a machine.

The rebuild ran in seconds with no Stan toolchain, because every chunk is
`eval = FALSE`. The rendered `.Rmd` was then taken through
`rmarkdown::render()` under R 4.6.1 and pandoc: it renders, the pipe table
renders as a table, the three citations resolve against `vignettes/bayesnec.bib`,
and the cross-references resolve.

Every `cmdstanr` name, argument and message quoted in the vignette was checked
against the installed `cmdstanr` 0.9.0 under R 4.6.1, and the `Provenance`
section of the vignette records which claims were read from there rather than
measured for the course.

## Judgement calls

Seven decisions changed the content relative to what #342 or the course module
specified, or are worth disagreeing with.

### Persistence of the CmdStan path

The course module persists a non-default CmdStan installation with
`options(cmdstanr_cmdstan_path = "C:/cmdstan/cmdstan-2.39.0")` in `.Rprofile`.
No version of `cmdstanr` reads that option. In `cmdstanr` 0.9.0 the string
`cmdstanr_cmdstan_path` appears nowhere in the namespace, the Rd database or the
lazy-load database, and `cmdstanr:::cmdstanr_initialize()` takes the path from
`Sys.getenv("CMDSTAN")` where that is set and from `~/.cmdstan` otherwise. A
reader following the module exactly would still meet "CmdStan not found" in every
new session, which is the failure the instruction says it prevents.

Both halves were then measured rather than only read, under R 4.6.1 and
`cmdstanr` 0.9.0, with `HOME` pointed at an empty directory so that no
`~/.cmdstan` could resolve. The target was a purpose-built directory holding two
stub installs, `cmdstan-2.38.0` and `cmdstan-2.39.0`, and being neither itself,
so that the resolution had a newer and an older candidate to choose between:

| setting | `cmdstan_path()` |
|---|---|
| `options(cmdstanr_cmdstan_path = ".../cmdstan-2.39.0")` | errors: *CmdStan path has not been set yet* |
| `CMDSTAN=<the parent directory>` | `.../cmdstan-2.39.0` |

So the vignette uses `CMDSTAN=C:/cmdstan` in `.Renviron`, says R has to be
restarted because `.Renviron` is read at startup, and drops the version number,
which means the line survives the next CmdStan upgrade unedited. It also records
the asymmetry that follows: `set_cmdstan_path()` has no such resolution, so the
by-hand fallback does need the version directory, and given the parent it warns
`Can't find CmdStan makefile to detect version number` and leaves
`cmdstan_version()` at `NULL`.

The course module has the same defect, at
`vignettes/0Software-setup.qmd:169-178` of `open-AIMS/cr_modelling_training`,
and this PR does not fix it there. Workshop participants will meet it.

### The tested R version

The course module specifies R 4.5.x with Rtools45, which is right for a workshop
taught on a fixed date. A vignette that names that pairing goes stale at the
next R minor release, and `DESCRIPTION` requires only R 4.1. The vignette gives
the rule instead --- Rtools is versioned against the minor version of R, and
CRAN's Rtools page states which goes with which --- and then makes the operative
version the one `check_cmdstan_toolchain()` names, quoting the message verbatim.
That also resolves a divergence the general rule cannot: `cmdstanr 0.9.0`'s
`rtools4x_version()` returns `"45"` for every R from 4.5 onward, so on R 4.6.1 it
asks for Rtools45 whatever CRAN pairs with that R.

### The macOS claim about the authors

The course module says the organisers have no Mac available. The evidence in
#342 is a path under `/Users/dbarneche/`, so the same sentence would have been
false of `bayesnec`. The vignette says only that these steps have not been
verified on macOS.

### The placement of `eval = FALSE`

#342 specifies "`eval = FALSE` on every chunk". It is set once in the `setup`
chunk via `opts_chunk$set()` rather than repeated on each chunk header, so that
a chunk added later inherits it instead of having to remember it. The rendered
file confirms the effect: zero executable chunks.

### The unused `[e5]` reference in example6

`vignettes/example6.Rmd.orig:37` and `vignettes/example6.Rmd` define a link
reference `[e5]` that neither file uses. Its URL is unchanged by this PR, and
removing it would put example6 in the diff for nothing.

### The Windows Subsystem for Linux, dropped

#342 asked for WSL to be named as the fallback for one situation, compilation
blocked by security policy, and the first draft did that in a section of its
own. RF asked for it out, and it is out: the sentence in *Route by operating
system* and the section both. *Compilation blocked by policy* would otherwise
have pointed at a section that no longer exists, so it now says that resolving
it needs whoever administers the machine, which is what is true without WSL.

One WSL mention is kept, in *Provenance*: the Linux check ran under WSL2 Debian
on an already-configured machine, and that is a limit on the measurement rather
than a route the vignette offers. Removing it would overstate the check.

### The rendered files of example1, example2 and README

`vignettes/example1.Rmd`, `vignettes/example2.Rmd` and `README.md` are
generated, and none is regenerated here. The cross-reference added to each is
prose, so the identical line was applied to the source and to the generated
file, which keeps the two in step without a refit --- `example1` and `example2`
are hours of sampling each. `example5` itself was re-precompiled rather than
hand-applied, which takes seconds because nothing in it is evaluated. This
follows the repo's practice for prose-only vignette changes rather than
introducing one, but hand-editing a generated file is the item here most worth
disagreeing with, so it is above the fold.

<details>

### Structure

Headings are noun phrases, per section 12 of the working-directory `CLAUDE.md`
and the style section of `notes/vignette_numbering.md`: *Software requirements*,
*Route by operating system*, *Windows*, *R and Rtools*, *R packages*, *The
CmdStan installation*, *Recurring problems on Windows*, *macOS*, *Linux*,
*Verification*, *Provenance*. Second person stays in the body, which is right for instructions and is what the previous vignette
used.

The course module heads each of its five recurring Windows problems with a bold
sentence. Bold lead-ins are not used in prose (section 12, rule 3), so each
became a `###` noun-phrase heading: *A partly built CmdStan directory*, *Chains
that do not start after an Rtools upgrade*, *The `cmdstanr` version and the
Rtools it expects*, *Slow compilation*, *Compilation blocked by policy*. That
also makes each findable by the reader who has just hit it.

### Output the reader cannot reproduce

Where the reader needs to see what a working installation reports, the expected
text is a plain fenced block rather than a chunk evaluated to produce it --- `The
C++ toolchain required for CmdStan is setup properly!`, and the Rtools and
directory-assertion messages, all quoted verbatim from `cmdstanr` 0.9.0. No
fabricated `summary()` output is shown: #342 rules out output the reader cannot
reproduce, so the verification section describes what to look for instead,
including the *Further Distributional Parameters* block that a `Beta` fit prints.

### Verification

`cmdstanr::check_cmdstan_toolchain()` and `cmdstanr::cmdstan_version()` first,
then a two-chain `bnec()` fit of **nec4param** on `nec_data` with
`family = Beta(link = "identity")`. The call is the idiom
`vignettes/example3.Rmd.orig` already uses on the same data. The link is passed
explicitly because `validate_family()` forces `link = "identity"` on every
family `bnec()` accepts, and `Beta()` on its own defaults to logit: writing it
out makes the call say what is fitted rather than leaving the reader to find
that out.

`vignettes/check_setup.R` from the course was the model for what to check, not a
file to copy: it ends in READY or NOT READY and asks the participant to return
its output, neither of which belongs here.

### Material omitted

The workshop logistics, the pinned course version of `bayesnec`, and Positron,
Git and `drc`. The Quarto callout blocks go too --- this is an `html_vignette`
built by `knitr`, which does not render them.

</details>

https://claude.ai/code/session_015YqtCtVr51ijDg65ecQDqi


beckyfisher:
## Review cycle

Four rounds between a developer and an independent reviewer agent, ending in
agreement that the change is sound. The rounds are not reproduced here; what
follows is the findings that were real defects and what was done about each.
Twenty-six findings were raised and twenty-five stood on independent
verification. The one that did not is recorded at the end, because a reviewer's
finding being wrong is worth the same visibility as one being right.

### The blocking defect

The step the whole Windows route rests on did nothing. The vignette told the
reader to persist a non-default CmdStan installation with
`options(cmdstanr_cmdstan_path = "C:/cmdstan/cmdstan-2.39.0")` in `.Rprofile`,
and then told them what would go wrong if they omitted it. No version of
`cmdstanr` reads that option. That replaced one non-working instruction at that
step with another, which is the defect #342 exists to fix.

Measured under R 4.6.1 and `cmdstanr` 0.9.0, with `HOME` pointed at an empty
directory so no `~/.cmdstan` could resolve, against a purpose-built directory
holding two stub installs, `cmdstan-2.38.0` and `cmdstan-2.39.0`, and being
neither itself, so that the resolution had a newer and an older candidate to
choose between:

| setting | `cmdstan_path()` |
|---|---|
| `options(cmdstanr_cmdstan_path = ".../cmdstan-2.39.0")` | errors: *CmdStan path has not been set yet* |
| `CMDSTAN=<the parent directory>` | `.../cmdstan-2.39.0` |

It is now `CMDSTAN=C:/cmdstan` in `.Renviron`, with the restart explained and the
version number dropped, since the parent directory resolves to the newest install
on its own and so survives the next CmdStan upgrade unedited. A second round
found the two sentences adjoining that instruction were still wrong: one sent the
reader to a `.Rprofile` line that is now something else, and the by-hand fallback
`set_cmdstan_path()` does need the version directory the variable omits --- given
the parent it warns `Can't find CmdStan makefile to detect version number` and
leaves `cmdstan_version()` at `NULL`. Both are corrected and the asymmetry is
stated rather than left for the reader to hit.

The course module has the same defect, at
`vignettes/0Software-setup.qmd:169-178` of `open-AIMS/cr_modelling_training`, and
this PR does not fix it there. Workshop participants will meet it.

### Claims that were too strong

- *"the back end the package is developed and tested against"* --- nothing under
  `tests/` sets a backend, so `R CMD check` samples through `rstan`. Deleted.
- The Rtools pairing --- the vignette gave a general rule that does not
  describe what `cmdstanr` looks for on R 4.6.x, which is the R the vignette
  itself cites: `rtools4x_version()` returns `"45"` for every R from 4.5 onward.
  The operative version is now the one `check_cmdstan_toolchain()` names, quoted
  verbatim, and the version-to-Rtools mapping is stated as a property of the
  installed `cmdstanr`.
- The Linux verification --- the source records a WSL2 Debian run on an
  already-configured machine and says in terms that it does not verify the
  instructions. The vignette asserted the route was verified. Now states the run
  and both limits.
- Attribution --- the 49 MB archive, the 500 KB/s download, the 4.9 minutes at
  `timeout = 1800`, the OneDrive failure and the `tbb.dll` mechanism are the
  course's measurements, not this package's, and read as ours. A `Provenance`
  section now attributes them, and separately lists every `cmdstanr` claim read
  from source rather than measured.

### Smaller corrections

`bnec()` calls `brms::brm()`, and it is `brm()` that writes the Stan program, not
the package. The described `summary()` output omitted the *Further Distributional
Parameters* block that a `Beta` fit prints. *Route by operating system* was a
subsection of *Software requirements* though it is not one. The `.Rprofile`
back-end line was filed under Windows though it applies everywhere. The routing
paragraphs were four sentences whose subject was the vignette. Two ruled verbs in
the `NEWS` text, one in the setup chunk's comment. Both sentences of the toolchain
error are now quoted, and `check_cmdstan_toolchain()` is marked as available only
after the section that installs `cmdstanr`.

Two measurements in the PR body's Evidence did not check out, and the corrected
form is worth recording: `grep -c "^#> Error"` returning 0 is **vacuous** on the
old example5, which had no setup chunk and so used knitr's default `## ` prefix.
The discriminating count is `^## `, 162 before and 4 after. A side effect worth
knowing beyond this PR: `vignettes/precompile.R`'s errored-chunk guard greps
`^#> Error`, so it could never have caught a failed chunk in the old example5.

### Scope added during the cycle

The package had three divergent installation texts --- example5, `vignette("example1")`
and `README.md` --- and a fourth paragraph in `vignette("example2")`. None was
broken by this PR, but example5 is now the installation vignette, so each of the
other three points at it. The identical prose line was applied to the source and
to the generated file in each case, so no refit is needed; example5 itself was
re-precompiled, which takes seconds because nothing in it is evaluated.

### The finding declined

A later round read the measurement table above as quoting a `cmdstan-2.38.0`
install that does not exist on the machine, and called it a typo in the stated
conditions. It was not: the run used a directory built for the purpose with two
stub installs, not the real `~/.cmdstan`, and both rows re-ran from it. The
finding was withdrawn once the directory was listed.

It is recorded because the gap behind it was real. The body named no path, so a
run against a synthetic directory and a run against `~/.cmdstan` were
indistinguishable from the text, and the body now says which it was. A stated
measurement that does not say what it was measured against invites exactly this,
whoever reads it.

## Status

`R CMD check` and `pkgdown` are the only checks on this branch and it changes no
code under `R/`, so the rendered vignette building is what they exercise here.
The rendered `example5.Rmd` contains no knitr chunk at all, so `R CMD check`
evaluates none of its code. Locally it renders under R 4.6.1 and pandoc with the
table and all three citations resolving.

`origin/dev` is merged in as of `d5a36cbc`, so the diff is the ten intended
files, and the only `NEWS.md` difference against current `dev` is this entry.

PR #348, opened 2026-09-14
Make posterior comparisons and prior samples reproducible (343)

### LINKED ISSUE #343: average_estimates(), compare_estimates(), compare_fitted() and sample_priors() do not repeat under a seed
## What

`average_estimates()`, `compare_estimates()`, `compare_fitted()` and
`sample_priors()` draw from the caller's random number stream with an unseeded
`sample()`, so each call advances it and none of them repeats.

## Why it matters

This is the other half of the distinction #337 drew. #337 fixed the diagnostics
that *reset* the stream --- the damaging case, because the next draw then had
nothing to do with the caller's seed. These four *advance* it instead, which is
the ordinary behaviour of any function that uses the RNG, and their
documentation already says to run `set.seed()` before the call. So this is not
the same defect.

What it does mean is that a result none of these functions names a `seed` for
depends on where in a session it is computed, and two calls in one session do
not agree. `check_fit()` and `dispersion()` both take a `seed` and now restore;
these take neither.

## Evidence

R 4.6.1, `bayesnec` at `dev` plus the #337 branch. Read from source: the call
sites are `R/average_estimates.R:114`, `R/compare_estimates.R:101`,
`R/compare_fitted.R:87` and `R/sample_priors.R:75` and `:82`.

Measured on the two `manec_example` components pulled out, one pair of runs
each. `set.seed(99); rnorm(1); average_estimates(list(a = nec4param, b =
ecx4param), estimate = "nec"); rnorm(2)` does not return what `set.seed(99);
rnorm(3)` returns, so the stream is advanced. The same call preceded by
`set.seed(1)` and by `set.seed(2)` returns two different estimates, so the
thinning draw reaches the reported number and not only the stream.

Found while reviewing #341.

<details>

Two questions, and the second is the one that needs a decision.

Whether each gets a `seed` argument matching `check_fit()` and `dispersion()`,
defaulting to a fixed value, or keeps answering to the ambient stream. The two
differ in what a user has to do to reproduce a number: the first makes the
default reproducible and the second keeps the documented `set.seed()` habit.
`average_estimates()`, `compare_estimates()` and `compare_fitted()` are the
public trio and should be decided together.

`sample_priors()` is a different case. It draws from the priors to plot them,
and `check_priors()` calls it, so whatever is decided for the trio may not
suit it.

Whichever is chosen, `with_preserved_rng_state()` in `R/helpers.R` is the
restore to use where one is wanted; do not repeat the block. Note that the
thinning `sample()` in the first three is a subset of each component's
posterior, so a change here changes the estimates themselves, not only the
stream --- `?average_estimates` and NEWS both need it.

</details>

beckyfisher:
## What

Make posterior averages, comparisons and prior samples reproducible with a
`seed = 10` argument, while restoring the caller's saved random number state.
`compare_posterior()` forwards the seed for both estimate and fitted-curve
comparisons. Closes #343.

## Why it matters

Repeated calls with the same inputs and RNG kind now return the same result
without changing subsequent simulation draws under the default Inversion normal
generator. Box-Muller's cached normal value is outside the saved state and
cannot be restored; subsequent normal draws can change with that generator.
Select another sample with the
function's `seed` argument; a preceding `set.seed()` no longer changes the
result. Posterior pairing affects the reported estimates, difference intervals
and probabilities, so these may differ from earlier versions.

The fixed default also applies to `sample_priors()` so prior plots repeat.
In the current source, `check_priors()` uses `brms::hypothesis()` and does not
call `sample_priors()`.

## Evidence

On WSL Debian with R 4.6.1 and brms 2.23.0, 246 assertions passed against an
installed build with `NOT_CRAN=true`, with no failures, warnings or skips.
The reviewer independently verified the 28 added stream assertions.
Tests use the packaged `manec_example` fits and a bounded normal prior; no
new models were fitted. The full model-fitting test suite was not run locally.

`R CMD check --no-manual --no-build-vignettes --no-tests` completed without
errors. It reported two warnings for unbuilt vignettes and one existing note
about `modifyList()` in `make_good_inits()`. Documentation and examples passed.

<details>

- Reuse `with_preserved_rng_state()` around random pairing and prior sampling.
- Include the prediction loop in `compare_fitted()` because `ndraws` also
  selects posterior draws at random.
- Use `chk_number()` to validate seeds, matching the diagnostic functions.
  `NULL` is rejected. Fix `sample.kind = "Rejection"`, retaining the caller's
  other RNG kinds and restoring the saved state afterwards. The Box-Muller cache limitation is documented.
- Test repeated defaults, explicit alternate seeds, ambient seed independence,
  absent `.Random.seed`, non-default RNG kinds, errors, dispatch through grouped
  fits, prediction subsampling and model-averaged NEC, ECx and NSEC estimates.
- Update roxygen documentation, generated help files and NEWS, including the
  change from ambient `set.seed()` usage.

</details>


beckyfisher:
Independent review completed in two cycles. The reviewer and developer agree
that the PR is sound, with one documented limitation corrected.

The original text promised complete RNG-stream restoration. R's Box-Muller
normal generator caches a value outside `.Random.seed`, so subsequent normal
draws can change despite restoring the saved state. The public documentation, NEWS and
shared-helper documentation now state this exception and identify the default
Inversion normal generator for preserving subsequent draws.

The tests now compare subsequent normal and uniform draws under Inversion with
both Mersenne-Twister and L'Ecuyer-CMRG, including an error path. They also
characterise the Box-Muller cache limitation across all four affected APIs.

Validation against an installed build: 246 assertions passed, with no failures,
warnings or skips. The reviewer independently verified the 28 new assertions.

The PR description has been updated to include this qualification and the latest
validation results.

beckyfisher (commented):
This looks sound

Issue #311, opened 2026-09-10
The test suite passes rstan-only arguments to bnec()


beckyfisher:
## What

Twelve `bnec()` calls in `tests/testthat/` pass `open_progress = FALSE`. That is
an argument of `rstan::sampling()`. Under the `cmdstanr` backend `brms` raises
`unused argument (open_progress = ...)` and the fit fails.

## Why it matters

The tests pass today because `brms` defaults to `rstan`. They fail for anyone
who has `options(brms.backend = "cmdstanr")` set in their session or profile,
and the failure is reported as a model that would not fit rather than as an
argument that is not accepted, which is a slow thing to diagnose.

The same argument caused a real failure in the vignettes: precompiling with the
`cmdstanr` backend made every model in example1 to example4 fail, and `bnec()`
reported only `None of the models fit successfully`. It was removed from the
vignettes on #306; the tests were left alone as out of scope for that change.

## Evidence

```
tests/testthat/test-bnec_record.R:57
tests/testthat/test-bnec_hurdle.R:175
tests/testthat/test-check_fit.R:177
tests/testthat/test-get_priors.R:166, 185, 204
tests/testthat/test-failed_models.R:163
tests/testthat/test-dispersion.R:19, 58, 225
tests/testthat/test-inits_functions.R:488
tests/testthat/test-zero-inflated-counts.R:124
```

Reproduced outside the suite on 2026-09-10, R 4.6.1, brms 2.23.0, cmdstanr
0.9.0:

```r
options(brms.backend = "cmdstanr")
bnec(y ~ crf(x, model = "nec4param"), data = dat, open_progress = FALSE)
#> unused argument (open_progress = .x5)
```

Without the argument the same call fits.

<details>
<summary>Implementation detail</summary>

Remove the argument rather than replace it. It defaults to `interactive()`, so
in a test run it is already `FALSE`, and under either backend deleting it
changes nothing about what is fitted.

Worth deciding at the same time whether the suite should be run under both back
ends, or whether the package should refuse an argument the chosen backend cannot
accept rather than passing it through to `brm()`. The second is the larger
question and belongs on its own issue if it is wanted.

</details>

PR #349, opened 2026-09-14
Make fitted tests backend-neutral (#311, #328)

beckyfisher:
Addresses #311 and #328.

## What

Twelve fitted tests no longer pass `open_progress = FALSE` through `bnec()` or
`bnec_hurdle()`. The argument is specific to the `rstan` backend. Omitting it
lets the same tests run when `brms` is configured to use `cmdstanr`.

The branch is based on `dev` after #331 merged the #328 fixture reuse. That
change builds five Stan programs where the affected files previously built
fourteen. This pull request completes the combined test-infrastructure batch
with #311. It also corrects stale descriptions of #328's withdrawn worker-count
policy and prevents a manually dispatched timing run from truncating a
fractional worker count.

## Why it matters

The test suite previously failed when a contributor had
`options(brms.backend = "cmdstanr")` in the session or profile. The failure was
reported as a model that did not fit, although the model and data were valid.
The retained fixture reuse removes nine repeated compilations without changing
an assertion.

## Evidence

On R 4.6.1 with `brms` 2.23.0, `cmdstanr` 0.9.0 and CmdStan 2.39.0, a revised
`nec4param` test fit compiled and sampled through the `cmdstanr` backend. Both
chains completed and `bnec()` returned a `bayesnecfit`. All eight edited test
files parse, no `open_progress` call remains under `tests/`, and
`R CMD build --no-build-vignettes` succeeds.

The workflow parses as YAML. Direct tests of its worker-input rule accept
positive integral counts and reject fractional, non-finite, out-of-range and
non-positive values.

The fixture reuse merged in #331 was measured directly: the affected files
compiled five Stan programs rather than fourteen with assertion counts unchanged
at 71 and 36. Its four-platform R CMD check matrix passed. This pull request's
matrix tests the combined state.

<details>

`open_progress` defaults to `interactive()` for the `rstan` sampling path, which
is already false in non-interactive test runs. The `cmdstanr` path does not
accept the argument. Deleting it therefore changes no model equation, family,
prior, seed, chain count, iteration count or assertion.

#328's addendum withdrew the proposed CI worker-count change after paired runs
could not distinguish its effect from runner variance. This branch does not
restore that setting. It retains only #331's deterministic fixture reuse and
the workflow instrumentation that already merged to `dev`. A blank dispatch
input leaves `TESTTHAT_CPUS` unset, and a supplied value must represent a
positive integer before conversion.

No scientific judgement was required for either part.

</details>


beckyfisher:
The pull request body now includes the corrected #328 worker-count description and validation of manually supplied worker counts. The latter rejects fractional, non-finite, out-of-range and non-positive values before integer conversion.

beckyfisher:
Independent review completed after three cycles. No defect was found in the
test changes or their scientific behaviour.

The review identified and corrected these maintenance problems:

- The workflow said that a blank dispatch input used a runner-computed worker
  count, although it leaves `TESTTHAT_CPUS` unset. The description now states
  that testthat uses its default of two workers.
- Manual dispatch silently truncated a fractional worker count. Validation now
  rejects fractional, non-finite, out-of-range and non-positive values before
  integer conversion.
- The work queue retained #328's withdrawn worker-count plan and stale pull
  request status. It now records #331's merged fixture reuse, #349's scope, the
  withdrawn policy, and the current open pull requests.

The reviewer accepted the corrections and found no remaining actionable issue
at `99ad0115`.

PR #350, opened 2026-09-15
Declare the predictor prior scale

### LINKED ISSUE #317: A logged predictor with no negative values is logged a second time
## What

The `nec` and `ec50` prior decides whether the user has already logged the
predictor by testing whether any predictor value is negative. A user who
supplies `log(concentration)` for a series whose lowest tested concentration is
at or above 1 produces no negative values, is classified as having supplied
concentrations, and receives a prior built from the log of a log.

The test is `min(x) < 0` on `master` and on `dev`, so this is a defect in the
released package and not something the `dev` stack introduced.

## Why it matters

Which prior a user receives depends on the units the concentrations were
recorded in. The same experiment in mg/L and in ng/L is classified differently,
because dividing by a constant moves the logged values below zero. Nothing
warns.

A second, smaller consequence: a dose of exactly 1 gives `log(1) = 0`, which is
neither negative nor retained by the `u[u > 0]` filter, so that dose contributes
nothing to the prior at all.

```
distinct supplied values:            0  1.0986  2.3026  3.4012  4.6052
values used by the prior:               1.0986  2.3026  3.4012  4.6052
```

## Evidence

R 4.6.1, brms 2.23.0. `master` at 6f6bd25d (2.1.3.1, the CRAN release) and
`dev` at fc33ca84 with PR #315 applied. Priors only; nothing was fitted.

A 10 to 10000 series over seven doses, six replicates, supplied as
`log(concentration)`, against the same experiment recorded in units a
millionfold smaller:

| | supplied in mg/L | supplied in ng/L |
|---|---|---|
| `master`, released | `gamma(5, 0.347436)` | `normal(-8.05905, 23.30496)` |
| `dev`, `"uninformative"` | `lognormal(1.75032, 0.467504)` | `normal(-8.05905, 24.87075)` |
| `dev`, `"regularizing"` | `lognormal(1.75032, 0.393875)` | `normal(-8.05905, 1.48468)` |

The right-hand column is what the entry is meant to be. The left-hand column is
the misclassification, and the discriminator is the same in all three rows.

**How much it changes the prior.** A true threshold was placed at every interior
tested dose over ten designs --- five concentration ranges by five and seven
doses, all supplied logged with every logged value at or above zero --- and
scored by the truncated prior CDF at the true value. Extreme doses are excluded
because they sit at the truncation bounds by construction.

| route | prior set | range of the truncated CDF | outside the central 95% |
|---|---|---|---|
| classified correctly | `"uninformative"` | 0.166 to 0.834 | 0 of 40 |
| classified correctly | `"regularizing"` | 0.052 to 0.949 | 0 of 40 |
| misclassified | `"uninformative"` | 0.029 to 0.916 | 0 of 40 |
| misclassified | `"regularizing"` | 0.011 to 0.913 | **4 of 40** |

The misclassified prior is still peaked inside the tested range and is still
usable. It is skewed toward the lower doses, and under `"regularizing"` it
excludes a true threshold near the bottom of the series in four of forty cells,
which is the case #302 exists to protect.

<details>

### Where the decision is made

`predictor_prior()`, `R/define_prior.R`:

```r
spans_negative <- min(u) < 0
z <- if (spans_negative) u else log(u[u > 0])
```

On `master` the same decision is made in `set_distribution()`,
`R/set_distribution.R`, which returns `"gaussian"` for `min(x) < 0`, `"Beta"`
where the maximum is at or below 1, and `"Gamma"` otherwise. `define_prior()`
then maps those to `normal(median(x), 10 sd(x))`, `beta(2, 2)` and
`gamma(5, 1/(median(x)/2))`.

A narrow third case exists on `master` only: a logged series whose values all
fall within [0, 1], which is concentrations between 1 and about 2.72, is
classified as `"Beta"` and receives the fixed `beta(2, 2)`.

### What is recoverable and what is not

The model frame keeps the expression the user wrote:

```
crf(log(x), ...)  ->  columns  y | log(x)
crf(logx,   ...)  ->  columns  y | logx
```

So a predictor logged inside the formula is detectable by name. A predictor the
user logged in their own data frame before calling `bnec()` is not
distinguishable from concentrations by any property of the data. Any fix is
therefore partial unless the scale is declared.

### Options, not a recommendation

1. **Detect the formula expression.** Treat a predictor whose model frame column
   is `log(...)` as already logged, regardless of sign. Fixes the common case,
   leaves `crf(logx, ...)` wrong, and introduces a rule keyed on syntax rather
   than on data.
2. **An explicit argument**, for example `predictor_scale = c("auto",
   "concentration", "log")`, defaulting to the current behaviour. Complete and
   explicit, at the cost of an argument users must know about.
3. **Warn rather than change.** Where the predictor is non-negative and its
   values are evenly spaced on the recorded scale while spanning more than about
   two orders of magnitude once exponentiated, say what has been assumed. Cheap,
   changes no default, and still leaves the wrong prior in place.

Option 1 or 2 changes the `"uninformative"` entry, which is the CRAN default, so
it is a change to published behaviour and needs its own audit against the 30
cells of #302 and #305 rather than only the measurements above.

### Why this was not fixed in #315

#314 scoped changing an `"uninformative"` default out, and #315 changes only the
`"regularizing"` spread. The interaction is recorded in `predictor_prior()`'s
roxygen and in that pull request's body: the four failing cells above are the
misclassification combined with the narrowing, and the narrowing is #305's, the
lognormal regularizing entry on that branch being bit-for-bit what #305
produced.

### Definition of done

A decision is recorded on which of the three options is taken, or that none is;
if a default changes, the 30-cell audit of #302 and #305 is rerun and reported;
and the dose at exactly 1 is either retained or its exclusion is stated as
deliberate.

</details>

beckyfisher:
## What

Adds an explicit `predictor_scale` declaration to the default-prior routes. A user who supplies an already logged predictor can now select `predictor_scale = "log"`, including when every value is non-negative. `"concentration"` makes the recorded-concentration interpretation explicit, and `"auto"` retains the existing sign-based choice.

Closes #317.

## Why it matters

The `nec` and `ec50` prior no longer has to depend on the units used to record an experiment. On the declared log scale, `log(1) = 0` remains a tested predictor value rather than being discarded as a zero control.

## Scientific decision

No property of a numeric column can distinguish precomputed logged values from recorded concentrations. The new declaration is therefore explicit. The default remains `"auto"` so existing calls and published analyses retain their priors.

`"log"` keeps every distinct supplied value and builds the existing normal prior on that scale. `"concentration"` omits zero controls, logs the positive concentrations, and builds the existing lognormal prior. It refuses negative values before any model or grouped level is fitted. The historical uninformative widths remain unchanged: `10 * sd(x)` on an already logged predictor and the 95% coverage width on recorded concentrations. The `"regularizing"` set retains its 98% coverage rule.

## Validation

- `test-define_prior.R`, including the 30-cell #302/#314 regression gates
- `test-fit_bayesnec.R`, with `brm()` mocked to inspect the fitted prior
- `test-bnec.R` and `test-bnec_group.R`, including pre-loop refusal
- `test-bnec_hurdle.R` and `test-bayesnechurdlefit-methods.R`, including
  zero-component-call refusal for factorised fits and amendments
- installed-package call to `get_priors(..., predictor_scale = "log")`
- `R CMD check --no-manual --no-build-vignettes --no-tests`: two existing vignette warnings and two existing code/package-structure notes; no new finding

<details>

The declaration is passed through `bnec()`, `get_priors()`, `amend()`, `bnec_group()`, factorised hurdle fits and joint hurdle/zero-inflated prior construction. `validate_predictor_scale()` centralises the accepted values and checks a declared concentration scale before model or group loops. `predictor_prior()` repeats the validation as a backstop for internal routes.

The internal vector previously named `predictor_scale` was renamed `prior_predictor`; it represents the whole predictor used to prime and bound both blocks of a joint fit, not the user's scale declaration.

</details>


beckyfisher:
The PR body was updated to include the factorised hurdle preflight coverage and its zero-component-call regression tests.

beckyfisher:
Independent review completed after two cycles.

One real problem was found in the original PR: factorised hurdle fits and amendments validated the growth and survival components sequentially. If negative predictor values occurred only among deaths, the growth component could begin sampling before the survival component rejected `predictor_scale = "concentration"`.

This is fixed by validating the complete predictor before either component call and explicitly forwarding the validated declaration to both components. Regression tests now verify that invalid full predictors cause zero component calls. The reviewer independently reproduced the corrected behaviour and found no remaining defects on re-review.

Issue #344, opened 2026-09-14
ecxhormebc5 fails to initialise on a negative predictor under a positive family, rather than being excluded with a reason


beckyfisher:
## What

`ecxhormebc5` fails to initialise on a strictly positive family when the
predictor is negative over part of its range, and is reported as a failed fit
rather than excluded with a reason. It fits normally where the predictor is
non-negative.

## Why it matters

The user sees `Error : Failed to fit model ecxhormebc5.` once per call, which
names neither the cause nor anything they could change. `check_models()` already
declines `ecxsigm` on a negative predictor with a reason, so the machinery for
saying so exists.

The estimates do not change: a model that fails contributes no weight, so the
model-averaged result is the same either way. What is gained is a reason instead
of an error, and the compile and initialisation attempts not spent.

## Evidence

`example8` precompiled on the AIMS HPC, jobs 902840 and 906481, `bayesnec`
2.1.3.36, `cmdstanr`, 2026-09-13 and 2026-09-14.

`ecxhormebc5` failed in all six `lum31` fits, which use
`crf(log(conc), ...)` under `Gamma(link = "identity")`. `log(conc)` runs
-5.203 to 2.754 on that data and is negative for 64 per cent of observations.
The same equation fitted without difficulty in the same document on `coral_pam`,
which uses `crf(diuron, ...)` under `Beta(link = "identity")` with a predictor
that is never negative.

The mean function is

```
bot + (top - bot + exp(slope) * x) / (1 + exp(exp(beta) * (x - ec50)))
```

so `exp(slope) * x` is negative wherever `x` is, and a draw with a large
`exp(slope)` puts the mean below zero, which a Gamma with an identity link
cannot represent. Initialisation samples from the prior and has to find a point
where the mean is positive at every observation. The failure message is
"Initialization failed", not divergences or a high R-hat, and it is unchanged by
sample size: identical on 44 rows and on 1452.

<details>
<summary>Implementation detail</summary>

### The obvious rule would be wrong

Three other equations use the same `exp(slope) * x` term and are **not** affected:

| equation | mean function contains | on `lum31`, section 4 |
|---|---|---|
| `nechorme` | `(top + exp(slope) * x) * exp(...)` | fitted, weight 0.00 |
| `nechorme4` | `bot + ((top + exp(slope) * x) - bot) * exp(...)` | fitted, weight 0.00 |
[...]

PR #351, opened 2026-09-15
Exclude ecxhormebc5 on negative constrained predictors

beckyfisher:
## What

`check_models()` now excludes `ecxhormebc5` before fitting when the predictor
contains negative values, the mean uses an identity link, and the response
family requires a positive mean. The exclusion is recorded in the fitted-call
metadata and reported with its cause.

## Why it matters

These calls previously spent time compiling and attempting initialisation, then
reported only a failed fit. The candidate set and estimates do not change,
because the failed equation contributed no model weight; the user now receives
the reason before fitting begins.

## Evidence

Issue #344 records failure in all six `lum31` fits under
`Gamma(link = "identity")` with a partly negative log-concentration predictor.
The regression tests cover both positive support ranges and confirm that
`ecxhormebc4`, non-negative predictors, Gaussian means and a Gamma log link are
unchanged.

<details>

- The new rule is inside the existing negative-predictor gate in
  `check_models()`.
- The condition derives positive support from `mu_support()` and separately
  requires the identity link.
- The public `models()` documentation and `NEWS.md` state the restriction.
- The development version is incremented to `2.1.3.36`.

Checks completed:

- `NOT_CRAN=true devtools::test(filter = "check_models")`: 49 passed.
- Source package build: passed.
- Installed-package reproduction against the built archive: passed.
- `R CMD check --no-manual --no-build-vignettes --no-tests`: completed with
  pre-existing code notes and expected warnings from omitting vignette builds.

</details>

Addresses #344.


beckyfisher:
Independent review completed after two cycles. One genuine problem was found
and fixed: the original regression test supplied an already-transformed
negative predictor and checked only `ecxhormebc5` on the non-negative case.
Commit dc789b85 now exercises positive concentrations through
`crf(log(conc), ...)` and asserts that `ecxhormebc4` and `ecxhormebc5` are both
retained when the transformed predictor is non-negative. The reviewer confirmed
the correction and found no remaining actionable issues.

PR #352, opened 2026-09-15
clarify formula environment support (#319)

beckyfisher:
## What

The formula documentation now distinguishes symbol resolution in R from
function availability in Stan. The `env` argument is also ignored for an
existing formula, as its documentation states.

This follows the independent review of #334, which implemented #319.

## Why it matters

A locally defined R function can now be found while the model frame is built,
but `brms` also writes that call into Stan code. An arbitrary R helper therefore
still fails during Stan compilation unless a corresponding Stan function is
supplied. The earlier documentation did not state that boundary.

## Evidence

The review reproduced the unsupported case with `squared()` and
`rstan::stanc(..., allow_undefined = FALSE)`, which reported an undeclared Stan
identifier. A new test defines `square()` locally for R, uses Stan's built-in
function of the same name, and checks the generated program with the Stan
parser. The two focused test files pass 62 expectations on R 4.6.1.

<details>

- The roxygen source, generated Rd file and NEWS entry now state that a custom
  predictor function needs a Stan definition through `stanvars`, or the
  transformed predictor must be computed in the data first.
- `bayesnecformula()` validates `env` only for character input. Formula objects
  retain their existing environment and ignore `env`, including `env = NULL`.
- Regression tests cover both the API behaviour and valid Stan code generation.

</details>


beckyfisher:
## Review findings that were real problems

The independent review of #334 converged after two cycles. Both findings were
accurate and are addressed in this pull request.

- The documentation implied that resolving a caller-defined R predictor
  function made it fit-ready. `brms` also emits the call into Stan, where an
  arbitrary R helper is undefined. The documentation now states the Stan
  requirement, and a parser-level test covers a function defined for both R and
  Stan.
- `env` was documented as unused for formula objects but was validated for all
  inputs. Validation now applies only to character formulas, and a regression
  test confirms that an existing formula retains its environment.

The second review found no further actionable issue.

beckyfisher (commented):
This looks solid

PR #353, opened 2026-09-15
Load checkout when precompiling vignettes (#340)

### LINKED ISSUE #340: precompile.R renders vignettes against the installed package, not the checkout it is run from
## What

`vignettes/precompile.R` never loads the checkout it is run from. The vignettes
call `library(bayesnec)` in their own setup chunks, which resolves to the
**installed** package, so a vignette rendered from a worktree is rendered against
whatever is in the user library rather than against the branch.

## Why it matters

Rendering a vignette is how a branch's effect on reported numbers is checked. If
the render silently uses a different version, that check establishes nothing, and
nothing in the output says which package produced it.

Nothing announces the mismatch. The installed copy and the branch can state the
same `Version:` string, since a branch's DESCRIPTION is usually only bumped once.
The log prints the backend, the core count and the Stan cache, and not the
library the package was loaded from.

Measured while preparing #238 on 2026-09-13. Current `dev` was merged into the
branch and the vignette re-rendered twice to see what had changed. Both renders
produced output byte-identical to the pre-merge render, which was correct and
uninformative: the installed package was 2.1.3.34 and had none of #318's code,
and the merge had changed nothing about what ran. It was caught only because a
byte-identical diff looked implausible after a `dev` merge that changed
`R/nsec.R` by 97 lines. Rendering against a scratch library built from the merged
source then showed every censoring warning gone and every lower credible bound
changed.

## Evidence

`vignettes/precompile.R` on `dev` at `1ef66e15` contains no `load_all()`, no
`library(bayesnec)` and no `.libPaths()` call. It sets `NOT_CRAN`, the brms
backend, `mc.cores`, the Stan program cache and the optional chunk cache.

Each vignette loads the package itself, for example
`vignettes/example9.Rmd.orig`:

```r
library(bayesnec)
library(ggplot2)
```

<details>

### Options

1. **Load the checkout.** `pkgload::load_all(quiet = TRUE)` before the knit loop.
   The vignettes' own `library(bayesnec)` is then a no-op against the already
   attached namespace. `pkgload` is a `devtools` dependency and is not currently
   in `DESCRIPTION`, so adding it to `Suggests` would be part of this. Note the
   recorded trap that `load_all()` exports internals by default, so a render
   passing this way can still differ from one against an installed build; passing
   `export_all = FALSE` avoids that.
2. **Assert rather than fix.** Compare the loaded package's source to the
   checkout and stop if they differ. This adds no dependency, but leaves the
   caller to install first.
3. **Report only.** Print the path the package was loaded from and its version
   alongside the backend and core count already printed. This does not prevent
   the mistake and would have been enough to catch this one.

Option 1 is what a reader would expect from a script that lives in the repository
and rebuilds that repository's vignettes. Option 3 is worth doing regardless of
which is chosen.

### What to check

- The HPC route in `hpc/` runs the same script inside a container; whichever
  option is taken has to work there, where the package may legitimately be
  installed rather than loaded from source.
- `.github/workflows/precompile-vignettes.yaml` sets `BAYESNEC_BACKEND=rstan`
  and runs on a runner that installs the package. An assertion that the loaded
  package matches the checkout must not fail that job.
- #190 re-renders the whole set before release, and is the run that most needs
  this to be right.

### Definition of done

A render states which package it used, and a render from a worktree cannot
silently use a different one. #190's run is the first that should benefit.

</details>

beckyfisher:
## What

Vignette precompilation now loads the repository checkout by default with
`pkgload::load_all(export_all = FALSE)`. The script reports the `bayesnec`
version, source path and loading mode before knitting.

The HPC job explicitly selects installed-package mode because it already
installs the deployed checkout into an isolated library and verifies that path.

## Why it matters

A direct render from a branch can no longer silently use a different
installation from a user library. Its output is produced by the checked-out
code while retaining the export boundary used by an installed package.

## Evidence

The loader resolved `/tmp/bayesnec-340` in its default source mode after an
older installed namespace had first been attached. Installed mode resolved a
separate build under `/tmp/bayesnec-340-lib/bayesnec`. An invalid mode stopped
with a specific error, and 459 unexported namespace bindings remained absent
from the attached package environment.

An installed-source package check passed the changed dependency and vignette
files. The check reported the repository's existing namespace findings and the
expected missing-vignette warnings caused by `--no-build-vignettes`; it reported
no new error.

Closes #340.

<details>

- Add `pkgload` to `Suggests`.
- Default `BAYESNEC_PRECOMPILE_PACKAGE` to `source`; accept only `source` and
  `installed`.
- Use `helpers = FALSE`, `attach_testthat = FALSE` and `export_all = FALSE` for
  the source checkout.
- Set `BAYESNEC_PRECOMPILE_PACKAGE=installed` only in `hpc/run.precompile`,
  after that route's existing job-library assertion.
- Document both paths in `hpc/README.md` and record the user-visible correction
  in `NEWS.md`.

Checks performed:

```text
R CMD build --no-build-vignettes /tmp/bayesnec-340
_R_CHECK_FORCE_SUGGESTS_=false R CMD check --no-manual \
  --no-build-vignettes --no-tests bayesnec_2.1.3.35.tar.gz
```

</details>


beckyfisher:
Independent review completed after two cycles. One substantive problem was
identified and fixed:

- Source mode disabled test helpers but still allowed `pkgload` to attach
  `testthat` through its separate default. This could expose unqualified testing
  functions that are unavailable in the installed HPC environment and require
  `testthat` unnecessarily during a direct render. Commit `31bd2c76` now passes
  `attach_testthat = FALSE`; the reviewer verified the corrected search path and
  found no remaining actionable issues.

PR #355, opened 2026-09-15
Speed up and isolate the example8 production render

beckyfisher:
## Summary

- construct the Gamma family at each example8 call site so its closures cannot retain the growing knitr evaluation environment
- add named render slots with isolated deployment trees, job names, libraries, outputs, provenance records and Stan caches
- allow a matching remote container to be reused without another transfer
- ignore slot-specific deployment records

## Why

The parallel production render was serialising prior fitted objects through a retained family environment. That enlarged future globals to several gigabytes and caused excessive aggregate memory use. Reconstructing the equivalent family removes that captured state. Named slots make it possible to validate the change alongside an existing render without sharing writable paths.

## Validation

- vignette code extracted and parsed successfully
- synthetic family-environment check reduced serialised state from about 50 MiB to 0.2 MiB while preserving the family and link behaviour
- shell syntax and option-validation checks passed
- an isolated 16-worker production render started successfully alongside the existing render and reported the expected branch, worker count and private cache path

The production render is still running, so its final elapsed time and peak memory are not yet available.

## Stack

This PR targets the head branch of #228 so the diff contains only the render changes. It can be merged there while #228 remains open.


beckyfisher:
## Grounds for closing

### The family workaround

The first commit constructs the Gamma family at each `example8` call site so
that its closure cannot retain the knitr evaluation environment. That hole was
closed in the package itself by 1c31bff5, merged to `dev` in #354 (#329).
`validate_family()` now rebuilds any evaluated family through
`bnec_default_family()`, whatever it arrived as, and the comment there names
this symptom: keeping an object such as `fam <- Gamma(link = "identity")`
unchanged put the whole calling environment into `brm_args` and sent it to every
future worker. A vignette-source workaround for it would now be a second fix for
a fixed defect, and one that a reader of `example8` would have to be told the
reason for.

### The render slots

The named slots in `hpc/precompile-hpc.sh` exist to run a 16-worker production
render of `example8` alongside another. #357, merged to `dev` today, removes the
production render as the thing that fits the models: with `BAYESNEC_FIT_STORE`
set, the 189 fits are read from the store the compendium at
open-AIMS/grouping-structures assembles from a 189-task array. Measured there on
2026-09-15, the array held 22.2 hours of fitting and ran in 52 minutes, and the
whole vignette then rendered against the complete store in 38.5 minutes without
sampling. There is no longer a long parallel render for the slots to isolate.

### The residue

Isolating concurrent renders is useful independently of `example8`, and nothing
here rules on that. Should it be wanted, it belongs in a fresh pull request
against `dev` rather than in this one: this branch targets the head branch of
#228, which is about to be rebuilt for the log-scale re-analysis, so anything
left here is orphaned when that happens. The three commits are recorded at
b0990c95 and can be cherry-picked.

The branch is left in place.

PR #356, opened 2026-09-15
example7: apply review comments, add a summary and the worked-example figure

beckyfisher:
## What

Applies the review comments on `example7` from the *Modelling growth data*
document, adds a summary at the front, and restores the worked-example figure
that was dropped when the vignette was rewritten.

## Why it matters

The reviewer asked for examples, on the grounds that a reader who has not seen
the effect in their own data is being asked to trust tables of bias that are
abstract on their own. Figure 2 is that: one simulated dataset, fitted under all
six conventions, with each model-averaged curve and its credible band drawn
against the true curve. It sits between the old Figures 1 and 2, where the
reviewer asked for it.

Two of the review comments turned out to be more than wording. The NSEC was
defined as the concentration at which the response *first becomes statistically
indistinguishable* from the control, which is the wrong way round: read up the
concentration series the response starts indistinguishable and becomes
distinguishable. And the `nec` parameter was used in Section 2.3 without being
defined, which the reviewer flagged.

## Evidence

Figure 2 shows what the bias tables cannot. The three conventions that put a
zero in place of a measurement produce a curve that flattens onto zero rather
than continuing down, and the flattening begins a little below the true ErC50 —
so the curve is displaced over a range that includes the concentrations the
estimates are read from, not only where the substitution was made. The `deleted`
panel shows the fitted curve stopping early, because removing every negative row
removed the highest concentrations outright. The `gamma` panel shows a credible
band that is widest at the control, where the generated spread is no wider than
anywhere else.

Six new fits were run for it. The exemplars already saved draw a different
realisation for each convention, so putting them side by side would have shown
part convention and part realisation noise; all six are now fitted to the same
dataset. The realisation is the one a seed fixed on the cell and convention
names had already chosen for `p1`, so it was not picked after seeing the fits.

A rounding defect surfaced while checking the summary. The transcribed table
stored `rel_bias` at two decimals and the prose rounds it again to one, so a
value on a `.x5` boundary rounded the wrong way. Four of the 126 figures were
reported a step out, including `p1`/`measured` ErC50 as -0.5% where the raw
value is -0.4%.

<details>

### The review document

`ignore/Modelling growth data.docx`, 21 tracked
insertions, 5 deletions and 9 comments. All are applied. Beyond the two above:
the removal list in Section 1 is reordered so the two data-altering conventions
sit together and now records that keeping the values is itself a tested
convention; the section roadmap becomes a list; "the subscript `r`" is reworded,
since the vignette writes `ErC50` inline and the subscript is OECD's; "a defect
of any fitting approach" and "the second of them" are reworded; and the
`deleted` convention's effect on replication is qualified.

### The summary

Supplied by the author and applied verbatim. It quotes no figures, so nothing
in it can go stale against a re-run, and the transcribed metrics table stays
where it was, before Section 5.4.

### Numbering

Figures 2 to 6 become 3 to 7, Sections 5.3 to 5.6 become 5.4 to
5.7, and the four cross-references to them are updated.

### The compendium

`open-AIMS/negative-response-conventions` at `1721e49`, which
`vignettes/bayesnec.bib` now cites. It adds `analysis/vignette_blocks.R` (the
block generator, previously a scratch script that was lost once),
`analysis/run_example_panel.R` with `hpc/run.panel` for the six panel fits, and
`analysis/vignette_panel.R` which turns those fits into the transcribed blocks.
The Beta arm fits a response scaled to (0, 1], so its curve is multiplied back
by `ppc_scale()` before it shares an axis with the other five.

### Checks

Knits with 0 error chunks, 7 figures. `prose-check` run on the added
lines and on the summary in isolation. The rendered HTML contains no home
directory, library path or compiler output.

</details>


beckyfisher:
The summary section has been replaced with text supplied by the author, applied
verbatim, and the pull request body is updated to match.

It quotes no figures, so the restructure that the drafted summary needed is
reverted: the transcribed metrics table and its helpers are back before Section
5.4 and the colour and level definitions before Section 5.2, which is where they
were. Nothing else in the vignette changes, and the review edits, Section 5.3
and the renumbering are as described in the body.

Knits with no error chunks and seven figures.

PR #359, opened 2026-09-15
Filter model equations by parameter count

### LINKED ISSUE #301: Select model equations by number of curve parameters
## What

Add a way to select model equations by the number of curve parameters they
estimate, so a user can restrict a model set to equations their design can
support.

```r
models(max_pars = 3)
bnec(y ~ crf(x, models("decline", max_pars = 3)), data = dat)
```

## Why it matters

A concentration-response design with few distinct concentrations cannot support
a four- or five-parameter curve. Where it is asked to, the mean function is close
to saturated: it can pass through almost any set of concentration means, which
leaves the curve and any group-level deviation nearly confounded. The fit does
not fail cleanly. It samples, reports a plausible estimate, and returns hundreds
of divergent transitions.

There is currently no way to ask which equations are simple enough. `models()`
already computes each equation's parameters through `show_params()` and returns
them, so the information is present and only the selection is missing. A user has
to know the equations well enough to name a subset by hand.

## Evidence

Measured on `CoralColour` (180 rows, **five** distinct concentrations, 45
chambers of 4 fragments), `Beta(link = "identity")`, `ogl(chamber)`, 2 chains,
2000 iterations, `adapt_delta = 0.80`, seed 228:

| equation | curve parameters | divergent | R-hat | min ESS |
|---|---|---|---|---|
| `nec4param` | 4 | 291/800 | 1.098 | 29 |
| `ecx4param` | 4 | 150/800 | 1.029 | 126 |
| `ecxll3` | 3 | **0**/800 | 1.012 | 201 |
| `ecxwb2p3` | 3 | **0**/800 | 1.014 | 165 |
| `ecxwb1p3` | 3 | **0**/800 | 1.019 | 181 |
| `ecxexp` | 2 | **0**/800 | 1.043 | 117 |

The group-level standard deviation is estimated at 0.42 to 0.45 by the
four-parameter and the three-parameter equations alike, so the effect being
estimated is stable and it is the sampler that fails. Raising `adapt_delta` to
0.99 and quadrupling the iterations reduced the divergence rate from 36 per cent
to 1.4 per cent but left a maximum R-hat of 1.381 and a minimum effective sample
size of 5, so this is not a step-size problem and cannot be tuned away.

<details>

`R/models.R`: `models()` already resolves a set of equations and calls
`show_params()` on it. The parameter count is `length(names(bf_<model>$pforms))`.
Adding a `max_pars` argument that filters the resolved set before returning is
the whole change.

Equation counts as of 2.1.3.33:

| parameters | equations |
|---|---|
| 2 | `ecxexp`, `ecxlin` |
| 3 | `ecxll3`, `ecxsigm`, `ecxwb1p3`, `ecxwb2p3`, `nec3param`, `neclin` |
| 4 | `ecx4param`, `ecxhormebc4`, `ecxll4`, `ecxwb1`, `ecxwb2`, `nec4param`, `nechorme`, `nechormepwr`, `nechormepwr01`, `neclinhorme`, `necsigm` |
| 5 | `ecxhormebc5`, `ecxll5`, `nechorme4`, `nechorme4pwr` |

`check_models()` continues to remove equations that are invalid for the family
and predictor, as it does now, so the filter does not need to consider validity.

Tests: that `max_pars` returns only equations at or below the count; that it
composes with the existing group and numeric-range arguments; that a value
admitting no equation gives an informative error rather than an empty set
reaching `bnec()`.

Documentation: `?models`, and a sentence in the vignette that discusses design
and model choice.

</details>

beckyfisher:
## What

- Add an optional `max_pars` argument to `models()`.
- Apply the limit after resolving an existing model group, response range, or
  fitted object.
- Allow the named equation list returned by `models()` to be used directly in
  `crf()`, including the fitting call shown in issue #301.
- Validate the limit and report an informative error when no equations remain.
- Document the new filter and add regression coverage for every model's curve
  parameter count.

## Why it matters

Designs with few distinct predictor concentrations may not support equations
with four or five curve parameters. Callers can now select a suitably bounded
model set before fitting, while existing `models()` behaviour remains unchanged.

## Evidence

- Focused `models()` and formula-parser tests: 111 assertions passed.
- Source package installed successfully in a clean temporary library.
- Installed-package checks covered the complete count distribution, standalone
  and grouped filtering, and the empty-result error.
- `git diff --check` passed.

<details>

The count is derived from each equation's `pforms`, the same parameter metadata
returned by `show_params()`. The filter therefore follows the package's own
equation definitions rather than maintaining a second production lookup table.

`models()` with no arguments still returns the model-group list. Supplying
only `max_pars` resolves the full model set before filtering it.

An empty filtered selection reports that no selected equations meet the limit,
which remains accurate when the limit is combined with a group or fitted
object.

</details>

Closes #301


beckyfisher:
Independent review completed after three cycles; the developer and reviewer
agree that the PR is sound. The PR body has also been updated to describe the
current implementation and evidence.

Real problems found and fixed:

- The issue's direct `crf(x, models(...))` syntax initially failed because the
  formula parser accepted only character vectors. It now accepts the named list
  of equation formulas returned by `models()` and retains strict rejection of
  malformed lists.
- The empty-selection error made a false package-wide claim when a group or
  fitted object was filtered. It now refers to the selected equations and has a
  grouped-empty regression test.
- Documentation did not distinguish unfiltered `models()` from a missing
  `object` with `max_pars`, and later described accepted model-list values too
  broadly. The function, formula, generated reference, and vignette text now
  state the precise behaviour consistently.

PR #366, opened 2026-09-16
Order the pkgdown articles index and navbar explicitly (#363)

### LINKED ISSUE #363: Order the pkgdown articles index explicitly rather than by vignette file name
## What

Set the order of the articles index and the Articles navbar dropdown explicitly
in `_pkgdown.yml`, in place of the alphabetical order the vignette file names
currently produce.

## Why it matters

`_pkgdown.yml` has no `articles:` key, so pkgdown lists the vignettes
alphabetically by file name. That order is an accident of when each vignette was
written. It happens to approximate a usable reading order, because each vignette
was written after the ones it cites, but nothing holds it there: the next
vignette added takes its position from its number alone. `example8` and
`example9`, both in open pull requests, are the first case where the number and
the content disagree.

`example5`, the C++ toolchain and CmdStan setup, is the one placement the
current order gets wrong on any reading. It sits sixth, between the vignettes on
posterior comparison and hurdle families.

## Evidence

Measured on 2026-09-16 against `origin/dev` at `3179f9cb` and the published
development site. The site listing below was read before PR #362 merged, so it
shows eight entries; `dev` now holds nine rendered vignettes and the site gains
`example9` at its next build.

`_pkgdown.yml` on `origin/dev` sets `url`, `template` and `development` only.
The articles index at
`https://open-aims.github.io/bayesnec/dev/articles/index.html` and the Articles
dropdown on the same site list the same eight entries in the same order:
`example1`, `example2`, `example2b`, `example3`, `example4`, `example5`,
`example6`, `example7`. The two listings agree because pkgdown derives the
navbar dropdown from the index it builds.

PR #362 (#190) merged at 2026-09-16T00:18 and added `vignettes/example9.Rmd`,
so "A complete analysis workflow" is now on `dev`. `example8` is the only one of
the ten still absent: `vignettes/example8.Rmd.orig`, "Grouping and factor
covariates", is on `issue-6-33-grouping-vignette` (PR #228).

### The deference graph

Counted from the body text of all ten `.Rmd.orig` files, excluding the
link-definition blocks at the top, and covering the three link styles in use:
`[e2b]` reference links, inline `articles/exampleN.html` URLs, and
`vignette("exampleN")` calls.

| Vignette | Defers to | Is deferred to by |
|---|---|---|
| `example1` | 2, 2b, 3, 4, 5 | 2, 5, 7, 8, 9 |
| `example2` | 1, 2b, 4, 5 | 1, 6, 9 |
| `example2b` | 4 | 1 (×5), 2 (×3), 9 (×2) |
| `example3` | — | 1, 8, 9 |
| `example4` | — | 1, 2, 2b |
| `example5` | 1 | 1, 2 |
| `example6` | 2 | 7, 8, 9 |
| `example7` | 1, 2, 6 | — |
| `example8` | 1, 3, 6 | — |
| `example9` | 1, 2, 2b, 3, 6 | — |

`example3` and `example4` cite no other vignette and are cited by the core
three, so they are self-contained. `example7`, `example8` and `example9` cite
others and are cited by none, so they are the applied end of the set. Nothing in
the graph contradicts the relative order of `example1`, `example2` and
`example2b`.

No vignette refers to another by position. A search of all ten files for "next
vignette", "previous vignette", "first vignette", "earlier vignette" and
"vignette *N*" returned nothing, so reordering the index contradicts no prose.

## Proposed order

Four sections. Within them the order follows the deference table above, so that
no vignette is placed before one it sends the reader to.

### Getting started

What a reader works through on first use.

1. `example5` Installation and setup. The C++ toolchain, CmdStan and the choice
   of back end. `example1` gives a two-line `install.packages()` and defers here
   for the detail, so this is placed first for the reader whose first fit fails
   to compile rather than as required reading.
2. `example1` Single model usage. One model equation, the families, and reading
   a fit.
3. `example2` Multi model usage. The model set, stacking weights and model
   averaging, which every later vignette assumes.
4. `example2b` Model details. The 23 equations, the parameter definitions, the
   model groups and which equations suit which response type. `example1` sends
   the reader here five times in its body and `example2` three times, each time
   for a decision made while working through that vignette, so it belongs beside
   them rather than at the end.

### Extending the analysis

What is changed once a fit runs. Both cite no other vignette.

5. `example3` Priors. Default and user-specified priors.
6. `example4` Comparing posterior predictions. `compare_posterior()` across
   separate fits.

### Designs and response types

The data shapes that need something other than the default fit.

7. `example6` Hurdle and zero-inflated concentration-response models. Placed
   before `example7`, `example8` and `example9`, all three of which send the
   reader to it.
8. `example7` Modelling growth data and other potentially negative response
   values.
9. `example8` Grouping and factor covariates.

### A worked analysis

10. `example9` A complete analysis workflow. An end-to-end analysis that cites
    five of the nine before it and is cited by none, so it reads as the
    consolidation of the set.

<details>

### The change

One file, `_pkgdown.yml`, on a branch off `dev`. Add an `articles:` block after
`development:`:

```yaml
articles:
  - title: Getting started
    contents:
      - example5
      - example1
      - example2
      - example2b
  - title: Extending the analysis
    contents:
      - example3
      - example4
  - title: Designs and response types
    contents:
      - example6
      - example7
      # example8 is added here by #228
  - title: A worked analysis
    contents:
      - example9
```

### Sequencing against the open pull requests

`example9` landed with PR #362. `example8` is still on PR #228, which is
conflicting with `dev` and has PR #355 stacked on it, so it is not close.

pkgdown resolves each `contents:` entry against the vignettes present in the
build, so naming a vignette that has not landed is what has to be avoided.
Waiting for #228 would hold this behind an unrelated conflict, so the block goes
in now with the nine vignettes on `dev`, and `example8` is added to it inside
#228 alongside the vignette itself.

`_pkgdown.yml` is listed in `.Rbuildignore`, so `R CMD check` never reads it and
this cannot alter package behaviour. It goes straight to `dev` under the rule
for changes of that kind, and the pkgdown job on `dev` is the same check a pull
request would run. Confirm before pushing, because a malformed block breaks the
site build: build the site locally with
`pkgdown::build_site()` and check that no vignette is reported as missing from
the index and that no named entry fails to resolve. Adding an `articles:` block
makes the listing explicit, so every vignette added after this must also be
added to the block.

### File names

Renaming the files to match the reading order was considered and rejected. The
names `exampleN` are referenced from nine files under `R/` and from the `.Rd`
files generated from them, from cross-links between the vignettes themselves,
and from every published article URL, including those on the CRAN release site.
The index order is presentation and is set in `_pkgdown.yml`; the file names are
an interface.

### Definition of done

- `_pkgdown.yml` has an `articles:` block giving the order above.
- A local `pkgdown::build_site()` reports no vignette missing from the index and
  no unresolved entry.
- The development site shows the new order in both the articles index and the
  Articles dropdown.

### Points left open

- The section titles. "Getting started", "Extending the analysis", "Designs and
  response types" and "A worked analysis" are proposed here, not settled.
- The position of `example5`. The deference table puts it low: two vignettes
  cite it and it is toolchain detail rather than a first read. It is placed
  first here on the argument that a reader whose first fit fails to compile
  should not have to hunt for it. Ninth, beside the other reference material,
  is defensible on the evidence.
- Whether `example8` supersedes part of `example4`. `example4` states that
  `bnec` does not allow an interaction with a fixed factor and recommends
  independent fits per factor level as the workaround. `example8` adds
  group-level terms. That is a content question for #228 rather than an ordering
  one, but it affects whether the two belong in the same section.

</details>

beckyfisher:
## What

`_pkgdown.yml` gains an `articles:` block that sets the order of the articles
index and of the Articles dropdown. The order is the one proposed in #363,
grouped into four sections: Getting started, Extending the analysis, Designs and
response types, and A worked analysis.

`notes/vignette_numbering.md` gains the obligation the block creates. The block
aborts the site build on a vignette that is present but unlisted, and on an
entry that names no vignette in the build, so the register that a new vignette's
author is told to read now names `_pkgdown.yml` as a second place the number is
recorded and gives the timing that avoids both.

Closes #363. The base branch is `dev`, so the keyword does not fire on merge and
the issue is closed by hand.

## Why it matters

With no `articles:` key pkgdown lists the vignettes alphabetically by file name.
That order records when each vignette was written, not what a reader needs
first, and the next vignette added takes its position from its number alone.
`example5`, the C++ toolchain and CmdStan setup, currently sits sixth, between
the vignettes on posterior comparison and on hurdle families.

## Evidence

Measured on 2026-09-16 with pkgdown 2.2.1 under R 4.6.1, at `611e6842`. The
`articles:` block is unchanged since that commit, the later commits on this
branch touching only the config comment and `notes/`, so the measurements below
hold for the head. The pkgdown job on this pull request installs pkgdown 2.2.1
as well, so the local measurements and the CI build are on the same version. The
3m40s figure below is run 35074558280, for `611e6842`.

`pkgdown::check_pkgdown()` reports "No problems found", which is the validation
named in the issue: it calls `data_articles_index()`, so every vignette present
is listed and every listed entry resolves. The nine vignettes on `dev` are
`example1`, `example2`, `example2b`, `example3`, `example4`, `example5`,
`example6`, `example7` and `example9`. `example8` is on #228 and is added to the
block there.

`pkgdown::build_articles_index()` was run and the rendered `articles/index.html`
read. The index body lists the four section titles with the nine vignettes in
the order above, and the Articles dropdown in the same page holds all nine,
grouped, with no "More articles..." link. The pkgdown job on this pull request
then built the whole site and wrote `articles/index.html` and all nine article
pages in 3m40s.

The issue's implementation sketch has a defect, found in
`pkgdown:::navbar_articles()` and confirmed by rendering. An `articles:` block
in which no section sets `navbar:` replaces the dropdown with a plain link to
the index: the menu built from the sketch was `{text: "Articles", href:
"articles/index.html"}` and nothing else. The block as committed gives each
section a `navbar:` key, which is what keeps the dropdown and puts the nine
entries in it. Without that the change would have removed the dropdown the issue
set out to reorder.

No check would have reported that. `check_pkgdown()` calls `check_urls()`,
`data_open_graph()`, `data_articles_index()` and `data_reference_index()`, and
none of them reads the navbar. Run against the sketch config it reports "No
problems found", and the site builds and publishes the plain link. The comment
in `_pkgdown.yml` is the only guard, which is why it states the behaviour rather
than pointing at it.

## Points the issue left open

Three, and each is a decision for the reviewer rather than a defect in the
block. They are above the fold because a reader who approves without expanding
the detail should still know they were asked.

The four section titles are proposed, not settled. "Extending the analysis" is
the weakest of the four: `example3` covers priors, which is not an extension but
the first thing to change on a fit that does not sample cleanly, and `bnec()`'s
own documentation sends the reader there. "Priors and posterior comparison" is
one alternative, and listing `example3` under Getting started is another.

`example5` is placed first on the argument that a reader whose first fit fails
to compile should not have to hunt for it. The deference graph in #363 would put
it ninth, beside the other reference material. A one-entry Setup section ahead
of Getting started is a third option.

Whether `example8` belongs beside `example4`, which recommends independent fits
per factor level as the workaround for the group terms `example8` adds, is a
question for #228.

<details>

### The change

Two files. `_pkgdown.yml` takes one block after `development:`, with a comment
recording the pkgdown behaviour the block depends on and nothing that is
recoverable from #363. `notes/vignette_numbering.md` takes a fifth rule and a
change to its opening instruction; its `example7` and `example9` rows also said
those vignettes lived on feature branches, and both are on `dev`, so the two
cells are corrected.

`navbar: ~` on the first section gives its four entries with no heading and no
leading divider, which is the idiom `dplyr` and `ggplot2` use. `yaml` parses it
to a named `NULL`, so `has_name(section, "navbar")` is `TRUE` and the section
still counts toward `length(navbar) == length(articles_index)`. Each later
section names its heading, and `pkgdown:::navbar_articles()` emits a divider and
that heading before its entries. All four sections set the key, so pkgdown
appends no "More articles..." link.

### The failure mode of a later vignette

`data_articles_index()` calls `config_abort()` on a vignette that is present in
the build but absent from the block. Removing the `example7` entry and
re-running it gives

```
In _pkgdown.yml, 1 vignette missing from index: "example7".
```

and naming a vignette that is not in the build gives

```
In _pkgdown.yml, articles[3].contents[3] (example8) must be a known topic name
or alias.
```

so the listing cannot diverge from the vignettes without the build saying so.
Making the order explicit requires that, so a pull request adding a vignette now
adds its entry as well.

The second abort sets the timing. `pkgdown:::package_vignettes()` matches
`\.[Rrq]md$` and so never sees a `.Rmd.orig`, and
`notes/implementation/00_protocol.md` tells a session not to run
`precompile.R`, so a new vignette exists as `.Rmd.orig` alone until its render
lands. An entry added over that gap fails every pkgdown run in the interval.
Register rule 5 therefore holds the entry until the rendered
`vignettes/exampleN.Rmd` exists rather than tying it to the source. It also
notes that the `precompile vignettes` workflow stages `vignettes/*.Rmd` and
`vignettes/*.png` and nothing else, so the entry is a separate commit on
whichever branch the rendered `.Rmd` lands on, and that the pkgdown job
reports either omission on a push to `master` or `dev` and on a pull request
whose base is one of them, and nowhere else. The obligation is in the register
rather than only in the config comment because a vignette author has no reason
to open `_pkgdown.yml`. For `example8`, which #228 adds as
both `.Rmd.orig` and rendered `.Rmd`, it is also recorded on that pull request.

### The vignette file names

`exampleN` is referenced from eleven files under `R/`, twenty-one times in all,
counted with `grep -rnE 'example[0-9]+b?' R/` on this branch. Nine of the eleven
reference a vignette from a roxygen `vignette()` call; `ecx.R` and `helpers.R`
do so only from ordinary comments. It is referenced again from the `.Rd` files generated from those
blocks, from the cross-links between the vignettes, and from every published
article URL including those on the CRAN release site. The index order is
presentation and lives in `_pkgdown.yml`; the names are an interface, so
renaming the files to match the reading order was rejected.

### Checks

`_pkgdown.yml` is listed in `.Rbuildignore`, so `R CMD check` never reads it and
this cannot alter package behaviour. `pkgdown.yaml` runs on `pull_request` to
`dev`, so the site build on this pull request is the check that applies.

#363 decided this would go straight to `dev` under the rule for a change that
cannot alter package behaviour, and that is not what happened. The rule was
written for `notes/` and prose, which no job reads; a malformed `articles:`
block breaks the published site, and the navbar defect above is an instance of
the class. The four-platform `R CMD check` matrix this opens is the price of
that, and the rule is worth narrowing so `_pkgdown.yml` sits outside it.

### The published release site

`development: mode: auto` resolves this branch to `docs/dev`, so the new order
appears at `/bayesnec/dev/articles/index.html`. The root site is published from
`master`, which does not have the block, and stays alphabetical until `dev`
merges to `master`.

</details>


beckyfisher:
## Review outcome

Five review cycles were run against this branch. What follows is only the
findings that were real defects and are now fixed; the reasoning behind each fix
is in the body above, which was rewritten as they were addressed.

Two were defects in the change as first pushed.

- **The `example8` sentence in the `_pkgdown.yml` comment was an instruction to
  a future pull request**, not a description of the file, and nothing would have
  prompted its removal once #228 merged. It is now on #228 and, in general form,
  in the register.
- **"referenced from nine files under `R/`" counted roxygen `vignette()` calls
  only.** Eleven files reference a vignette, twenty-one times; `ecx.R` and
  `helpers.R` do so from ordinary comments.

Three were gaps in what the change recorded rather than in what it does.

- **`pkgdown::check_pkgdown()` does not read the navbar.** It calls
  `check_urls()`, `data_open_graph()`, `data_articles_index()` and
  `data_reference_index()`, and reports "No problems found" on a config that
  publishes a plain link in place of the dropdown. No job would report a later
  edit that drops a `navbar:` key, which is why the comment states the behaviour
  rather than pointing at it.
- **The obligation the block creates was recorded only in the config comment**,
  which a vignette author has no reason to open. It is now rule 5 of
  `notes/vignette_numbering.md`, the register such an author is told to read.
- **#228 held nothing about the entry it now has to add.** Comment 5694710152
  gives the branch, the four lines and the definition of done.

Three were defects in the register text added during the review, each caught on
the following cycle.

- **The first wording tied the `_pkgdown.yml` entry to the commit that creates
  `vignettes/exampleN.Rmd.orig`.** `pkgdown:::package_vignettes()` matches
  `\.[Rrq]md$` and never sees a `.Rmd.orig`, and
  `notes/implementation/00_protocol.md` tells a session not to run
  `precompile.R`, so following that instruction would have failed every pkgdown
  run until a render landed.
- **The second wording tied it to the commit that adds the rendered `.Rmd`**,
  which the `precompile vignettes` workflow cannot produce: its pull request
  stages `vignettes/*.Rmd` and `vignettes/*.png` and nothing else. Rule 5 now
  states the invariant instead.
- **Rule 5 named only the pull-request trigger.** `pkgdown.yaml` also runs on a
  push to `master` or `dev`, which is how a missing entry that reaches `dev`
  presents itself.

Three findings were declined or left open rather than fixed. The four section
titles and the placement of `example5` stay as #363 proposes them, because that
issue states both are proposed rather than settled; the alternatives argued in
review are recorded above the fold for you to choose between. The review also
found that the pkgdown site publishes `CLAUDE.md` as a public page —
`/bayesnec/dev/CLAUDE.html` returns 200 with the branch policy and the modelling
gotchas in it — which is real but pre-existing and unrelated to this change, so
it is raised separately rather than absorbed here.

PR #370, opened 2026-09-16
Compare fit data columns by name (#361)

### LINKED ISSUE #361: c.bnecfit() refuses fits of the same data when a hormesis equation is among them
## What

`c.bnecfit()` stops with *\"Dataset values differ across fits. Datasets need to
be identical across the multiple fits\"* when combining equations fitted to one
and the same data frame, if the set mixes a hormesis equation with a
non-hormesis one.

The data do not differ. Only the **column order** of `fit$data` does.

## Why it matters

`c.bnecfit()` and `+.bnecfit()` are the documented way to build a
`bayesmanecfit` out of separately fitted equations, and `?c.bnecfit`'s own
example is adding one equation to an existing set. The refusal blocks that for
any set containing `nechorme`, `nechorme4`, `nechormepwr01`, `ecxhormebc4` or
`ecxhormebc5` alongside anything else, which is most of `"all"`.

A monolithic `bnec()` call is unaffected: it hands its `prebayesnecfit` list
straight to `expand_manec()` and never reaches the check. So the same 18
equations combine when `bnec()` fits them and refuse when they are fitted one at
a time and combined afterwards.

## Evidence

Measured on `dev` at `5f98d8d6`, fitting `example8`'s `fit_pam` call one
equation at a time --- `yield ~ crf(diuron, "all") + ogl(chamber)` on
`coral_pam`, `family` chosen by `bnec()`, 18 equations, 414 rows each:

| equations | `names(fit$data)` |
|---|---|
| the 13 non-hormesis ones | `yield, chamber, diuron` |
| `nechorme`, `nechorme4`, `nechormepwr01`, `ecxhormebc4`, `ecxhormebc5` | `yield, diuron, chamber` |

Every value in all 414 rows is equal between the two orders, and the column
names are the same set. `check_data_equality()` compares
`as.matrix(x$fit$data)` with `identical()`, which is order-sensitive, so it
reports 828 differing cells --- the whole of both swapped columns --- and stops.

<details>

`R/helpers.R`, `check_data_equality()`. The first check is

```r
data_are_equal <- lapply(mod_fits, function(x) as.matrix(x$fit$data)) |>
  Reduce(f = identical_value) |>
  is.matrix()
```

`brms` orders `fit$data` by the order the variables appear in the model formula,
and the hormesis equations reference the predictor earlier in their non-linear
formula than the others do. Nothing about the data or the fit differs.

The second check in the same function already compares `rownames()` of the terms
attribute **after `sort()`**, so it is order-insensitive by construction. The
first one was presumably meant to be as well.

The fix is to compare the columns by name rather than by position: order both
sides by `names()` before `as.matrix()`, or compare the data frames column by
column keyed on name. The `as.matrix()` coercion is worth a second look while
there --- it turns a mixed data frame into a character matrix, which is why the
example above reports a factor level against a formatted number --- but that is
a separate question from the ordering.

A test would fit two equations on one data frame, one hormesis and one not, and
assert that `c()` returns a `bayesmanecfit`. The existing tests do not catch it
because they combine equations pulled out of a single `manec_example`, which
were all fitted in one `bnec()` call and so share one `fit$data`.

</details>

beckyfisher:
## What

`c.bnecfit()` now compares stored fit data by column name, so separately fitted equations can be combined when `brms` stored identical columns in a different order. Genuinely different values remain an error.

## Why it matters

Hormesis and non-hormesis equations can refer to variables in different formula orders. `brms` then stores the same data columns in different orders, which previously prevented `c.bnecfit()` and `+.bnecfit()` from combining those valid fits.

## Evidence

The installed-package regression file passed 47 assertions. The new test reverses the stored column order and confirms that `c()` returns a `bayesmanecfit`; it then changes one value and confirms that the existing refusal remains.

All required GitHub checks passed: R-CMD-check on macOS release, Windows release, Ubuntu release and Ubuntu devel, plus pkgdown.

<details>

`check_data_equality()` orders every stored data frame by its column names before applying the existing matrix comparison. Rows, names, values, and the formula-term comparison retain their existing checks.

</details>

Closes #361.


beckyfisher:
Independent review follow-up

The PR body has been updated with the completed verification results. The independent review completed one cycle and identified no actionable problems in the original PR, so no review-driven fixes or disputed findings were required. The reviewer then reviewed this disposition and agreed that the PR is sound.

PR #371, opened 2026-09-16
Show grouping levels in ggplot output

### LINKED ISSUE #368: Distinguish the levels of a grouping in autoplot() and ggbnec_data()
## What

`autoplot()` and `ggbnec_data()` gain a way to distinguish the levels of a
grouping in the observed points, so that a fit that includes `ogl(plate)` or
`(top | plate)` can be drawn with the plates told apart.

## Why it matters

Every figure of a grouped fit currently draws the observations as one
undifferentiated cloud. On `lum31` that cloud is the whole point of the
section: the vertical spread at a concentration is the per-plate gain
difference, which is what the group-level term absorbs. A reader looking at the
figure cannot see that the spread is between plates rather than within them,
which is the one thing the display is there to show.

The same applies to `bnec_group()`, where the levels are in separate fits and a
reader comparing them has no way to see which observations belong to which.

## Evidence

`vignettes/example8.Rmd.orig`, section *Across-concentration grouping*, states
that control wells range from about 250,000 to 1,450,000 relative light units
across seventeen plates, and that the pooled fit has nowhere to put that spread
except the residual. The figure beside it shows a band and a cloud, and neither
the plates nor the spread between them is visible in it. The prose states the
claim alone.

The raw-data figure added earlier in the same vignette draws the plates with
`aes(group = plate)` and a per-plate mean line, and the structure is immediately
legible. That is a hand-rolled `ggplot` on the data frame, not the package's own
display of the fit.

<details>

The grouping is already known to the object: `ogl()`, `pgl()` and
`(par | group)` all record the grouping variable, reachable through the
`bayesnecformula` and present as a column in `fit$fit$data`.

Two decisions to make before writing code.

#### The function the argument belongs on

`ggbnec_data()` returns the data frame
`autoplot()` draws, and adding a column there lets a user build any display they
want, which is the more general fix. `autoplot()` would then gain an argument
that maps it to colour or shape. Adding it only to `autoplot()` would be less
work and leaves a user building their own figure no better off.

#### The mapping

Colour is the obvious mapping and fails at seventeen
plates, where a categorical scale is unreadable. Alternatives worth measuring
against the `lum31` figure: a per-level mean line of the kind the raw-data
figure uses, faceting, or drawing the levels in one muted colour with the
population curve on top. The `dataviz` guidance on categorical colour applies;
above about eight levels colour stops distinguishing.

Not in scope: the fitted curve stays the population curve. Drawing a per-level
curve is a different request and depends on the term, since `ogl()` displaces
and `pgl()` reshapes.

Definition of done: the `lum31` `ogl(plate)` figure in `vignette("example8")`
shows the between-plate spread as between-plate; a `bnec_group()` fit's levels
can be told apart; `?autoplot.bayesnecfit` and `?ggbnec_data` document the
argument; `tests/testthat/test-autoplot.R` covers a grouped fit with and
without it.

Stop and ask rather than guess about the mapping at high level counts --- what
works at three plates and what works at seventeen may not be the same argument,
and picking one silently would settle a display question the vignette depends
on.

</details>

beckyfisher:
## Grouped observations

`ggbnec_data()` and `autoplot()` now accept `group`, naming a group-level
variable from the fitted formula. The returned plotting data include the
selected levels, and the plot draws their observed per-level means beneath the
population curve. `bayesnecgroupfit` objects are also supported directly, with
one panel per independently fitted level.

## Reason

Grouped fits previously drew every observation as one undifferentiated cloud.
The new mean traces show whether vertical spread occurs between grouping levels
without relying on a categorical colour scale, which is unsuitable for the 17
plates in the motivating `lum31` analysis. A grouping recorded at only one
predictor value receives a mean marker because a one-point line has no extent.

## Verification

The plotting tests pass 49 assertions against both the source and an installed
package. A broader selection covering plotting methods, grouped fits, hurdle
methods and class expansion passes 344 assertions; its two warnings are existing
hurdle-fit warnings. Package loading, S3 consistency, documentation, usage and
examples pass in `R CMD check`.

<details>

- `ggbnec_data()` returns the selected values in a stable `group` column and
  records the source column in the `group_var` attribute.
- `autoplot()` adds per-level mean traces for levels spanning the predictor and
  mean markers for levels confined to one predictor value.
- New methods combine `bayesnecgroupfit` plotting data and facet its separate
  fitted curves by level.
- `group` follows `...` in the `ggbnec_data()` signatures, preserving existing
  positional arguments passed to `ecx()`.
- Tests cover single and model-averaged fits, grouped-fit dispatch, invalid
  selections, single-value groupings and positional compatibility.

The package check was run without rebuilding vignettes. Its remaining findings
are the existing `modifyList` import note, the worktree `.git` note, and the
expected missing-`inst/doc` warnings from `--no-build-vignettes`.

</details>

Closes #368


beckyfisher:
Independent review completed after two cycles. One real issue was found and fixed: the motivating `example8` branch still used the ungrouped `autoplot(fit_plate)` call. PR #228 now uses `group = "plate"` in both the source and precompiled vignette, and its figure was regenerated to show all 17 plate-level mean traces (commit 26cb60f6). No package-code defects were found. Because that vignette now uses this API, PR #371 should merge before PR #228.

Issue #367, opened 2026-09-16
example9: show a non-constant coefficient of variation, and confirm the disp() sub-model fixes it


beckyfisher:
## What

`vignette("example9")` gains a worked example of a non-constant coefficient of
variation: the diagnostic that reveals it, the `disp()` sub-model that answers
it, and a posterior check confirming the sub-model fixed what the diagnostic
found.

`vignette("example8")` will forward-reference it. That vignette establishes a
`disp("power")` sub-model on the `lum31` assay from a table of within-cell
coefficients of variation, and stops there: it shows the problem and the
remedy, and never confirms the remedy worked.

## Why it matters

A reader who accepts `disp("power")` on the strength of a CV table has no way
to check the same thing on their own data, because the table is a summary of
the raw response rather than of the fit. The check that does transfer is a
posterior one, and `example9` is where the diagnostics live.

On `lum31` at 15 minutes for zinc, observed within-cell CV runs from about 0.05
at the control to about 0.36 at the bottom of the series. A single `Gamma`
shape parameter compromises between them and is dominated by the noisy low end,
so the fitted variance at the control is far larger than the data show. Two
estimates depend on that: `nsec()` is defined against the control's lower
quantile, and the model weights depend on how variance is apportioned along the
series.

## Evidence

Measured on the `lum31` zinc 15-minute arm, 17 plates, in
`notes/scripts/example8/disp_screen.R` and `disp_smooth.R`. Observed CV 0.051 at
the control against 0.362 at the bottom; the constant-dispersion fit implies
0.43 throughout, about eight times the observed control value.
`disp("power")` brings the implied control CV to 0.079, and an unconstrained
smooth agrees with it there.

<details>
<summary>Implementation detail</summary>

The example needs three things, in this order.

A fit with constant dispersion, and a PIT-ECDF plot of it. The probability
integral transform values of a correctly specified model are uniform; where the
variance is overstated at one end of the predictor they are not, and the ECDF
leaves the simultaneous confidence band. `pp_check(fit, type = "pit_ecdf")` in
`brms` draws it.

The same fit with `disp("power")`, and the same plot, which should sit inside
the band. State what it shows for these data with the numbers, including where
it still departs.

A residual-against-fitted display binned by predictor, which is the check that
names *which end* of the series the variance is wrong at. The PIT plot says the
calibration is wrong; it does not say where.

Worth deciding while writing: whether the example uses `lum31` or a smaller
[...]

PR #372, opened 2026-09-16
Confirm the dispersion sub-model with a posterior check (#367)

beckyfisher:
## What

`vignette("example9")` gains a third worked response: one plate of the `lum31`
copper series at 15 minutes, on which the coefficient of variation a `Gamma`
holds constant is not constant. The section runs the diagnostic that reveals it,
the `disp("power")` sub-model that answers it, and the same diagnostics on the
refitted model.

`vignette("example8")` establishes that sub-model on this arm from a table of
within-cell coefficients of variation and stops there. Its *Non-constant
dispersion* section now forward-references the posterior check instead of the
issue number.

## Why it matters

A reader who accepts `disp("power")` on the strength of a table of observed
coefficients of variation cannot repeat that check on their own fit, because the
table summarises the raw response rather than the model. The check that does
transfer is a posterior one, and `example9` is where the diagnostics live.

## Before reviewing

This targets **`issue-6-33-grouping-vignette` (PR #228), not `dev`**, and cannot
merge until #228 does. `example9` is on `dev`, but `lum31`, `example8` and the
exploratory scripts the issue cites are only on #228's branch.

The rendered `vignettes/example9.Rmd` is stale and is not rebuilt here, per the
note in `.github/workflows/precompile-vignettes.yaml`. Every chunk was knitted
against the two saved fits and every inline value checked against that output.
The section adds about 20 minutes to the `example9` render.

## Evidence

Measured on one plate of the `lum31` copper series at 15 minutes, 44 wells over
11 concentrations by four replicate wells, fitted with the `decline` set at
`bnec()`'s sampling defaults, `seed = 17`, `cmdstanr` 0.9.0 against CmdStan
2.39.0, `brms` 2.23.0, R 4.6.1. `ecx4param` holds the most weight in both sets.

| | constant dispersion | `disp("power")` |
|---|---|---|
| `check_fit()` groups flagged on spread, of 11 | 7 | **0** |
| `sd_ratio` at the control, 1 expected | 0.09 | 1.02 |
| span of `sd_ratio` over the eleven groups | 0.05 to 2.80 | 0.49 to 2.11 |
| PIT values in the middle half of (0, 1), 0.5 expected | 0.721 | 0.535 |
| control-mean posterior SD (RLU) | 154,859 | 13,653 |
| *NSEC* reference, as a share of the control mean | 0.70 | 0.97 |
| *NSEC*, model-averaged (mg/L) | 0.109 | 0.054 |
| EC10 95% interval width, model-averaged (mg/L) | 0.122 | 0.011 |

The observed coefficient of variation within a concentration runs from 0.046 at
the control to a maximum of 1.271, which is 28 times the control value, so a
single `shape` cannot serve the series. Three results follow.

The PIT-ECDF plot leaves its simultaneous band under constant dispersion and
stays inside it under the sub-model, so the diagnostic the issue proposed does
separate the two fits. `check_fit()` agrees, going from 7 of 11 groups flagged
to none.

The estimated exponent is 0.78 (0.64 to 0.93). The replicate wells give 0.77
(0.66 to 0.88) for the same quantity with no fit involved, by regressing the
within-concentration `log(cv)` on `log(mu)` at an R-squared of 0.965. No curve
enters a within-concentration standard deviation, so the sub-model is describing
the replicate variation the design measured rather than absorbing a failure of
the curve.

The *NSEC* halves, from 0.109 to 0.054 mg/L. That is the mechanism the
vignette's own Overview asserts, measured rather than inferred: the
control-mean posterior is 11.3 times wider under constant dispersion, so the
reference `nsec()` sets at a lower quantile of it sits at 0.70 of the control
mean against 0.97, and a reference placed that low is crossed further along the
curve. The constant-dispersion fit therefore reports the **less protective** of
the two figures while its sampler diagnostics look clean. EC10 changes by less,
0.087 to 0.069 mg/L, being measured at a fixed fraction of the control mean
rather than at a quantile of its posterior.

Estimates are model-averaged over the screened set, as step 5 of the vignette
does. The constant-dispersion weights are flat --- `ecx4param` 0.254 against
`ecxll4` 0.196, `nec4param` 0.189, `ecxwb1` 0.187, `ecxwb2` 0.173 --- so no one
equation describes the averaged value. The diagnostics are run on the
highest-weighted equation, which is what `pp_check()` and `check_fit()` take.

<details>

### Base branch

This targets `issue-6-33-grouping-vignette` (PR #228), not `dev`. `example9` is
on `dev`, but `lum31`, `example8` and the exploratory scripts the issue cites are
only on #228's branch, so both halves of the issue's definition of done can be
met in one change only from there. It cannot merge until #228 does.

#### The dataset and its scope

The issue left the dataset open. RF settled it: one arm, no grouping term,
copper so that `decline` is the right candidate set, and one plate.

Copper falls from its control. Zinc rises above its control at low
concentrations, which is why `example8` fits `all` there, so a decline equation
is not valid for the zinc arm.

One plate rather than the whole arm, because gain is set per read. Across the 16
copper plates the control mean runs from 234,000 to 1,060,000 RLU, an
across-plate coefficient of variation of 0.34 against a within-cell 0.048.
Ungrouped on the whole arm, the dispersion estimated would be mostly gain. One
plate removes that without needing a group-level term, so nothing is collapsed.
`Cu 19Mar24 Rep1B` is the plate with the fewest readings at the instrument's
floor, one of 44; a reading held at a bound understates exactly the quantity the
section is about.

#### The candidate set

The whole `decline` set is fitted rather than one equation, and screened.
`bnec_record()` reports 3 of the 14 excluded before fitting; `ecxll5` fails to
fit, which on 44 observations is the five-parameter equation; and the screen
removes one more. The screen is run again after the refit and removes a
different equation, because a dispersion sub-model changes the posterior every
equation is sampled from.

#### The displays

`check_fit()` does the work the issue's "residual-against-fitted display binned
by predictor" asks for, and better than a hand-rolled one: on a single plate its
groups **are** the four-well cells, which is the same unit the observed
coefficients of variation are computed on. The third display is its `sd_ratio`
against concentration for both fits on one pair of axes.

#### The censored reading

`brms::pp_check()` reports `Censored responses are not included`, so the
vignette's own `pit()` helper takes the same subset, 43 of the 44 wells.

### Rendered output

Stated above. The 20 minutes is 8.8 for the constant-dispersion set and 11.5 for
the sub-model, measured on 22 cores with `cmdstanr`.

### Corrections to pre-existing Overview text

Two sentences in the Overview are amended, both in text this PR already touches.

"ECx ... read from the fitted curve relative to that curve's own asymptotes"
contradicted step 4 and `NEWS.md`, which record that every ECx has been measured
from the control since 2.2. It now says ECx is measured from the same control
mean, at a fixed fraction of it rather than at a quantile of its posterior,
which is why it responds less.

"One statistic for the whole fit does not detect this" now reads "One **ratio**
statistic". The argument there is about a dispersion ratio, whose free parameter
absorbs the discrepancy. A PIT-ECDF is a whole-fit statistic that is not a ratio
and does detect it, which this section demonstrates, so the sentence as written
was contradicted by the new material.

The Overview's priority claim --- that *NSEC* depends on the variance assumption
most --- needed no change. Copper supports it: the *NSEC* halves while EC10
changes by a sixth.

### Checks

- `prose-check` reports no violations on `example9.Rmd.orig`; the eight hits on
  `example8.Rmd.orig` are pre-existing and outside this edit.
- No duplicate chunk labels; each of the 59 chunks in `example9.Rmd.orig` has a
  distinct one.
- `notes/vignette_numbering.md` and `_pkgdown.yml` are unchanged, since no
  vignette is added.
- `NEWS.md` gains one bullet under *Documentation*.

</details>

Addresses #367. The PR does not target the default branch, so the issue is
closed by hand once this merges.


beckyfisher:
## The review cycle

Four rounds with an independent reviewer. Every finding was re-measured against
the fitted objects before being acted on. The substantive ones follow; the
remainder were render defects and wording, listed at the end.

### The candidate set was wrong for the dataset

The first two rounds reviewed a section that fitted `ecxll5` alone to the `lum31`
zinc arm. `vignette("example8")` fits `all` to that arm because the zinc response
rises above its control at low concentrations, so a decline equation excluded the
hormesis equations a priori. Neither review round asked whether the equation
suited the shape of the response. RF caught it.

The section was rebuilt on one plate of the copper series, which falls from its
control, so `decline` is the right set. That change also dissolved two problems
the earlier rounds had spent their effort on, below.

### Claims the fits did not support

#### The direction of the error

The section said a constant
`shape` simulated too little spread at the bottom of the series. Its own table
showed the implied coefficient of variation exceeding the observed one at every
concentration. The pooled residual spread did open at the bottom, but 1.59 of
1.72 was between cells rather than within them --- one curve and one plate factor
failing to reach each plate, which is lack of fit in the mean, the second of the
three explanations `vignette("example1")` lists for excluding before a variance
function is added.

#### The censoring argument

The section attributed part of the
sub-model's overshoot at the bottom to readings held at the instrument's floor,
on the reasoning that a coefficient of variation computed from bounded values
understates the spread. Restricting to the cells with no censored well *lowered*
the observed value, from 0.361 to 0.309, widening the gap rather than narrowing
it. The claim was removed.

#### The *NSEC* explanation

The section said the sub-model had revised the control-mean posterior. `nsec()` reads that
posterior with `re_formula = NA`, and its coefficient of variation was 0.112
under both fits: unchanged. On the copper plate, where no group-level term
contributes to it, the mechanism is real and is now measured rather than
asserted --- posterior SD 154,859 against 13,653, and the reference at 0.70 of
the control mean against 0.97.

### Estimates read from a single equation

The estimates came from `pull_best()`, whose own documentation warns that on a
flat set one equation "describes the model-averaged estimate hardly at all", and
step 4 of this vignette repeats the warning. The constant-dispersion weights are
flat: `ecx4param` 0.254, `ecxll4` 0.196, `nec4param` 0.189, `ecxwb1` 0.187,
`ecxwb2` 0.173. Step 5 of the same vignette reports the model average.

Correcting it strengthened the result. The *NSEC* falls by a factor of 2.0 rather
than 1.9, and the EC10 interval narrows elevenfold rather than fivefold.

### Displays printed and not read

Three rounds each found a display whose numbers the prose did not state, or
stated wrongly. A maximum was described as lying at the bottom of the series when
it sat one concentration short of it, three times in three different drafts. A
comparison table of 22 rows and four numeric columns had three cells read. A
residual figure had a y-range set by a dozen outliers, so none of the four values
the prose quoted could be read off it. The `disp("power")` ratios were described
as sitting on the line of agreement while three of the eleven sat at about 0.49,
which on a logarithmic axis is as far from agreement as 2.06.

### Corrections to pre-existing Overview text

The new section is the first place in the vignette where either could be tested.

"ECx ... read from the fitted curve relative to that curve's own asymptotes"
contradicted step 4 and `NEWS.md`, both of which record that every ECx has been
measured from the control since 2.2.

"One statistic for the whole fit does not detect this" now reads "One **ratio**
statistic". The argument there is about a dispersion ratio, whose free parameter
absorbs the discrepancy; a PIT-ECDF is a whole-fit statistic that is not a ratio
and does detect it, which this section demonstrates.

### The recurring fault

One pattern produced findings in three successive rounds: a mechanism asserted in
prose without the measurement that would settle it. It appeared as the censoring
argument, then in the sentence written to replace it, then in a ratio substituted
for the quantity it was meant to describe. Each replacement was plausible and
each was checkable in a few lines. The rule the section now follows is that a
causal clause either states its measurement or comes out.

### The remainder

Render defects: an inline `paste()` that produced "at ranks 11"; an interval bound
rendering as 1.58 × 10^4, because `knitr` switches to scientific notation at
10,000; `rev()` on a `confint()` row, which reverses the values but not their
names, so the printed bounds were labelled backwards in the one chunk asking the
reader to compare two intervals. Wording: a plate described as the best when two
tied; two different overstatement ratios quoted four paragraphs apart without
saying they were different estimators; an unmeasured causal clause about why the
model weights concentrate, replaced by the counts.

### Deliberate omissions

`check_sampling()` is shown on the first fit and not the refit. The claim the
section needs is that the sampler is not the problem, and running the same
display twice on a one-equation-per-set comparison repeats what step 3 teaches.
Both fits are clean: maximum R-hat 1.005, minimum bulk ESS 1226, no divergent
transitions.

The sub-model leaves `sd_ratio` spanning 0.49 to 2.11 rather than sitting at 1.
No group is flagged, and the reason is the design: a standard deviation computed
from four values has a sampling standard deviation of about 41 per cent of
itself. The section reads the display as showing that the ratios no longer trend
with concentration, rather than that each is exactly 1.

The model weights are reproducible given a seed but not across seeds, because
`loo_model_weights(method = "stacking")` optimises from a random start and the
draw index is drawn from the ambient random state (#216). Within a render the
printed table and every inline recomputation read the same stored index and
agree. Across renders the third decimals differ, and `ecxwb1` and `ecxwb2` can
swap rank. Everything the prose leans on is stable across seeds: five equations
hold more than one per cent of the weight in every state, `ecx4param` is highest
in every state, and the factor of 2.0 on the *NSEC* does not change.

PR #375, opened 2026-09-16
Map fitted groups to colour in autoplot()

### LINKED ISSUE #374: Map a grouping to colour in autoplot(), where the level count allows it
## What

`autoplot()`'s `group` argument gains a way to map the grouping to colour, for
the level counts where colour separates them.

## Why it matters

#371 distinguishes the levels of a grouping by drawing a grey line through the
levels that span more than one concentration and a white diamond at the mean of
those that sit at one. That reads well at seventeen plates, where seventeen
colours would not.

It does not answer the case with few levels, where the question is which level
a point belongs to rather than whether the levels are parallel. Three climate
scenarios in `coral_pam`, two toxicants in `lum31`, or seven herbicides are all
counts a categorical colour scale handles, and a grey line through three
scenarios tells a reader they differ without telling them which is which.

## Evidence

`vignettes/example8.Rmd.orig` draws its three raw-data figures by hand rather
than through `autoplot()`, and the coral one maps `climate` to colour and shape
to do exactly this. That is a `ggplot()` call on the data frame, so it is
available before a fit and not after one: the same figure of a *fitted* object
cannot currently be drawn.

The counts the vignette would use it on are 3 (`climate`), 2 (`toxicant`), 7
(herbicides), 17 (`plate`), 45 and 54 (`chamber`), and 363 (`conc_group`). The
first three are colour's range and the last three are not, so whatever is added
has to leave #371's behaviour reachable rather than replace it.

<details>

`ggbnec()` takes `group` as a logical and branches on it at `R/autoplot.R:522`,
choosing `stat_summary()` with `geom = "line"` for levels spanning more than one
predictor value and `geom = "point"` with `shape = 23` for the rest. The colour
is hard-coded `grey40` and the fill `white`.

Three decisions to make before writing code.

#### The caller's side

`group = "plate"` currently means "distinguish the levels of `plate`". Adding a
second argument, say `group_aes = c("auto", "line", "colour")`, keeps the
existing call working and names the choice. An `"auto"` that picks by level count
is one option, and if it is taken the rule should be stated in the
documentation: a figure that changes form when a dataset gains a level is a
figure a reader cannot rely on.

#### The marks that take the colour

The observations are drawn once, at `R/autoplot.R:552`, as `shape = 21` with
`fill = "grey30"`. Mapping `fill` to the grouping colours the points. Whether
the per-level line or diamond should also take it, or stay grey so that the
population curve and the levels remain distinguishable, is the display question.

#### The scale

`bayesnec` currently imports no palette beyond what `ggplot2` provides.
`scale_fill_viridis_d()` and the default hue scale are both available without a
new dependency. Leaving the scale off the plot entirely, so the caller adds
their own, is the most flexible and the least helpful by default.

Definition of done: a fitted object with a three-level grouping can be drawn
with the levels in colour; #371's line and diamond remain reachable and stay the
default above whatever count is chosen; `?autoplot.bayesnecfit` documents the
argument; `tests/testthat/test-autoplot.R` covers both forms; and the
`coral_pam` figure in `vignette("example8")` can be drawn from the fit rather
than by hand.

Stop and ask rather than guess about `"auto"`. Choosing the form from the data
is a decision about how much the package should do silently, and it is the kind
of thing that is easier to add later than to remove.

</details>

beckyfisher:
## Group colours

`autoplot()` now accepts `group_aes = "colour"` when `group` selects a
group-level variable. Observations and their per-level mean marks use a
categorical colour scale, and the legend names the source variable.

The existing grey mean-line and mean-diamond display remains the default with
`group_aes = "line"`.

## Reason

The grey display shows between-level variation when a fit contains many levels,
but it cannot identify individual levels. Colour makes a small grouping such as
the three climate scenarios in `coral_pam` identifiable without removing the
display suited to the 17 plates in `lum31`.

No automatic level-count threshold is introduced. The caller chooses the
display explicitly, so adding a level to the data does not silently change the
figure.

## Verification

The plotting tests pass 61 assertions against both the source and an installed
package. The neighbouring installed-package tests for single fits,
model-averaged fits and class expansion pass another 162 assertions. A package
check passes loading, S3 consistency, documentation, usage and examples; its
remaining findings are the existing `modifyList` note, the worktree `.git`
note, and the expected missing-`inst/doc` warnings from
`--no-build-vignettes`.

<details>

- `group_aes = "colour"` is available on the `bayesnecfit` and
  `bayesmanecfit` methods, including the model-average and individual-model
  branches.
- The selected group maps to point fill, mean-line colour and mean-diamond
  fill. The fitted population curve remains black.
- ggplot2 supplies the discrete scales, so callers can replace them without a
  new package dependency.
- Tests cover the unchanged default, both kinds of per-level mean mark,
  model-averaged propagation, mixed line-and-diamond groupings, consistent
  colour scales, legend labels and invalid values.

</details>

Closes #374


beckyfisher:
Independent review completed after two cycles. One real problem was found and
fixed: mixed group designs trained fill and colour on different subsets, so a
level's observations and mean line could receive different hues and appear in
two legends. Commit `b2a5b3fd` gives both scales the complete level set and adds
a regression test for identical mappings, complete breaks and one merged guide.
No other defects were found.

beckyfisher:
Using this in `vignette("example8")` turned up a case the reason section names
and the implementation refuses.

```r
autoplot(fit_pam, group = "climate", group_aes = "colour")
#> Error: `group` must name a group-level variable in the fitted formula: "chamber".
```

`coral_pam` is fitted with `ogl(chamber)`, so `chamber` is the only value
`group` accepts, and the three climate scenarios cited above as the motivating
example cannot be drawn. The restriction may well be deliberate: colouring by a
covariate the model holds no parameter for invites a reader to see a fitted
difference that was never estimated, and that is a good reason to refuse it.
Recording the case rather than proposing a change.

What the level counts leave, in that vignette: `plate` has 17 levels and
`chamber` 45 and 54, which a categorical scale cannot separate, and
`conc_group` has 11 levels each sitting at a single concentration, so colour
repeats what the position on the axis already gives and spends about 40 per cent
of the figure width on a legend of names like `Zn 23Apr24 Rep6B-c01`. The
vignette therefore uses `group` with the default marks and says in the prose why
`group_aes = "colour"` is not used, so that a reader who wants it knows it is
there.

The default display earns its place in the same vignette. `fig-plate` draws
seventeen plate traces and makes the between-plate spread legible, which the
section had been asserting in prose and not showing.

PR #377, opened 2026-09-16
Support dispersion for model sets

### LINKED ISSUE #373: dispersion() fails with an internal terms.formula() error on a bayesmanecfit
## What

`dispersion()` on a `bayesmanecfit` fails with an internal error from
`terms.formula()` rather than either returning the per-equation values or
refusing the class by name.

## Why it matters

The per-equation dispersion values already exist on the object. `expand_nec()`
calls `dispersion()` once per equation while the set is being built, and
`expand_manec()` assembles the results into the `dispersion_Estimate`,
`dispersion_Q2.5`, `dispersion_Q97.5` and `dispersion_P_over_1` columns that
`summary()` reports. A user who reads those columns in a summary and then calls
`dispersion()` on the same object to get the draws behind them meets an error
that names neither the object nor the reason.

The error message points at `terms.formula`, which appears nowhere in the user's
call, so it reads as a malformed formula rather than as an unsupported class.

`?dispersion` documents `model` as "An object of class `bayesnecfit`", so a
`bayesmanecfit` is outside the documented scope. This is a report about the
failure mode rather than about missing behaviour, though supporting the class
looks cheap given the values are already computed.

## Evidence

`bayesnec` 2.1.3.35, R 4.6.1, `brms` 2.23.0, `cmdstanr` backend, WSL2 Debian.

```r
library(bayesnec)

set.seed(333)
x  <- rep(c(0.1, 0.3, 1, 3, 10, 30), each = 8)
mu <- 0.9 * exp(-0.8 * pmax(x - 1, 0))
d  <- data.frame(x = x, tot = 20L, suc = rbinom(length(mu), 20, mu))

f1 <- bnec(suc | trials(tot) ~ crf(x, model = "nec3param"), data = d,
           family = binomial(link = "identity"), iter = 2000, seed = 333)
f2 <- bnec(suc | trials(tot) ~ crf(x, model = "ecxexp"), data = d,
           family = binomial(link = "identity"), iter = 2000, seed = 333)

dispersion(f1, summary = TRUE)
#>  Estimate      Q2.5     Q97.5     P(>1)
#> 0.9804792 0.5375373 1.7478920 0.4793750

manec <- c(f1, f2)
class(manec)
#> [1] "bayesmanecfit" "bnecfit"

dispersion(manec, summary = TRUE)
#> Error in terms.formula(formula, data = data) :
#>   argument is not a valid model
```

The family is `binomial` here deliberately, so the statistic is defined and the
failure is not a consequence of an unsupported family. The same error occurs
under `Beta`, where a single fit correctly returns `numeric(0)`.

<details>

### Cause

`dispersion()` in `R/dispersion.R` handles `bayesnechurdlefit`
explicitly and then assumes a `bayesnecfit`:

```r
formula <- model$bayesnecformula
model <- model$fit
mod_dat <- model.frame(formula, data = model$data)
```

Both fields are `NULL` on a `bayesmanecfit`, so `model.frame(NULL, data = NULL)`
reaches `terms.formula()` with a `NULL` formula.

```r
is.null(manec$bayesnecformula)   # TRUE
is.null(manec$fit)               # TRUE
```

### Possible resolutions

Refuse by class, alongside the existing `is_bayesnechurdlefit()` branch, with a
message naming `bayesmanecfit` and pointing at `pull_out()` or `pull_best()` for
a single equation and at `summary()` for the per-equation values.

Or support it, by mapping over `mod_fits` in the way the two-block branch already
maps over `growth` and `survival`, and returning a named list. That matches what
`expand_nec()` already does per equation, and would give a public route to the
draws behind the `dispersion_*` summary columns rather than only their summaries.

If the second is taken, the `@param model` line in the roxygen block needs
widening from `bayesnecfit`, and `summary = TRUE` should keep returning one row
per equation so the result lines up with the summary table.

### Scope

Not family-specific and not data-specific: it is reached by any
`bayesmanecfit`, whether assembled by `bnec()` with a model set or by `c()` on
single fits as above.

### Provenance

Found while revising the AIMS concentration-response training material against
the `dev` vignettes, in the module that introduces `dispersion()` as an
over-dispersion screen.

</details>

beckyfisher:
## What

`dispersion()` now supports a `bayesmanecfit`. It returns a named list of per-equation posterior draws by default, or a data frame with one row per equation when `summary = TRUE`.

## Why it matters

Users can inspect the draws behind the per-equation dispersion values already reported by `summary()`. The call no longer fails inside `terms.formula()` with an error unrelated to the supplied object.

## Evidence

Regression tests cover count-family draws and summaries, equation names, and unsupported-family results. The focused dispersion tests pass against R 4.6.1, and an installed source build passes a model-set smoke test.

Closes #373.

<details>

- Dispatch for `bayesmanecfit` objects maps the existing single-equation calculation over `mod_fits`.
- Empty per-equation summaries become `NA` rows, consistent with the existing model summary.
- Argument validation now precedes class dispatch, so all supported container classes apply the same checks.
- Roxygen documentation describes the return type for single fits and model sets.
- `R CMD check --no-manual --no-build-vignettes --no-tests` completed with two pre-existing notes and two warnings caused by the disabled vignette build and existing repository conditions.

</details>


beckyfisher:
## Independent review outcome

The independent review completed after one cycle. It found no genuine problem in the original PR, so no review-driven fix was required. The reviewer then assessed the developer response and agreed that the PR was sound.

All 49 focused dispersion tests and all five GitHub checks passed.

PR #379, opened 2026-09-16
Allow autoplot groups from supplied data

### LINKED ISSUE #378: autoplot(): allow group to name any variable in the supplied data
## What

`autoplot()` and `ggbnec_data()` accept `group` only where it names a variable
the fitted formula gives to `ogl()` or `pgl()`. Expand it to accept any variable
of the data frame supplied to `bnec()`, so that a covariate the model pools over
can still be marked on the figure.

## Why it matters

The case #374 opened with cannot be drawn. `coral_pam` holds 414 measurements of
photosynthetic yield over 54 chambers and three climate scenarios. Fitted as
`yield ~ crf(log(diuron_adj), "all") + ogl(chamber)`, the model estimates one
deviation per chamber, so `group = "chamber"` is accepted and draws 54 levels.
The question these data were collected to answer is which scenario each point
came from, and `group = "climate"` is refused. Three colours would answer it;
54 answer nothing.

The same shape recurs wherever a design records a factor the model pools over:
a replicate block, an assay date, a source population. A reader who wants that
factor shown has to redraw the figure by hand from the data frame, which shows
neither the fitted curve nor its credible band.

## Evidence

Run on the `dev` merge at `c9f204d3`, bayesnec 2.1.3.38, R 4.6.1, on the shipped
`coral_pam` with the zero control shifted to 0.1:

```r
fit <- bnec(yield ~ crf(log(diuron_adj), "nec3param") + ogl(chamber),
            data = d, family = Beta(link = "identity"), seed = 228)

autoplot(fit, group = "chamber", group_aes = "colour")   # draws 54 levels
autoplot(fit, group = "climate", group_aes = "colour")
#> Error : `group` must name a group-level variable in the fitted formula: "chamber".
```

`climate` is a column of `coral_pam` with three levels. It is refused because the
formula does not name it, not because the value is unavailable.

The refusal is raised in `prep_raw_data()`, `R/autoplot.R:203-212`, against
`attr(mod_dat, "bnec_group")`.

There is a reason behind the current restriction, recorded when #374 closed:
colouring by a covariate the model holds no parameter for invites a reader to
see a fitted difference that was never estimated. That argues for marking the
distinction in the figure, not for withholding the display, because the
alternative a reader takes instead is a hand-drawn plot where nothing marks it.

<details>

### The source of the variable

`prep_raw_data()` reads `brms_fit$data`, and `brm()` retains only the model
frame. `manec_example`'s fit keeps `y` and `x`; the `coral_pam` fit above keeps
`yield`, `diuron_adj` and `chamber`, and not `climate` or `fragment`. Relaxing
the check alone therefore fails on the motivating case. Three routes give the
value a source:

1. **Store the unused columns on the fit.** `bnec()` has the user's data frame.
   The columns the model frame does not use are retained on the returned object
   and `prep_raw_data()` looks there when the name is not in `brms_fit$data`.
   Row alignment is exact, because the same rows are dropped. Every fitted object
   grows by those columns whether or not the figure is ever drawn, and they
   would need excluding from the digest that keys a stored fit.
2. **Accept a vector.** `group` takes either a name or a vector of length
   `nrow(brms_fit$data)`. Nothing is stored, and the caller has to align it,
   which is unsafe wherever `na.omit()` or the family's zero
   shift dropped rows.
3. **Add a `data` argument.** `autoplot(fit, group = "climate", data = coral_pam)`
   resolves the name against the frame the caller supplies, and the rows are
   matched on the model frame's row names rather than on position.

Route 3 is the smallest change to the fitted object and the safest alignment.
Route 1 gives the shortest call. Route 2 is not recommended on its own.

### The label on an unfitted grouping

Whichever route is taken, the legend must not read as a fitted split. The
existing `group_label` is the place: where the name came from outside the
formula, the label says so, and the figure's own caption in the documentation
states that the curve and band are pooled over that variable.

### Edge cases

- A continuous variable, which needs either a continuous colour scale or a
  refusal naming the reason.
- A variable with more levels than a legend can show; `conc_group` in
  `example8` has 363.
- `group_aes = "line"` on a variable with no fitted deviation, which would draw
  one line per level from a single set of parameters and should be refused.
- A `bayesnecgroupfit`, where each level was fitted from its own subset, so the
  supplied frame has to be subset per level before it is bound.

### Tests

`tests/testthat/test-autoplot.R`: a name present in the supplied data and absent
from the formula returns a `group` column with the right levels; a name absent
from both errors and names both sources; row counts match the model frame after
rows are dropped; a `bayesnecgroupfit` resolves per level.

### Definition of done

`autoplot(fit_pam, group = "climate", group_aes = "colour")` draws three
colours over the pooled fit, and the `example8` coral figures use it in place of
the hand-drawn raw-data plot.

### The open decision

Which route to take. Route 1 changes the size of every fitted object and the
content of the data digest, which is not a decision to make inside this issue.

</details>

beckyfisher:
## What

- Retain columns omitted from the `brms` model frame once on each returned fit, in fitted-row order.
- Let `ggbnec_data()` and `autoplot()` colour observations by any retained categorical column.
- Mark external groupings as “not fitted”, keep the fitted curve and band pooled, and refuse `group_aes = "line"` for them.
- Preserve retained columns through `pull_out()`, `amend()`, `update()` and compatible `c()` operations, including grouped fits.

## Why it matters

This makes the motivating `group = "climate"` plot self-contained and safe for archived or derived fits. It avoids caller-managed row alignment (route 2) and the requirement to keep and resupply the original data frame (route 3), while limiting route 1's size overhead by storing omitted columns once per constituent fit rather than in every model component.

## Evidence

- `NOT_CRAN=true Rscript -e 'devtools::test(filter = "autoplot", stop_on_failure = TRUE)'`: 85 assertions passed.
- Installed-package check verified retained grouping, row alignment, legend labelling, and no unfitted per-group summary lines.
- A real installed-package `bnec()` fit retained and plotted an unused `climate` factor.
- `R CMD check --no-manual --no-build-vignettes --no-tests`: package syntax, loading, S3 registration, documentation and examples passed. The two warnings and two notes are the repository's existing worktree, `modifyList`, and omitted-vignette findings.

<details>

`bnec()` compares the supplied columns with the stored `brmsfit$data` and attaches only omitted columns to the completed bayesnec object. Rows are matched by row name while both frames are available. Plotting can therefore resolve an external grouping without accepting an unaligned vector or depending on a later copy of the source data.

External groupings must be factor, character, or logical and contain no missing values. The plot colours raw observations only; it does not add per-level fitted summaries or imply per-level curves. Existing fitted `ogl()`/`pgl()` grouping behaviour is unchanged.

Tests also cover model-averaged fits, independently fitted outer panels, missing and continuous group variables, and preservation through derived-fit operations.

</details>

Closes #378


beckyfisher:
Independent review completed in two cycles. One genuine problem was found in the original PR:

- The exported `group_fitted` documentation defined `TRUE` only for an `ogl()` or `pgl()` formula term, but independently fitted `bnec_group()` panels also return `TRUE`. Commit c94af9c1 corrected the public contract and added a regression assertion for the grouped-fit case.

The PR body was also updated to state the grouped-fit storage scope precisely and to record the 85 passing focused assertions. The reviewer rechecked the changes and response and considers the PR sound, with no unresolved findings.

PR #383, opened 2026-09-17
Support sampler screening for grouped fits

### LINKED ISSUE #376: check_sampling() and screen_models() do not accept a bayesnecgroupfit
## What

`check_sampling()` and `screen_models()` gain a `bayesnecgroupfit` method, so
that a grouped fit can be screened the way a single fit or a model-averaged set
can.

## Why it matters

`screen_models()` is how a reader is told to drop equations whose chains did not
explore the posterior, and `vignette("example9")` sets out the cutoffs and the
reasoning. Neither function accepts a `bayesnecgroupfit`, so the one kind of fit
that holds several model averages at once cannot be screened at all:

```
screen_models(fits_tox)
#> Error: check_sampling requires a bayesnecfit, a bayesmanecfit or a
#> bayesnechurdlefit.
```

A `bnec_group()` call fits a complete model set per level, so it is the call
most likely to contain an equation that sampled badly at one level and well at
another. It is also the call whose output is hardest to inspect by hand, because
`summary()` on seven levels is several hundred lines.

## Evidence

In `vignette("example8")`, `fits_herb` holds seven levels of eleven equations
and `fits_tox` two levels of fourteen. The five single-object fits in the same
vignette are screened; these two cannot be, so the vignette has to state the
inconsistency rather than resolve it.

Measured on the `lum31` fits of that vignette, screening is not cosmetic. On
`fit_pooled`, 0.617 of the weight sits on equations that fail the screen, and
dropping them changes EC10 from 0.328 to 0.279 mg/L and the N(S)EC from 0.197 to
0.162. On `fit_ogl` the median moves little and the lower bound of the EC10
interval changes from 0.015 to 0.152, because the equation dropped is unbounded
below the observed range.

<details>

Both functions dispatch on class and stop in `check_sampling()`. A
`bayesnecgroupfit` holds `$fits`, a list of `bayesnecfit` or `bayesmanecfit`
objects, and `$levels`, so the natural form follows the existing
`group_lapply()` and `group_estimate_table()` helpers in
`R/bayesnecgroupfit-class.R`.

Two decisions to make.

#### The shape of the returned table

`check_sampling()` returns one row per equation. For a grouped fit the level has
to appear as well, and whether that is a `level` column on the same table or a
list of tables per level decides how the result reads for seven levels of eleven
equations, which is 77 rows.

#### The meaning of a screened grouped fit

Dropping an equation at one level and not another leaves the levels with
different candidate sets. That is defensible, since `bnec_group()` exists so
that levels may differ, but it makes the model averages less comparable and a
reader should be told. The alternative, dropping an equation everywhere if it
fails anywhere, keeps the sets aligned and discards good fits. Neither is
obviously right and the choice should be an argument rather than a default
nobody sees.

Definition of done: `check_sampling()` and `screen_models()` accept a
`bayesnecgroupfit`; `?check_sampling` and `?screen_models` state what screening
a grouped fit does about levels that disagree;
`tests/testthat/test-check_sampling.R` covers a grouped fit where one level
fails and another does not; and `vignette("example8")` screens its two
`bnec_group()` fits alongside the other five.

Stop and ask rather than guess about the second decision. It is a statement
about what a model average over a screened set means, not a convenience.

</details>

beckyfisher:
## What

`check_sampling()` now accepts grouped fits and returns one diagnostics table
with a `level` column. `screen_models()` screens each independently fitted level
independently by default, with an explicit `group_action = "common"` option for
analyses that require a shared candidate set.

The two grouped fits in `vignette("example8")` now apply per-level screening
before their estimates are reported.

## Why it matters

A failed equation can now be identified and removed from the level where it
failed without discarding a successful fit of the same equation at other
levels. The returned grouped object remains compatible with the existing
grouped summaries and estimators.

## Evidence

The focused test file passed 70 assertions through `devtools::test()`. Before
the review fix added the final assertion, the same file passed against an
installed package build. A source-package check excluding the full test suite,
manual and vignette build completed without errors; its warnings and notes were
the existing consequences of skipping vignette construction and the existing
`modifyList` import diagnostic.

<details>

- Grouped diagnostics concatenate per-level results into a single table.
- `group_action = "per_level"` preserves the independent meaning of
  `bnec_group()` fits and is the default selected for this issue.
- `group_action = "common"` removes an equation everywhere when it fails at any
  level, and stops when removal would leave no candidate equations for a level.
- Regression tests cover the same equation passing at one level and failing at
  another, both screening policies, the all-removed guard, and N(S)EC comparison
  after levels reduce to single ECX equations.

</details>

Closes #376


beckyfisher:
## Review resolution

The independent review completed after three cycles. Two findings identified
real problems and were fixed:

- The grouped vignette comparisons still requested the literal `nec` parameter.
  Per-level screening can reduce a level to a single ECX equation, which made
  that comparison fail. Both calls now use the N(S)EC comparison path, and a
  regression test covers grouped levels reduced to single ECX equations.
- The PR body retained the earlier count of 69 assertions after that regression
  raised the current focused result to 70. The evidence and coverage summary
  now describe the current branch.

The reviewer verified both responses and found no remaining defects or optional
changes.

PR #384, opened 2026-09-17
Document prior scope for extended formulas (#381)

### LINKED ISSUE #381: Priors name a parameter and no coefficient, so reuse on an extended formula reaches the contrasts
## What

The priors `bayesnec` builds name a non-linear parameter and no coefficient, so
each applies to every population-level coefficient of that parameter. Within
`bayesnec` that is always the intercept alone. Reused on a formula that gives
the parameter a predictor, the same prior lands on the contrasts as well.

## Why it matters

Reading the priors off a `bnec()` fit and reusing them in `brms` is the route
the AIMS training material teaches for a factor interaction, in module 8 of
`cr_modelling_training`: the equation and the priors are taken from the fitted
object so that the interaction model differs from it only in the right-hand side
of each parameter line. The priors do not survive that change of right-hand
side intact.

Where the reused prior has unbounded support the effect is a shifted centre.
Where it does not, it is a hard constraint. `ecxwb1` takes `beta(5, 2)` on `top`
and `beta(2, 5)` on `bot`, and a beta density is zero outside (0, 1), so both
contrasts are confined to (0, 1) and cannot be negative: the interaction model
cannot represent a level whose `top` or `bot` is lower than the reference
level's.

## Evidence

Measured on 2026-09-17, bayesnec 2.1.3.39, brms 2.23.0, on the zinc data that
module uses. `get_priors()` returns four priors, each with an empty `coef`:
`beta(5, 2)` on `top`, `beta(2, 5)` on `bot`, `normal(0, 5)` on `beta` and
`normal(2.387, 24.803)` on `ec50`.

Passed to `brms::make_stancode()` for the interaction formula, which needs no
sampling:

```
lprior += beta_lpdf(b_top | 5, 2);
lprior += beta_lpdf(b_bot | 2, 5);
```

`b_top` and `b_bot` are length 3 in that program — `Intercept`, `hardness30`
and `hardness406`. `brms` warns while building it:

```
It appears as if you have specified an upper bounded prior on a parameter that
has no natural upper bound.
```

Stamping `coef = "Intercept"` is inert for a parameter fitted as `~ 1`. The two
Stan programs differ in one place:

```
-  lprior += beta_lpdf(b_top | 5, 2);
+  lprior += beta_lpdf(b_top[1] | 5, 2);
```

`b_top` is length one there, so the two are the same model. On the interaction
formula the same stamp confines the prior to the intercept and leaves the
contrasts on the `brms` default.

There is an argument against changing anything: a package is not obliged to make
its output safe for a model it did not fit, and the reuse above is outside what
`bnec()` supports. Against that, the workflow is one this project's own training
material recommends, and the failure is silent.

<details>

### The construction of the priors

`R/define_prior.R` constructs them with `prior_string(..., nlpar = "top")` and
so on, leaving `class` at its default of `"b"` and `coef` empty. Every
right-hand side `bayesnec` generates is `~ 1`, and a group-level term enters as
a random effect on a parameter of its own:

```
y ~ ogl + top * exp(-exp(beta) * (x - nec) * step(x - nec))
top ~ 1
beta ~ 1
nec ~ 1
ogl ~ 1 + (1 | grp)
```

The grouping is class `sd`, which these priors do not touch, so the class `b`
vector is length one for every parameter in every formula the package builds.
The behaviour is correct as it stands; what is at issue is what the output means
once it leaves.

### The change

Add `coef = "Intercept"` to the `prior_string()` calls for the population-level
parameter priors. The generated Stan program then changes text, so a compilation
cache key changes and any test comparing `make_stancode()` output needs
updating, and `get_priors()` output printed in the vignettes gains a value in a
column that is currently blank.

### The alternative

Leave the behaviour and document it: a note in `?get_priors` and `?pull_prior`
saying the priors apply to every population-level coefficient of the parameter,
and what that means if the formula is later extended. That changes no code, and
it reaches only a reader who consults the page before running.

### The open decision

Which of those two, or neither.

</details>

beckyfisher:
## What

The documentation for `get_priors()` and `pull_prior()` now states that their
output is specific to the intercept-only non-linear parameter formulas generated
by `bnec()`. Before reusing a prior in a direct `brm()` call, the added
coefficients must be checked against the meaning and support of the parameter.

## Why it matters

A parameter-level prior applies to every population-level coefficient after a
right-hand side such as `top ~ 1` is extended to `top ~ treatment`. The prior
therefore reaches the treatment contrasts as well as the intercept. For a beta
prior, that silently prevents a negative contrast. A no-intercept cell-means
formula such as `top ~ 0 + treatment` differs: each coefficient is a
treatment-specific `top`, so applying the same bounded prior to each can be
appropriate.

Changing the prior to `coef = "Intercept"` does not provide a general package
fix. In brms 2.23.0, a prior cannot specify both `coef` and `lb` or `ub`, while
those bounds are part of several priors generated by `bayesnec`. The
documentation now directs the user to inspect the target formula with
`brms::get_prior()` and rebuild priors when the added coefficients have
different meaning or support.

Closes #381.

<details>

The change covers both routes by which a user can obtain the prior rows:

- `?get_priors` records the intercept-only invariant, the effect of extending a
  non-linear parameter formula, the valid no-intercept cell-means case, the brms
  boundary restriction, and when reconstruction against the target formula is
  required.
- `?pull_prior` states that its complete stored prior table has the same
  restriction and refers to `?get_priors` for the full explanation.
- `NEWS.md` records the user-facing clarification.

The package's prior construction and fitted models are unchanged. This keeps
the bounds and the prior round trip through `bnec(prior = get_priors(fit))`
intact.

Validation used R 4.6.1 and brms 2.23.0:

- the full `test-get_priors.R` file with `NOT_CRAN=true`: 71 passed, no skips,
  and one existing response-direction warning;
- source-package check without tests, the manual, or vignette rebuilding:
  examples and all Rd checks passed. The result retained the existing vignette
  warnings and the existing `modifyList` note; the temporary worktree also
  produced a note for its `.git` file.

</details>


beckyfisher:
The independent review cycle is complete, and the PR body has been updated to
describe the current change.

One review comment identified a real problem in the original PR: the guidance
said every extended parameter formula required rebuilt priors. That excluded a
valid no-intercept cell-means formula such as `top ~ 0 + treatment`, where each
coefficient is a treatment-specific `top` and can correctly share the bounded
prior. Commit `29b440ba` now limits the warning to coefficients whose meaning or
support differs, including ordinary treatment contrasts and slopes. The
reviewer checked the response and revised diff and found no further actionable
issues.

PR #385, opened 2026-09-17
Preserve retained data through expand_manec

### LINKED ISSUE #380: expand_manec() does not carry retained_data, so a reassembled set cannot colour by a retained column
## What

`expand_manec()` returns a model set with no `retained_data` element. A caller
that reassembles a set from separately computed fits is therefore left without
the columns #379 retains, and `autoplot()` cannot colour by a variable the model does not
fit.

## Why it matters

#379 keeps the columns omitted from the `brms` model frame so that
`autoplot(fit, group = "climate", group_aes = "colour")` can resolve them. That
works for a set `bnec()` produced, because `bnec()` attaches `retained_data` to
the finished object after `expand_manec()` has returned. It does not work for a
set assembled any other way. `c.bnecfit()` copies the columns onto its result
under a rule of its own; `expand_manec()` is exported and has none, so the feature is available
through one route into a model set and not the other.

## Evidence

The `example8` fits are computed as one array task per equation in
[open-AIMS/grouping-structures](https://github.com/open-AIMS/grouping-structures)
and reassembled with `expand_manec()`. Measured on 2026-09-17 against bayesnec
2.1.3.39, on the `coral_pam` call
`bnec(yield ~ crf(log(diuron_adj), "all") + ogl(chamber), data = coral_pam, ...)`:

Each unit holds the columns.

```r
u <- readRDS("units-379/c8a4096e2556e8cc__nec3param.rds")
names(u$fit$retained_data)   # "climate" "diuron" "fragment"
nrow(u$fit$retained_data)    # 414
```

The set reassembled from those units does not.

```r
f <- readRDS("store-379/c8a4096e2556e8cc.rds")
f$retained_data              # NULL
autoplot(f, group = "climate", group_aes = "colour")
#> Error : `group` must name a group-level variable in the fitted formula or a
#> column retained from the data supplied to bnec(). Available columns are:
#> "yield", "chamber", "diuron_adj".
```

The error is accurate about what the object holds. What is missing is the step
that would have put the columns there.

<details>

### The rule that already exists

`c.bnecfit()` collects `retained_data` from its inputs and keeps it where every
input is identical, messaging and dropping it where they are not
(`R/bnecfit-methods.R`, in the branch that allots `bayesmanecfit`). The reason
for that condition holds for any reassembly: a set whose members were fitted to
different data has no single frame to align observations against.

The same block applied in `expand_manec()` after the class is allotted would
close the gap for every caller. The work-around a caller has otherwise is to
reach into the constituents and attach the element by hand, which means
depending on the name of an element rather than on the interface.

### The prebayesnecfit case

`expand_manec()` accepts `prebayesnecfit` objects as well as fitted ones, and a
`prebayesnecfit` recovered from a stored unit may not hold `retained_data` even
where the fit it came from did. Whether the element survives
`recover_prebayesnecfit()` decides whether the fix belongs in `expand_manec()`
or earlier.

### Tests

`tests/testthat/test-autoplot.R` already covers retention through `pull_out()`,
`amend()`, `update()` and `c()`. The case to add is a set built by calling
`expand_manec()` directly on fits that each retain the same columns, asserting
the result retains them, and a second where two inputs retain different columns,
asserting it does not and says so.

### The open decision

Whether `expand_manec()` should set the element, or whether attaching it in
`bnec()` afterwards is deliberate and the documentation should instead tell a
caller assembling a set by hand that they must do the same. The first makes the
two routes behave alike; the second keeps `expand_manec()` to combining
equations and nothing else.

</details>

beckyfisher:
## What

`expand_manec()` now preserves compatible `retained_data` from separately
fitted objects recovered with `recover_prebayesnecfit()`. The completed model
set stores one copy of an identical retained frame. Different retained frames
are reported and omitted.

## Why it matters

A set reconstructed from separately computed equations can now use an unused
categorical column in `autoplot()`, as a set returned directly by `bnec()` can.
The result does not duplicate the retained frame in every component fit.

## Evidence

- On R 4.6.1, the `autoplot`, `bnecfit-methods`, and `expand_classes` test files
  passed 233 assertions. Two existing LOO and WAIC warnings were reported for
  the stored `manec_example` fit.
- `R CMD INSTALL` completed from the source package. Checks against that
  installed package reconstructed a set, resolved the retained plotting group,
  confirmed single-copy storage, preserved the `c()` behaviour, and omitted a
  conflicting frame with the documented message.

<details>

`recover_prebayesnecfit()` includes optional retained columns so that they
remain available when separate single-model objects are reconstructed. `expand_manec()`
reads those copies as assembly metadata, removes them from `mod_fits`, and uses
one shared helper to attach an identical frame to the result. `c.bnecfit()` now
reaches the same rule through `expand_manec()` while preserving its input-level
comparison when model names are deduplicated.

Regression tests cover direct reconstruction with identical and different
frames, plotting by an unused categorical column, and the absence of duplicated
frames in component fits.

</details>

Closes #380


beckyfisher:
Independent review completed in one cycle. No review comment identified a
genuine problem in the posted PR, so no review-driven change was required. The
reviewer checked the developer response and the unchanged head (`eab1bc22`);
developer and reviewer agree that the PR is sound. The 233 focused assertions
passed, with only the two existing LOO and WAIC warnings from the stored example
fit.

Issue #386, opened 2026-09-19
Default priors for non-complete response curves


beckyfisher:
Some concentration response experiments don't (and sometimes cannot, particularly for WET testing) show a complete response.

## What

Three of the default priors are derived from the observed range of the response, and on a design whose highest concentration has not reached the lower asymptote all three describe the design rather than the curve.

1. The `bot` prior is located at the mean response over the highest concentrations tested, which on such a design sits above the true asymptote. `regularizing_location()` records this as a known bias (`R/define_prior.R:385`), and the standard-error floor does not widen the prior to cover it, because a bias is not noise.
2. The spread of both the `top` and the `bot` prior is a multiple of `sd(response)` on the gaussian branch and of a quantile of the response on the positive branch. As the response flattens both quantities fall, so the prior narrows in exactly the regime where `bot` is least identified.
3. The `nec` and `ec50` priors are truncated to `[min(predictor), max(predictor)]` (`R/define_prior.R:1278`). Where the true threshold lies above the highest concentration tested it is outside the prior support, not merely in its tail.

## Why it matters

A whole effluent toxicity dilution series is set by the sample available rather than by the toxicity of it, so a mildly toxic sample gives a curve still falling at the undiluted end. The same design arises wherever a solubility limit, a supply limit or an ethics constraint caps the highest concentration.

On such data `bnec()` runs to completion and `summary()` reports a credible interval for `bot`, for `nec` and for `ec50`. Nothing in the output records that the prior placed the true value outside its own central mass, and nothing records that a threshold estimate sits at the truncation bound rather than in the interior. `ecx()` already messages when its estimate reaches the upper bound of `x_range` (`R/ecx.R:247`); `nec()` has no equivalent, and neither reports on the prior.

## Evidence

Prior-only, on `issue-382-example8-group-terms` at `30fcdbfa`, R 4.6.1, bayesnec 2.1.3.39, 2026-09-19. Nothing was fitted. Priors were obtained through `get_priors()` under `devtools::load_all()`, which runs the same `check_data()` and `define_prior()` sequence `bnec()` runs. The script is in the session scratch and is not committed; every figure below reproduces from the description given here.

[...]

PR #387, opened 2026-09-19
Warn when incomplete ranges constrain default priors

beckyfisher:
## What

`bnec()` now warns before fitting when the modelled response at the highest concentration has declined by less than 50% from the mean at the lowest concentration. Binomial data are assessed as proportions, rate data per unit exposure, and joint hurdle response and survival blocks separately. `bnec_group()` evaluates each level before fitting any of them and reports all affected series or blocks in one warning.

For a threshold fit, or a threshold-only model average with a common bound, `nec()` now reports when its median or upper interval limit reaches the upper bound of the fitted `nec` prior. The existing default truncation to the tested predictor range is retained, so this change does not enable extrapolation beyond the data.

## Why it matters

An incomplete concentration range can place the lower-asymptote prior above the true asymptote and can exclude a threshold above the tested maximum. The fit currently completes without identifying either condition. The new diagnostics state the limitation before sampling and when the estimate is read.

The 50% rule identifies a design that has not observed the midpoint of a decline towards zero. It is a diagnostic, not a completeness test: a larger decline does not prove that the lower asymptote was observed. A complete user-supplied replacement for the affected `bot`, `nec`, and `ec50` priors suppresses the default-prior warning for that response block; a partial prior does not. Equations without any of those parameters do not trigger it.

## Evidence

The reproduction in #386 gives endpoint declines of 2%, 36% and 92%. The first two now warn before fitting; the complete design does not. Tests also cover grouped series with level-specific model validity, variable trials and exposures, hurdle blocks with asymmetric custom priors, equations without affected defaults, non-positive control scales, partial and complete model-specific priors, an unconstrained NEC posterior, custom bounds, mixed NEC/NSEC models and blocks, differing component bounds, and bounds transformed through `xform`.

Closes #386.

<details>

### Prior-range diagnostic

`check_response_range()` computes the decline from the mean at the lowest predictor value to the mean at the highest on the response used to construct the model's default priors. It runs once before the equation loop in `bnec()`. The grouped route runs it once over each level before its level loop, retains which equations and affected defaults apply to each level and response block, and passes a private marker to prevent duplicate inner warnings.

The check is not inferred where the lowest-concentration mean is non-positive, because the response ratio is not defined on that scale. Rate fits now also pass their per-exposure response consistently to default-prior construction.

### NEC bound report

`report_nec_prior_bound()` reads the upper bound from the prior stored on the fitted object, including a user-supplied bound. It detects contact on the fitted scale and applies `xform` only when displaying the bound, so decreasing and non-monotone transformations cannot hide a constraint. Its wording does not attribute a custom bound to the tested range, and it requires the authoritative combined estimate type and every model-average component to be `NEC`. It therefore does not infer constraint for a mixed NEC/NSEC model set, a joint fit containing a smooth block, or unequal component bounds.

### Tests

- Built the source package and installed it with its tests.
- Ran `test-bnec.R`, `test-bnec_group.R`, `test-nec.R` and `test-rate.R` against the built and installed package; all passed.
- A non-CRAN `R CMD check` without examples or tests completed. Its two warnings and two notes are the existing worktree `.git`, unresolved `modifyList`, and deliberately unbuilt-vignette diagnostics.
- A wider `NOT_CRAN=true R CMD check` passed package installation, namespace loading, documentation, examples and several fitting test files without a failure. The later sampling tests stopped producing output locally and the run was terminated after approximately ten minutes without log activity.
- `git diff --check` passed.

</details>


beckyfisher:
Independent review completed after five cycles. The reviewer and developer agree this PR is sound.

The review identified and fixed these substantive problems in the original revision:

- The range diagnostic used raw outcomes. It now uses binomial proportions, rates per exposure, and separate survivor-response and survival views for joint hurdle fits; rate fits also pass the rate-scale response into default-prior construction.
- Any custom prior silenced the warning. Suppression now requires complete replacement of the affected `bot`, `nec`, or `ec50` defaults, tracked separately by equation and hurdle block. Equations without affected parameters do not warn.
- Grouped checks lost level-specific model and prior information. They now retain a level-by-block map, so a level is reported only when that same fitted block both uses an affected default and has an incomplete range.
- NEC-bound reporting overreached for custom bounds, mixed NEC/NSEC fits, smooth hurdle blocks, and unequal component bounds. It is now provenance-neutral and requires the authoritative combined estimate and every component to be NEC with one common bound.
- Bound contact was compared after `xform`, which fails for decreasing or non-monotone transformations. Detection now occurs on the fitted scale, with transformation used only to display the bound.

Regression coverage was added for each case. The focused suite passes from a freshly built and installed package, and `git diff --check` is clean.

beckyfisher:
Reviewed. The diagnostic is worth having and the placement is right, and the
change is one of the three options #386 listed rather than the one it turns on:
no prior entry changes, and the truncation at `max(prior_predictor)` is retained,
which was the item #386 recorded as needing a decision. So this should not close
#386. The six issues that do are #389 to #394, and the plan is on `dev` at
`notes/tasks/386-incomplete-designs-human.md`.

Three changes are proposed for this branch.

The `rate()` denominator now divides the response before `add_brm_defaults()` in
`R/fit_bayesnec.R`. That is a correct fix, matching `amend()` at `R/amend.R:329`
and the documented `rate()` semantics, and it changes the default priors, the
initial-value band and therefore the posterior of every `poisson` and
`negbinomial` `rate()` fit. It arrives here with no NEWS entry, no mention in the
body, and no test asserting the resulting prior. It is #389 and should be lifted
into its own pull request.

The 50% decline rule measures the magnitude of the effect rather than whether the
response has flattened. A sublethal endpoint whose true `bot` is 0.6 of the
control is a complete design and warns on every fit; a design that has fallen 92%
and is still declining steeply is incomplete and is silent. It also reads the
control as its denominator, so the hormesis case #386 asked about is not handled.
#390 replaces it with a comparison of the two highest levels against the
within-level standard error, which never reads the control and so needs no
hormesis special case.

`report_nec_prior_bound()` tests contact by `signif(estimate, 3) == signif(bound,
3)`, so a median at 0.997 of the bound is not reported, and it refuses to report
for a mixed NEC/NSEC model average, which is the common default `bnec()` path.
#392 replaces it by counting draws at or above the bound, which needs no
tolerance, and reporting per component. Two of its properties are kept there: the
assessment on the fitted scale before `xform`, and reading the bound from the
prior stored on the fit rather than from the data.

What to keep unchanged: the placement of the check in `bnec()` before the
equation loop and in `bnec_group()` before the level loop, with the private
marker; `uses_response_range_defaults()` including the partial-prior case; and
the per-block treatment of a hurdle fit.

Issue #389, opened 2026-09-19
Default priors for a rate() fit are built on counts, not rates


beckyfisher:
## What

`fit_bayesnec()` builds the default `top` and `bot` priors from the raw counts of
a `rate()` fit. The denominator is not divided out, so the priors are on the
count scale while the parameters they describe are rates.

## Why it matters

Under the identity link `bnec()` assigns, `brms` writes a rate denominator
multiplicatively on the response scale, so the mean is the rate and `top`, `bot`
and `nec` are counts per unit exposure. That is stated in the `rate()` section of
`?bayesnecformula`. The default priors are therefore displaced by a factor of the
exposure on every `poisson` and `negbinomial` `rate()` fit, which changes the
initial-value band as well and so changes the posterior.

## Evidence

`R/fit_bayesnec.R:59-64` divides by `trials` for the binomial families and by
nothing else:

```r
if (family$family == "binomial" || family$family == "beta_binomial") {
  response <- y / tr
} else {
  response <- y
}
```

`amend()` does divide, at `R/amend.R:326-332`, with the recorded reason that the
priors must be on the scale the original fit used (#136). So the two routes build
different priors for the same data, and neither the tests nor the documentation
assert which is intended.

The fix is three lines and is already written, as part of PR #387. It is pulled
out here because it changes fitted results and belongs in a pull request that
says so.

Plan: `notes/tasks/386-incomplete-designs-human.md` phase 1, specification §2.1.

<details>
<summary>Implementation detail</summary>

### The change

In `fit_bayesnec()`, beside the existing `trials` division:

```r
denominator <- retrieve_var(bdat, "rate_var")
# and, inside the !skip_check branch, denominator <- checked_df$mod_dat$denom
if (!is.null(denominator)) {
  response <- response / denominator
}
```

`response` is what reaches `add_brm_defaults()`, which passes it to
`define_prior()` and to the initial-value search, so one division serves both.

### Tests

- `tests/testthat/test-define_prior.R`: the `top` and `bot` entries for a `rate()`
  fit with a non-unit denominator, against the entries the same data produce with
  the denominator divided out by hand.
- The same default prior from `bnec()` and from `amend()` for the same `rate()`
  data. That equality is the defect's own statement and nothing asserts it now.

### Definition of done

The two routes agree, a test asserts the entry, and `NEWS.md` records it under a
behaviour-change heading naming `poisson` and `negbinomial`. The pull request
body gives the before and after entries for one worked example.

[...]

PR #396, opened 2026-09-19
Build the default rate priors on the rate scale (#389)

beckyfisher:
## What

`fit_bayesnec()` now divides a `rate()` denominator out of the response it
builds the default `top` and `bot` priors from, as it already divides `trials`
out for the binomial families. Closes #389.

This changes fitted results. Both the prior and the initial-value band change
for every `poisson` and `negbinomial` fit whose formula includes a `rate()` term
with a denominator that is not 1, and the posterior changes with them. No other
family accepts a `rate()` term, and a fit without one is unaffected.

A `rate()` denominator written as an expression, `rate(hours / 24)`, is now an
error. The denominator is recorded as a plain column, so an expression was
reduced to the first variable in it while `brms` kept the expression; without
the refusal this change would divide by the wrong column.

## Why it matters

`bnec()` fits on `link = "identity"`, so `brms` writes a rate denominator
multiplicatively on the response scale. The mean is then the rate, and `top`,
`bot` and `nec` are counts per unit exposure. That is stated in the `rate()`
section of `?bayesnecformula`. The defaults were built from the raw counts, so
every entry was displaced by the exposure: on a design whose exposures are 1, 2,
4 and 8, a `top` a user reads as a rate of about 20 was given a prior centred
near 64.

The same vector reaches `make_good_inits()`, so the starting values were
displaced with the prior. Under `amend()` the two disagreed with each other:
that route already divided the denominator out when it rebuilt a prior, but the
initial-value search that follows it ran on the counts.

## Evidence

The worked example is the series the new tests use: `nec4param`, `poisson`, a
true `top` of 20, exposures of 1, 2, 4 and 8, and counts reaching 171.

| entry | before | prior mean | after | prior mean |
|---|---|---|---|---|
| `top` | `gamma(2, 0.0312591579564325)` | 63.98 | `gamma(2, 0.102613435946769)` | 19.49 |
| `bot` | `gamma(2, 0.28761459644077)` | 6.95 | `gamma(2, 0.762494787633288)` | 2.62 |

`beta` and `nec` are unchanged, since neither is built from the response.

Ten of the thirteen assertions in the six compilation-free tests fail with the
division disabled. The three that hold either way are the count-scale prior
mean, which states the value before the change, and the two guard tests for a
formula with no `rate()` term and for a denominator of 1, which are no-ops by
construction. The `amend()` test fails without the division as well.

Two routes already divided, so this is `bnec()` agreeing with them rather than a
new convention. `get_priors()` divides at `R/get_priors.R:216-218`, and
`amend()` at `R/amend.R:329-332` with the recorded reason that a prior rebuilt
for a model added to an existing set must be on the scale the original fit used
(#136). `?get_priors` states that the priors previewed from a formula and the
priors a fit used agree where no prior was supplied; on `rate()` data they did
not, and a new test pins that they now do.

On the refusal: for `y | rate(exh / 24) ~ crf(x, "nec4param")` the model frame
held `exh` while the `brms` formula kept `exh / 24`, so with `exh = 24 * ex` the
`top` entry came out as `gamma(2, 2.463)`, a prior mean of 0.81 against a true
`top` of 20. The same mismatch already reached `expand_classes.R`, which pins
the recorded column at 1 rather than the expression, and the plotting paths,
which divide the observations by that column.

`tests/testthat/test-define_prior.R` and `test-rate.R` pass under
`NOT_CRAN=true`, against an installed build as well as under
`devtools::load_all()`.

## Decisions taken here

The three copies of the division, in `fit_bayesnec()`, `get_priors()` and
`amend()`, are left as three. Factoring them into one internal helper would make
the agreement structural rather than asserted, but it rewrites two functions
that are already correct, on a branch whose only purpose is a behaviour change
that has to be readable on its own.

A `rate()` expression is refused rather than recorded. Recording it would mean
teaching the prediction grid, both plotting paths and the model frame to hold an
expression instead of a column, which is a larger change than the defect
warrants.

The centring constant of a `disp()` variance function is left alone and filed as
#397. It is computed from the raw counts while the mean it centres is the rate,
which predates this change. One entry beside it does change here: the slope
prior of `disp("loglinear")` on a `negbinomial` fit, which is scaled by
`sd(response)`. `c0`, the `power` and `twosided` slopes and the dispersion
initial values are fixed constants for that family and are unaffected.

A non-syntactic column name in a `rate()` or `cens()` term ends in a parse
error from `str2lang()` rather than a message naming the column. That predates
this change and is filed as #398; it is noted here because the refusal above
tells a user to compute the denominator as a column, and a spreadsheet column
with a space in its name is the path into it.

These three lines also appear in PR #387, which is being re-scoped. Nothing else
from that branch is here. This is phase 1 of
`notes/tasks/386-incomplete-designs-human.md`, specification §2.1, and every
later phase is measured against the priors this one produces.

<details>

### The change

`R/fit_bayesnec.R`, three lines. `rate_var` is read from the model frame beside
`trials_var`; inside the `!skip_check` branch it is replaced by
`checked_df$mod_dat$denom`, so that the divisor and the response it divides come
from one object; and `response` is divided by it after the binomial branch.

`check_data()` has built `mod_dat$denom` since #136, with a comment saying that
the response divided by it is what the priors are built from. Nothing read it.

The division is applied to `response` rather than inside `define_prior()`
because `response` is what `add_brm_defaults()` passes to both `define_prior()`
and `make_good_inits()`. One division therefore serves the prior and the
initial-value band; placing it in `define_prior()` would have corrected the
prior and left the starting values on the count scale. It does not serve the
`disp()` centring constant, which is built in `make_brmsformula()` from a
separate read of the response; that is #397.

`R/bayesnecformula.R`, in `split_calls()`, refuses a `rate()` argument that is
not a symbol. `split_calls()` has one caller, inside the model frame method, and
`bnec()`, `get_priors()` and `amend()` each build a model frame, so one refusal
covers every consumer of the recorded column. The error deparses the call whole
rather than rebuilding it around its argument, because `check_formula()`'s aterm
pattern is unanchored and a user-defined `myrate()` reaches the same branch.

### Tests

Seven in `tests/testthat/test-define_prior.R`, one in `test-rate.R`.

All but one compile nothing. They drive `fit_bayesnec()` with `brm()` mocked and
read the prior off the call, following `fit_call()` in `test-fit_bayesnec.R`,
which cannot be shared because `testthat` runs each file in its own subprocess:

- the `top` and `bot` entries against the entries the same data produce with the
  denominator divided out by hand, with the count-scale entries asserted to
  differ, and the two prior means stated as numbers;
- a formula with no `rate()` term, which must keep the entries it had;
- a denominator of 1, likewise;
- the fitted entries against what `get_priors()` previews, on `negbinomial`;
- the vector `make_good_inits()` is given, with that function mocked as well,
  which is the initial-value band the first paragraph of the NEWS entry claims
  changes;
- the `skip_check = TRUE` path, which is the read `amend()` uses and the only
  line the fitting test below does not reach.

The last compares the `bnec()` and `amend()` routes end to end and is wrapped in
`skip_on_cran()`. `amend()` rebuilds a default prior only for a model added to a
set it has already fitted, so one real fit is needed; it is one equation, two
chains and 400 iterations, and the added model's prior is read off a mocked
`brm()` call, so only one Stan program is compiled. Measured at 80 s.

`test-rate.R` gains the refusal: the expression form errors and names itself,
and the bare column still parses.

</details>


beckyfisher:
Three rounds of independent review. The body above describes the current state;
this records what the review changed, so a reader does not have to reconstruct
it from the commits.

## Defects in the change itself

A `rate()` denominator written as an expression divided by the wrong quantity.
`split_calls()` records the denominator as `all.vars()[1]` of the `rate()` call,
so for `y | rate(exh / 24) ~ ...` the model frame held `exh` while `brms` kept
the quotient; `make_standata()` returned a `denom` of 1, 2, 4, 8 against a
recorded column of 24, 48, 96, 192. Before this branch the mismatch reached the
prediction grid and the plotting paths; with the division added it reaches the
posterior, and on the test series gave a `top` prior mean of 0.81 against a true
`top` of 20. The form is now refused in `split_calls()`, which is the one place
every route passes through. Commit 2fed9d31.

The refusal then named a call the user had not written.
`check_formula()`'s aterm pattern is unanchored, so a user-defined `myrate()`
reaches the same branch, and a message rebuilt as `rate(<argument>)`
misreported it. The call is now deparsed whole. Commit 8d8e09ed.

## Claims that were not true as stated

The Evidence section said every new assertion failed without the fix. Measured,
it is ten of the thirteen assertions in the six compilation-free tests. The
three that hold either way are the count-scale prior mean, which states the
value before the change, and the two guard tests, which are no-ops by
construction. A test count of six that should have been seven was corrected with
it.

The claim that the initial-value band changes had no test, and neither did the
`skip_check = TRUE` read, which is the line `amend()` depends on. Both are now
pinned without compilation, the first by mocking `make_good_inits()` as well as
`brm()`.

A code comment said `check_data()` drops rows. It does not; it substitutes into
the response. The alignment argument was right and the sentence claimed more
than is true.

`?bayesnecformula` said "those four" over a list of three, and the
design-weights paragraph ran on from the new one. The NEWS heading named a model
rather than a fit, and the rejected helper refactor was stated only inside the
collapsed block, against section 14 of the contributing conventions.

## Defects found in existing code, filed rather than fixed here

- #397, the `disp()` centring constant computed from counts while the mean it
  centres is the rate: `3.04452` in the generated block against `2.67415`.
- #398, a non-syntactic column name in a `rate()` or `cens()` term reaching
  `str2lang()` and dying in the parser.

## Raised and not taken

Renaming the local `denominator` to `denom`, which would match the column but
make `checked_df$mod_dat$denom` read as a self-assignment. Fixing #397 on this
branch, which rewrites a literal in the Stan program and so needs its own
behaviour-change entry. Raising the `testthat` bound in `DESCRIPTION`, which is
`>= 3.1.8` while `local_mocked_bindings()` arrived in 3.2.0; that predates these
tests and belongs in its own change.

Issue #390, opened 2026-09-19
Detect an incomplete design from a contrast at the top of the series


beckyfisher:
## What

Replace the incomplete-range rule in PR #387. It triggers on how far the response
fell from the control, which does not separate a design that stopped early from
one whose lower asymptote is genuinely high. Trigger instead on whether the
response is still declining between the two highest concentrations tested, tested
as a one-sided contrast in a generalised linear model that uses the fitted
family's own mean-variance relationship.

## Why it matters

A user is told, before anything is fitted, that the top of the series has not
flattened. That report is what tells them to declare the design and get the
prior set built for it.

The rule as written misreports in both directions. A sublethal endpoint whose
true `bot` is 0.6 of the control is a complete design, and it warns on every fit.
A design that has fallen 92% and is still declining steeply is incomplete, and it
is silent. A rule that fires on ordinary well-designed data teaches users to
ignore it.

A whole effluent dilution series whose survival curve is still falling at the
undiluted end is the commonest incomplete design there is, and the survival block
of a hurdle fit holds one proportion per concentration. Any rule built on
within-level replication cannot be computed there at all.

## Evidence

`check_response_range()` in PR #387 computes `1 - mean(y at max(x)) / mean(y at
min(x))` and warns below 0.5.

The #386 measurements do not separate the two cases either: every design there
was simulated with a true `bot` of 0.1, so a small observed effect and an
unobserved asymptote are confounded in that simulation. The three designs it
reports, at 92%, 36% and 2% of maximum effect, are all incomplete by
construction, and no complete design with a high asymptote was run.

The rule also reads the control as its denominator, so on a hormesis design,
where the response rises before it falls, the decline is measured against a level
that is not the peak. #386 raised this as an open question and PR #387 does not
address it. A rule that compares the two highest levels never reads the control,
so the hormesis case needs no special handling.

`survival_by_x()` returns one proportion per unique predictor value, clamped to
`[eps, 1 - eps]` with `eps = 1 / (2 * length(response))` so that Stan can fit it
under the identity link. The survival block of a hurdle fit therefore has `n = 1`
at every level, and the values it holds are not the observed proportions where
any were 0 or 1.

Plan: `notes/tasks/386-incomplete-designs-human.md` phase 2, specification §2.2.

<details>
<summary>Implementation detail</summary>

### The statistic

```r
# R/check_data.R
check_response_flattened <- function(data, family, group = NULL,
                                     blocks = c("response", "survival"),
[...]

Issue #394, opened 2026-09-19
A prior set for designs that do not reach the lower asymptote


beckyfisher:
## What

`bnec()` gains `asymptote_observed`, a declaration of whether the lower asymptote
was observed. Under `asymptote_observed = FALSE` the `bot` prior spans from the
support floor to the observed endpoint, and the initial-value band is widened to
match.

## Why it matters

The `bot` prior is located at the mean response over the highest concentrations
tested. On a design that stopped before the asymptote, that mean sits above the
true value, and the spread narrows as the response flattens, so the prior is at
its narrowest exactly where `bot` is least identified. The fit completes and
`summary()` reports a credible interval with nothing recording that the prior
placed the true value outside its own central mass.

## Evidence

Prior-only, from #386. A `nec4param` response was simulated on a log-spaced
series of eight concentrations by five replicates, with `top` 0.9 and `bot` 0.1,
under three settings of the threshold and the decay rate. The quantity is the
truncated prior CDF at the true `bot` of 0.1.

| maximum effect | prior set | `bot` prior | CDF at the truth |
|---|---|---|---|
| 0.92 | `uninformative` | `normal(0.084, 0.960)` | 0.507 |
| 0.92 | `regularizing` | `normal(0.076, 0.384)` | 0.525 |
| 0.36 | `uninformative` | `normal(0.607, 0.287)` | 0.0386 |
| 0.36 | `regularizing` | `normal(0.579, 0.115)` | 1.45e-05 |
| 0.02 | `uninformative` | `normal(0.848, 0.101)` | 6.43e-14 |
| 0.02 | `regularizing` | `normal(0.913, 0.040)` | 2.30e-90 |

The spread falls with the location, from 0.960 to 0.101 and from 0.384 to 0.040,
because `sd(response)` falls from 0.384 to 0.040 across the three designs.

`regularizing_location()` records the bias at `R/define_prior.R:385` and names
`"uninformative"` as the set to use on such a design. The gaussian row above
qualifies that: the uninformative entry is affected too, at 6.43e-14 on the flat
design, so it is a mitigation on the positive branch and not on the gaussian one.

`init_limits()` records the same bias for the initial-value band at
`R/inits_functions.R:634`, and states that no width covers it: coverage was 0.160
at a width of four and 0.172 at five, against 0.003 at one.

Plan: `notes/tasks/386-incomplete-designs-human.md` phase 6, specification §5.
[...]

Issue #392, opened 2026-09-19
Add extrapolate to nec() and nsec(), reading the censoring record


beckyfisher:
## What

`nec()` and `nsec()` gain an `extrapolate` argument, which returns a no-effect
estimate beyond the prediction range for a caller who asks for it by name. The
default reports the estimate as censored, reading the record #395 puts on the
posterior.

## Why it matters

Once the tested-range truncation is removed (#393), a threshold beyond the tested
range is inside the prior support and the posterior can place mass there. A user
who wants that value needs a way to ask for it, and a user who does not must keep
getting an estimate inside the range they measured.

Both ends are in scope. #393 removes `lb` and `ub` in the same two lines, so a
threshold below the lowest concentration tested becomes admissible at the same
moment, and leaving it unreported would recreate #386's defect at the other end.
That end binds on a logged predictor or a design with no control; where the
predictor is a concentration with a zero control, `lb` was already the
lognormal's own support and nothing changes.

The two halves of a model-averaged N(S)EC cannot be extrapolated the same way,
and the difference decides what the argument can accept. A NEC is a sampled
parameter with no grid behind it, so releasing it is free. An NSEC is read off a
fitted curve, so extending it means evaluating that curve further out.

## Evidence

`nsec()` and `ecx()` already take `x_range`, at `R/nsec.R:129` and `R/ecx.R:128`,
and both recompute from the stored fit rather than refitting, so extending the
grid for a curve-read quantity is a post-processing call and not a refit.

`nec()` has no such control and needs none, because `b_nec_Intercept` is sampled
and no grid is involved.

An infinite limit is therefore available for a sampled NEC and is not computable
for an NSEC: a curve cannot be read off an infinite grid. On the default
`bnec()` set, which mixes `nec`-type and `ecx`-type equations, this is the common
case rather than an edge one.

Plan: `notes/tasks/386-incomplete-designs-human.md` phase 5, specification §4.2.
Depends on #395, which records which draws are censored and at what bound.

<details>
<summary>Implementation detail</summary>

### Signatures

```r
nec(object, posterior = FALSE, xform = identity,
    prob_vals = c(0.5, 0.025, 0.975), extrapolate = FALSE, ...)

nsec(object, sig_val = 0.01, resolution = 200, x_range = NA,
     xform = identity, prob_vals = c(0.5, 0.025, 0.975),
     extrapolate = FALSE, ..., dpar = NULL)
```

| `extrapolate` | behaviour |
|---|---|
[...]

Issue #391, opened 2026-09-19
Incomplete designs in the default-prior audit


beckyfisher:
## What

Extend `notes/scripts/prior_audit.R` with designs that do not reach the lower
asymptote, so that the prior changes planned in #386 have a baseline and can be
shown not to have regressed the complete-design cells.

## Why it matters

No existing audit covers this design. `notes/scripts/prior_audit.R` simulates
every design so that it descends to within 5% of `bot` at the highest dose, and
states that `bot` is therefore identified in every cell.
`notes/scripts/init_search_audit.R` excludes the cells whose predictor stops
short of the crossing from its width rule and reports them separately.

#386 names re-running the audit as the definition of done for any change to a
default prior entry. That is not available until the audit can simulate the
design in question.

## Evidence

The figures #386 quotes were produced by a script held in a session scratch and
not committed, so they are not currently reproducible from the repository. The
three designs it used, at 92%, 36% and 2% of maximum effect, are the cells to
add.

Two later phases read numbers this script does not yet report: the number of
proposals `make_good_inits()` makes, and whether the flatness rule of #390
reports. That rule has a stated level rather than a tuned threshold, so what the
complete cells measure is whether its realised rate agrees with that level on the
families whose variance is not homogeneous across the series.

Plan: `notes/tasks/386-incomplete-designs-human.md` phase 3, specification §6.

<details>
<summary>Implementation detail</summary>

### The design axis

Cross the existing design list with a completeness axis:

- complete, as now, descending to within 5% of `bot` at the highest dose;
- incomplete, the same generating curve with the series truncated so that the
  response at the highest retained concentration is a stated fraction of the way
  from `top` to `bot`. Use 92%, 36% and 2% of maximum effect, so the new cells
  reproduce the figures #386 quotes.

### Columns

Per cell, as now, the truncated prior CDF at the true value for `top`, `bot`,
`nec` and `ec50`. Add:

- the number of proposals `make_good_inits()` makes before the band accepts a
  full set, or the cap;
- whether the flatness rule of #390 reports.

### The rates #390 needs

The complete cells give its false-positive rate. The rule has a stated level
rather than a tuned threshold, so what is measured is whether the realised rate
agrees with that level on the families whose variance is not homogeneous across
the series. A rate far from `alpha` there is a defect in the variance handling,
not a threshold to adjust.

The incomplete cells give its miss rate, which decides whether the two
approximations in #390 are tolerable. Three cells beyond the existing family
sweep:

[...]

PR #399, opened 2026-09-19
Report a response still declining at the top of the series

beckyfisher:
## What

`bnec()` reports, before anything is fitted, where the response is still
declining between the two highest concentrations tested. The contrast is fitted
as a generalised linear model under the mean-variance relationship of the family
being fitted and tested one-sided at the 5 per cent level. This replaces the
rule in PR #387, which triggered on how far the response had fallen from the
control.

## Why it matters

The report is what tells a user to declare an incomplete design and get the
prior set built for it (#394), so a rule that fires on ordinary well-designed
data teaches them to ignore it.

The rule in PR #387 misreports in both directions. A sublethal endpoint whose
true `bot` is 0.6 of the control is a complete design and warns on every fit; a
design that has fallen 92 per cent and is still falling steeply is incomplete
and is silent. It also reads the control as its denominator, so on a hormesis
design the decline is measured against a level that is not the peak. A rule that
compares the two highest levels never reads the control, so the hormesis case
needs no special handling and neither does a response on a log-ratio or centred
scale whose control mean is not positive (#229).

A whole effluent dilution series whose survival is still falling at the
undiluted end is the commonest incomplete design there is, and the survival
block of a hurdle fit holds one proportion per concentration. Any rule built on
replication within a concentration cannot be computed there at all. This one
takes a binomial contrast on the counts, where the number of individuals behind
each proportion supplies the information replication would have.

## Evidence

Calibration was measured by calling the contrast directly on simulated blocks of
five predictor levels whose mean is equal at the two highest, so the true
contrast is zero and the report rate is the false-positive rate. Four
replicates per level, R 4.6.1.

| block | replicates | report rate on a flat top |
|---|---|---|
| gaussian, dispersion pooled over the series | 20000 | 0.0499 [0.0469, 0.0529] |
| gaussian, dispersion from the two levels | 20000 | 0.0508 [0.0478, 0.0538] |
| Gamma | 2000 | 0.061 |
| poisson | 2000 | 0.049 |
| negbinomial, size 5 | 2000 | 0.050 |
| binomial, 20 trials | 2000 | 0.053 |
| beta-binomial, 20 trials | 2000 | 0.053 |
| beta | 2000 | 0.059 |
| hurdle survival, 20 individuals | 2000 | 0.055 |
| bernoulli | 2000 | 0.088 |
| beta-binomial, trials 10, 20, 40, 80 within a level | 3000 | 0.088 |
| beta-binomial, trials 5, 10, 20, 100 within a level | 3000 | 0.137 |
| binomial, one row per level, 20 trials | 4000 | 0.058 |
| beta-binomial, one row per level, 20 trials, correlation 0.1 | 4000 | 0.184 |

Four rows sit above the nominal level. All are recorded rather than tuned,
because the audit that measures these rates over the incomplete-design cells is
#391 and does not exist yet, and each is a named row for it.

The last row is the largest miscalibration in the rule, and it follows from a
decision worth disagreeing with, so it is stated here rather than in the fold. A
binomial or beta-binomial response with one row per concentration is assessed
under a fixed dispersion, because quasibinomial cannot estimate one without
replication and the alternative was to pass the design over --- one composite
sample per dilution with twenty individuals scored is the ordinary whole
effluent layout, and passing it over reports on none of the designs that are
genuinely incomplete. For `binomial` that fallback is the family's own variance
function and the rate is nominal at 0.058. For `beta_binomial` it fits a
variance the family's declaration denies, and 0.184 is the result. Supplying
replicate rows restores the over-dispersion estimate on both.

The bernoulli row is the discreteness of a 0/1 response at this size rather
than anything in the contrast: at 20000 replicates it returned 0.082 with four
observations per level, 0.063 with ten and 0.049 with twenty-five, on one
simulated series. The rate is not monotone in that count, because the
discreteness depends on the level of the response as well as on the number of
observations, so those three figures do not bound it. A block of
one individual per level is not merely coarse but degenerate, reading 0.16 to
0.25 when the reviewer of this PR measured it, and it is now passed over: the
same measurement returns 0 on both the bernoulli and the survival block. The
guard does not reduce the report rate where the information is real, since it
cannot fire at ten individuals: a five-level survival series of true
probabilities 0.95, 0.8, 0.6, 0.4 and 0.2, ten individuals at each, so that the
contrast tested is 0.4 against 0.2, is reported on 0.27 of 4000 replicates
before and after.

The two beta-binomial rows with varying trials are the approximation the
specification records and does not assert a direction for. As a proportion the
beta-binomial variance is `p(1-p)/n [1 + (n-1) rho]`, whose bracket depends on
`n`, so quasibinomial absorbs it as a constant multiplier only where the trials
are constant within a level. The rows above were measured at an intra-class
correlation of 0.1 and were raised by the reviewer of this PR; the constant-trials
row is 0.054 on the same generator. `NEWS.md` now states the 5 per cent claim
for the families whose variance function the test fits exactly and names these
two as the exceptions.

The tests here hold a 500-replicate check that the gaussian and survival rows
report below 0.1 on a flat top and above 0.5 where the top is genuinely still
falling.

The contrast is tested by analysis of deviance rather than by the Wald statistic
`summary.glm()` prints, which is a departure from the word of the specification
and was measured before it was taken. A dilution series running from ten of ten
individuals alive to none alive separates: the Wald standard error is 77224 and
the two-sided p is 0.9995, so the rule would have been silent on the clearest
incomplete design there is. The analysis of deviance on the same data returns
p = 7e-08. The two agree exactly for a gaussian response, which is the row the
specification requires to be the two-sample t test: on one simulated design the
deviance route and `t.test(var.equal = TRUE, alternative = "less")` both
returned 0.8526378.

Nothing was added to `DESCRIPTION`. `glm()` and the quasi families are in
`stats`.

Plan: `notes/tasks/386-incomplete-designs-human.md` phase 2, specification
section 2.2. Closes #390, which `dev` will not close by itself.

<details>

### The statistic

`check_response_flattened()` in `R/check_data.R`. For each block it takes the
two highest distinct predictor values present in that block, fits
`response ~ level` with the level as a factor, and compares that against the
same model with the two levels merged. The one-sided p is half the
one-degree-of-freedom F or chi-square p where the contrast is negative, and one
minus that half where it is positive.

The level is stated rather than a multiple of a standard error, so the number in
the code has a meaning: on a design whose top is flat the true contrast is zero
and the rule reports at `alpha`.

### The family mapping

| block | model |
|---|---|
| gaussian | `glm(gaussian)` |
| Gamma | `glm(Gamma(link = "log"))` |
| poisson, negbinomial, zero-inflated poisson and negbinomial | `glm(quasipoisson(link = "log"))` |
| binomial, beta_binomial | `glm(quasibinomial)` on `cbind(successes, failures)` |
| bernoulli | `glm(binomial)` |
| beta | `glm(quasibinomial)` on the proportion |
| hurdle `mu` | the row above for `hurdle_mu_family()` |
| hurdle `hu` | `glm(binomial)` on the non-zero and total counts |

The `log` link is written out for the Gamma and count rows. The `inverse` link
`stats::Gamma()` takes by default reverses the sign of the contrast, which is
the same class of trap as the identity link `bayesnec/CLAUDE.md` records for the
fit itself. `tests/testthat/test-check_response_flattened.R` asserts the link of
each row.

A `rate()` denominator enters as `offset(log(denominator))` rather than by
dividing the response, which keeps the counts integral and puts the contrast on
the rate scale. #389 refuses a `rate()` denominator written as an expression, so
the column is bare and the offset is the exposure the fit uses.

### The dispersion

Estimated from the two levels being compared, because for every family but
gaussian the variance depends on the mean and a dispersion estimated over a
whole declining series is pulled up by the high-mean levels. Gaussian is the
exception: `bnec()` fits a single `sigma` unless a `disp()` term is supplied, so
`bnec()` and `bnec_group()` pass
`pool_dispersion = is.null(parse_disp_term(formula))`.

### The hurdle blocks

`survival_by_x()` clamps its proportions to `[eps, 1 - eps]` so that Stan can
fit them under an identity link, so `flatness_blocks()` recomputes the non-zero
and total counts from the response instead. Dispersion is held at 1, which is
the assumption the `hu` block of the fit makes. The `mu` block is the survivors
only, so where nothing survived the highest concentration its two highest levels
are not the design's; each block is assessed against its own and the report names
the concentrations used.

### Edge cases

- Fewer than two distinct predictor values in a block: silent.
- Too little information at the two levels compared: silent. For a family
  whose dispersion is estimated that is a block with no predictor value
  observed twice, because a factor fit on one observation per level is
  saturated and no contrast is defined. For a bernoulli response or a hurdle
  survival block, whose dispersion is fixed, it is fewer than two individuals
  at either level, because a two-by-two table of single trials reads p = 0.048
  for any pair (1, 0) against an exact conditional p of a half. Section 2.2
  asks for a report on an unreplicated design and it is not given: that shape
  is an ordinary continuous predictor rather than a degenerate design, and
  `nec_data` has 100 distinct predictor values in 100 rows, so a report would
  appear in every vignette and every example that fits it. What section 2.2
  wants protected is the next case, which is kept.
- Replication present and no observation differing from another at its own
  predictor value: reported as
  a check that could not be applied. Tested on the observations rather than on
  the estimated dispersion, because the estimate of a zero dispersion is a
  rounding residue -- 1.8e-32 on a constant gaussian response -- which no
  comparison against zero catches. Families whose dispersion is fixed are
  exempt.
- A `glm()` that does not converge, or a non-positive `rate()` denominator: the
  same. The report is a `message()` throughout and cannot error.
- A binomial family with no `trials()` term: silent, so that `check_data()`'s
  refusal is what the user sees.
- A family with no row in the table: silent.

A `binomial` or `beta_binomial` response with one row per concentration is the
exception to the replication rule. Those rows are mapped to quasibinomial,
whose dispersion needs replication, so the rule would pass over one composite
sample per dilution with twenty individuals scored in it --- the ordinary whole
effluent binomial layout, and this issue's own case in different clothes. The
identical counts reaching the check as a hurdle survival block are tested, so
the same data would give two answers depending on which family declared them.
The denominator is known, so the contrast falls back to a fixed dispersion and
is assessed on the individuals, which is what the survival block already does.
On the four-level design above the two routes now return the same p to every
digit. Where replication is present quasibinomial is kept, because
over-dispersion between replicate vessels is real.

The fallback trades the over-dispersion estimate for a test that exists at all.
On `beta_binomial`, whose declaration says to expect over-dispersion, that makes
the test anti-conservative, and by more than the fallback's other rows: on a
flat top of four levels of twenty trials with one row at each, over 4000
replicates, `binomial` reports on 0.058 and `beta_binomial` at an intra-class
correlation of 0.1 on 0.184. The alternative was to pass the design over, which
reports on none of the designs that are genuinely incomplete. It is stated in
`NEWS.md` and is an audit item for #391 beside the varying-trials rate.

### Placement, kept from PR #387

Raised once in `bnec()` before the equation loop and once in `bnec_group()`
before the level loop, with a private `.bayesnec_flatness_checked` marker that
`bnec()` strips before anything reaches brms.
`check_model_survival()` is now called a few lines earlier in `bnec()`, because
the report needs it, and `bnec_group()` calls it before its level loop. One
behaviour change follows and is in `NEWS.md`: an invalid `model_survival` is now
refused before any level is fitted rather than by the first level's `bnec()`
call, after that level had compiled and sampled. `uses_response_range_defaults()` is taken unchanged, including
its partial-prior case, because `fill_missing_priors()` fills the omitted rows
from the same defaults.

### Reporting

The message names the predictor and the response as the formula wrote them,
read from the model frame's column names. The model frame holds the predictor
transformed, so a fit on `crf(log(concentration))` would otherwise report log
concentrations as concentrations, which is the trap `bayesnec/CLAUDE.md`
records for the estimators. Where the reported mean is a proportion or a rate
rather than the mean of the response column, it is named as one.

`message()`, not `warning()`: both existing pre-fit reports use it, and one
complete design in twenty is too often for a warning on something advisory. The
message names the block, the two concentrations, the mean at each and the
p-value, and for a grouped call the level. The two cross-references the
specification asks for land with #394 and #392, so until then the message names
`get_priors()` and the `prior` argument as PR #387 already does.

### Tests

`tests/testthat/test-check_response_flattened.R`, 34 tests, no Stan program
compiled. Every family row on a design still declining and one that has
flattened; the Gamma and count link; a hurdle fit whose survival block is
falling and whose response block is not, and the reverse; a hurdle fit with no
survivors at the highest concentration; the raw counts against the clamped
proportions; a hormesis design; a non-positive control mean; a `rate()` fit
against the same counts with the denominator held at 1; the unreplicated and
no-variation cases; the pooled and local gaussian dispersion; the grouped route;
the prior that silences the report and the partial prior that does not; and that
the report is raised once per `bnec()` call rather than once per equation.

### The scope of the report

The decision of what to do about an incomplete design is the user's, and the
argument that records it is #394. This report only says what the data show.

</details>


beckyfisher:
## Review record

Three rounds of independent review. The body above describes the current state
of the change; this comment records what the review found wrong, so that the
reasoning is not lost in the diff. Six findings were real and are fixed. The
rest were confirmed correct or were accepted as recorded limitations.

### The report fired on the package's own example data

`nec_data` is 100 rows at 100 distinct predictor values, and every vignette and
a dozen `@examples` blocks fit it. A block with one observation per predictor
value has a saturated factor fit, which the first version reported as a contrast
that could not be applied, so every one of those calls acquired an advisory. The
rule now passes such a block over in silence, which is a deliberate departure
from section 2.2 of the plan and is recorded as one in the code, the tests, the
edge-case list above and `NEWS.md`.

### The same guard exempted bernoulli, where it was most wrong

The fix above was gated on whether the family estimates a dispersion, on the
ground that a binomial count needs no replication. That is true of twenty
individuals and false of one. Two single Bernoulli trials reading (1, 0) give a
deviance p of 0.048 against an exact conditional p of a half, so an unreplicated
bernoulli design reported on 0.199, 0.242 and 0.163 of flat tops at true
probabilities of 0.3, 0.5 and 0.8, and a survival block of one individual per
concentration on 0.204. All four are now 0. The guard asks where the
information lies in each family: replication where the dispersion is estimated, and
two individuals at each of the two levels where it is fixed.

### The reported concentrations were on the fitted scale, unlabelled

The model frame holds the predictor as the formula wrote it, so a fit on
`crf(log(concentration))` reported log concentrations as concentrations, and
neither number appeared anywhere in the user's data. This is the trap
`bayesnec/CLAUDE.md` records for the estimators. The message now names the
predictor and the response as the formula wrote them, and names the reported
mean as a proportion or a rate where it is one.

### The false-positive rate away from its nominal level

Quasibinomial absorbs a beta-binomial over-dispersion as a constant multiplier
only where the number of trials is constant within a level. Measured at an
intra-class correlation of 0.1, the report rate on a flat top is 0.054 at
constant trials, 0.088 at trials of 10, 20, 40 and 80 within a level, and 0.137
at 5, 10, 20 and 100. The bernoulli rate is above the nominal level at small
counts and is not monotone in them. `NEWS.md` now claims 5 per cent only for the
families whose variance function the test fits exactly and names all three
exceptions, which #391 measures.

### A `rate()` denominator was dropped on the `mu` block of a hurdle fit

Latent rather than live, because a `rate()` term is refused for anything but
poisson and negbinomial, so no hurdle fit has a denominator today. It would have
put that block's contrast on counts rather than rates the day one can.

### `dots$prior` partial-matched `prior_type` in `bnec_group()`

`$` partial-matches on a list, so a call passing `prior_type` and no `prior`
read the wrong element. The consequence was benign here, but `bnec()` has a
comment saying `prior` is an explicit argument precisely so that this cannot
happen. Read with `[[` now.

### The findings not acted on

The analysis of deviance is kept in place of the Wald statistic
`summary.glm()` prints, and an exact conditional test was considered and
rejected for the two rows whose dispersion is fixed. On a whole effluent series
of ten individuals per concentration falling from 0.4 to 0.2 at the top,
`fisher.test()` gives p = 0.085 and does not report where the deviance test
gives 0.032 and does. Exactness bought on a shape the information guard now
excludes is not worth losing the case this issue exists for.

`bnec_hurdle()` emits two reports that read alike, because it makes two `bnec()`
calls on two datasets and neither is a hurdle fit. The check would need a second
private argument to know where it is, and `bnec_hurdle()` names the component on
the line immediately above each report.

### The findings of no defect

The contrast is exactly the one-sided pooled-variance t test for a gaussian
response, to all printed digits, on both the pooled and the two-level paths. The
factor indexing, the explicit links and the sign of the contrast on each, the
`rate()` offset, the hurdle survival block reading raw counts, the private
marker never reaching `brms`, and the once-per-call placement were each checked
and are correct. The calibration table above was reproduced independently.

Issue #395, opened 2026-09-19
Report every censored no-effect and effect-concentration estimate as censored


beckyfisher:
## What

Record the censoring of the no-effect posterior where that posterior is realised,
inside `bnec()`, and report it from `summary()`, `print()` and `autoplot()`.
Apply the same censored summary to every estimator that currently deletes an
out-of-range draw: `nsec()` and `ecx()` as well as the stored N(S)EC.

At present a beyond-range draw is deleted and the remaining draws are summarised
as though nothing had been removed.

## Why it matters

The N(S)EC is the value most often reported from this package, and it is not
computed by `nec()`. `expand_nec()` stores `ne_posterior` and its summary `ne`;
`expand_manec()` stores `w_ne_posterior` and `w_ne`. `summary()` reads those
stored summaries through `clean_nec_vals()` and never calls `nec()`, so a
censoring report added to `nec()` alone would not reach the number a user
reports.

Every estimator that can return `NA` for a draw shares the same defect. `ecx()`
summarises with `quantile(..., na.rm = TRUE)` at `R/ecx.R:232`, and `nsec()` does
the same at six places in `R/nsec.R`. The draws deleted are those whose curve did
not reach the target, which are the largest, so each reported estimate is lower
than the quantity it is labelled as and its interval is narrower. Lower is the
protective direction, which is why this has not caused trouble; the narrow
interval is the more serious half.

The model-averaged N(S)EC is also a mixture of two quantities that handle the
same condition in opposite ways. For an `ecx`-type equation `ne_posterior` is an
NSEC read off the prediction grid, and a draw whose curve does not fall to the
reference within the grid is `NA`. For a `nec`-type equation it is the sampled
`b_nec_Intercept`, which is currently held inside the range by the prior
truncation. Remove that truncation, as #393 does, and one half of the mixture
deletes its above-range draws while the other keeps them at large values.

## Evidence

`expand_nec()`, `R/expand_classes.R:111-114`, takes the NSEC off the curve for an
`ecx`-type equation and `b_nec_Intercept` for a `nec`-type one.
`expand_manec()`, `R/expand_classes.R:403`, mixes the per-equation posteriors by
stacking weight through `w_nec_calc()`, and summarises the mixture at `:420`.

`estimates_summary()`, `R/helpers.R`, uses `na.rm = TRUE`, with the recorded
reason from #39 and D15 ruling 3 that an NSEC read off a curve is `NA` where the
[...]

Issue #393, opened 2026-09-19
Remove the tested-range truncation from the default nec and ec50 priors


beckyfisher:
## What

Remove the truncation of the default `nec` and `ec50` priors to the range of the
tested predictor. The prior distribution's own support becomes the only bound.

## Why it matters

A threshold above the highest concentration tested is currently outside the prior
support, not in its tail. The posterior cannot place mass there, so it piles
against the bound and `nec()` returns the highest concentration tested with a
narrow credible interval around it. That reads as a precise estimate of something
the experiment did not measure.

Removing the truncation does not identify such a threshold. What it changes is
the shape of the posterior: mass spreads over the region the data cannot
distinguish and the interval becomes wide, which is the correct statement. The
censoring added in #395 and #392 is what turns that into a report rather than a
number.

## Evidence

`R/define_prior.R:1277-1281` builds both entries with `lb = min(prior_predictor)`
and `ub = max(prior_predictor)`.

#386 measured the consequence prior-only. On an `ecx4param` design with a true
`ec50` of 45 against a highest concentration of 40, the truncated prior CDF at
the truth is 1.000 under both prior sets, as it is for a true `nec` of 60 on the
`nec4param` design. A CDF of exactly 1 means the prior support excludes the
value.

Plan: `notes/tasks/386-incomplete-designs-human.md` phase 5, specification §3.
This must not merge before the censoring of #395 and #392, or `bnec()` begins
reporting concentrations above the highest tested with no qualification, and the
two halves of a model-averaged N(S)EC handle a beyond-range draw in opposite
ways. See #395 for why that second point is not `nec()`'s business.

<details>
<summary>Implementation detail</summary>

### The change

```r
# The distribution predictor_prior() returns is proper on its own support, so
# the posterior stays proper with no truncation. The lower bound is kept only
# where the distribution requires it: a lognormal has zero density below zero,
# and brms would otherwise declare an unconstrained parameter whose every
# negative proposal Stan rejects.
x_lb <- if (grepl("^lognormal", x_pr)) 0 else NA
pr_nec <- prior_string(x_pr, nlpar = "nec", lb = x_lb)
pr_ec50 <- prior_string(x_pr, nlpar = "ec50", lb = x_lb)
```

`predictor_prior()`, `R/define_prior.R:950`, returns `lognormal(mu, sigma)` on a
predictor supplied as a concentration and `normal(mu, sigma)` on one already
[...]

PR #401, opened 2026-09-19
Report every censored no-effect and effect-concentration estimate as censored

beckyfisher:
## What

Every reported no-effect and effect-concentration estimate now states how much
of its posterior lies beyond the range the model was predicted over, and reports
as a bound any entry that falls among those draws. Up to now such a draw was
deleted and the rest were summarised as though nothing had been removed.

The record is written where the posterior is realised, inside `bnec()`, and every
reader takes it from there, so the number `summary()` prints and the number
`nec()`, `nsec()` or `ecx()` returns are the same number qualified the same way.

Closes #395. Phase 4 of #386; the specification is §4.1 to §4.5 and §4.9 to
§4.12 of `notes/tasks/386-incomplete-designs-claude.md`. `extrapolate` is §4.6
and belongs to #392, and is not in this branch.

## Why it matters

The draws that were deleted are those whose curve did not reach the target, which
are the largest. A reported estimate was therefore lower than the quantity it was
labelled as, and its interval was narrower than the posterior supports. Lower is
the protective direction, which is why this has not caused trouble; the narrow
interval is the more serious half, because it is read as precision the fit does
not have.

The N(S)EC is the value most often reported from this package, and it is not
computed by `nec()`. `expand_nec()` realises `ne_posterior` and stores `ne`;
`expand_manec()` realises `w_ne_posterior` and stores `w_ne`; `summary()` reads
those through `clean_nec_vals()` and never calls `nec()`. A censoring report added
to `nec()` alone would not have reached the number a user reports, which is why
the record is written at the point of realisation instead.

The model-averaged N(S)EC was also a mixture of two quantities treating the same
condition in opposite ways. For an `ecx`-type equation the stored draw is an NSEC
read off the grid and is `NA` beyond it; for a `nec`-type equation it is
`b_nec_Intercept`, currently held inside the range by the prior truncation. #393
removes that truncation, at which point one half of the mixture would delete its
beyond-range draws while the other kept them at large values. This branch settles
that before #393 arrives.

#391 raises the value of the report. The pre-fit advisory from #390 misses the
hurdle survival block on 0.90, 0.82 and 0.74 of cases at falls of 0.1, 0.2 and 0.3
of the span (`notes/prior_audit.md`), so a user on the commonest incomplete design
is often not told to look. This report does not depend on that advisory firing.

## Evidence

A censored estimate, on `manec_example`'s `ecx4param` fit re-expanded on a grid
ending at 0.9 rather than at 3.22, 100 draws:

```
     Estimate Q2.5   Q97.5
NSEC  >= 0.90 0.38 >= 0.90
NB: 60 of 100 draws of the NSEC lie above 0.9, the upper bound
    of the prediction range. An entry marked >= is that bound and not a quantile.
```

The median and the upper limit fall among the 60 censored draws and are reported
as the bound. The lower limit falls among the 40 draws that were identified and is
an ordinary quantile of the whole posterior.

The deleted-draw summary of the same posterior is 0.801 (0.048-0.890). It names
0.801 as the median NSEC on a fit where 60 of the 100 draws are known to exceed
0.9, and it names 0.890 as an upper limit although that is the 97.5th percentile
of the 40 draws that reached the target and not of the posterior. The lower limit
is 0.384 here and 0.048 there, because 0.048 is the 2.5th percentile of those 40
draws, which is the 1st percentile of the 100.

### The measured behaviour change

A fit with no draw beyond its prediction range is unaffected and its summary is
unchanged to the last bit. The change was measured across all nine vignettes.
Which estimates each one reports was taken from its parsed chunk headers rather
than from the text of the file, since the vignettes set `eval` globally and a
chunk is skipped only where it sets `eval = FALSE` itself. The fits behind those
estimates were rebuilt under the settings each vignette uses, and both summaries
were taken from one set of draws, so no fit was run twice, once under each
branch. Where `summary()` computes an ECx over a grid a bare `ecx()` call would
not use, both grids were measured.

| vignette | measurements | with a beyond-range draw | how determined |
|---|---|---|---|
| `example1` | 16 | 0 | ten single-equation fits rebuilt |
| `example2` | 17 | 0 | `exp_5` rebuilt, thirteen equations |
| `example2b` | — | — | ruled out: no `bnec()` call in any chunk |
| `example3` | 2 | 0 | two fits rebuilt for the two `nec()` calls |
| `example4` | — | — | ruled out: reports only `compare_posterior()` |
| `example5` | 1 | 0 | the installation-check fit rebuilt |
| `example6` | 99 | 9, over 8 estimates | every hurdle fit rebuilt |
| `example7` | — | — | ruled out: all 12 `bnec()` calls in `eval = FALSE` |
| `example8` | — | — | not measurable, see below |
| `example9` | 14 | 0 | four model sets rebuilt |

No reported entry falls among a censored draw, so no vignette estimate is
reported as a bound. Eight estimates in `example6` change, all of them ECx on a
hurdle fit, where the two blocks are predicted over different grids and growth
is fitted on survivors alone. Two of them change the output `summary()` prints,
which is at two decimal places:

| estimate | beyond the range | old | new |
|---|---|---|---|
| `summary(fit_a, ecx = TRUE)` EC50 growth | 19/3200 | 1.051 (0.7785, 1.234) | 1.052 (0.7776, 1.242) |
| `summary(fit_a, ecx = TRUE)` EC50 survival | 9/3200 | 2.079 (1.460, 2.452) | 2.080 (1.459, 2.456) |
| `summary(zib_f, ecx = TRUE)` EC50 survival | 3/8000 | 6.326 (5.195, 8.088) | 6.326 (5.195, 8.093) |
| endpoints loop, A EC50 growth | 6/3200 | 1.052 (0.7792, 1.238) | 1.052 (0.7778, 1.241) |
| endpoints loop, B N(S)EC growth | 1/3200 | 0.0494 (0.02968, 0.09282) | 0.0494 (0.02960, 0.09287) |
| endpoints loop, B EC50 growth | 1/3200 | 0.07254 (0.05290, 0.09662) | 0.07254 (0.05289, 0.09665) |
| endpoints loop, C EC50 growth | 8/3200 | 3.399 (0.8566, 10.10) | 3.402 (0.8560, 10.42) |
| endpoints loop, D EC50 growth | 3/3199 | 7.576 (5.035, 11.61) | 7.579 (5.021, 11.70) |

Only the first two differ once printed: the upper limit of the growth EC50 from
1.23 to 1.24 and of the survival EC50 from 2.45 to 2.46. The endpoints loop
prints `signif(median, 2)`, and no median changes at that precision. What is new
in each of these places is the note stating how many draws lie beyond the range,
which the release does not report at all.

`example8` could not be measured. It fits 189 models and loads them from a store
keyed by the fit call, which `vignettes/fit_store.R` reaches only where
`BAYESNEC_FIT_STORE` names a directory produced by the
`open-AIMS/grouping-structures` compendium. That variable is unset here and no
such directory exists, and the file's own header puts a sequential rebuild at
the better part of a day, so its 24 reported estimates are unmeasured.

The change shows on a design that has not reached its lower asymptote, which is
the case #386 is about. On the packaged `ecx4param` fit re-expanded over a
prediction range that stops before its curve reaches the reference, 60 of 100
draws are beyond the range, and the reported NSEC is `>= 0.9` (0.384 to
`>= 0.9`) where the deleted-draw summary gave 0.801 (0.048-0.890).

`NEWS.md` states the same figures and the same scope.

<details>

### The record

`censoring_record(upper, lower, above, below)` in `R/helpers.R`, attached as
attribute `"censored"` to the per-draw vector. The two ends are separate because
an `NA` alone does not say which one produced it.

The bounds are the ends of the prediction grid on the **fitted** predictor scale,
taken as the range of `sub_x_transformation(grid)` rather than as `max(grid)`. The
draws are on that scale — `nsec_off_curve()` puts its result there and
`b_nec_Intercept` is a parameter of the fitted model — and a decreasing `crf()`
transformation would otherwise select the wrong end. Non-finite grid entries are
dropped, because an `x_range` reaching 0 under `crf(log(x))` puts `-Inf` at the
foot of the grid.

Written at:

- `expand_nec()`, `R/expand_classes.R` — from `nsec_off_curve()` for an `ecx`-type
  equation, from a comparison against the bounds for a `nec`-type one, and off the
  combination for the two-threshold hurdle branch. `pmin()` of two threshold
  blocks holds no `NA`, so the record is derived from the combined values rather
  than merged from the two blocks' records.
- `expand_manec()` — assembled under the draw index that builds the mixture, so
  the fraction is the weighted one and not a count over equations. Where component
  bounds differ, `concat_censoring()` takes the smallest upper and the largest
  lower, which is the statement true of every component.
- `ecx()` and `nsec()` — built rather than read, because they recompute on a grid
  the caller may have changed through `x_range`.

### The censored summary

`summarise_censored()` places a censored draw at `Inf` or `-Inf` and takes
`quantile(type = 1)`, the inverse empirical distribution function. That is
exactly "it contributes its rank and nothing else": the reported entry is an
order statistic, so it is either a draw that was identified or a draw known to
lie beyond an end, and no finite value is invented for a censored draw. An entry
that comes back infinite is reported as the bound, flagged in a
`"censored_summary"` attribute giving the marks, the counts and the two bounds.

`type = 1` rather than R's default `type = 7` because that default interpolates
between two adjacent order statistics, and an interpolation that starts at a
finite draw and ends at a censored one is infinite for any weight above zero. A
quantile lying only fractionally inside the censored block was therefore
reported as the bound, which asserts more than the ranks support: on ten draws
with one censored at 100, the 97.5 per cent quantile sits 0.775 of the way from
the ninth draw into the tenth, and the smallest value consistent with the sample
is the ninth draw, not 100. The uncensored branch keeps `type = 7` and is
untouched, which is what makes the regression guard exact.

This is not the treatment #39 and D15 ruling 3 removed, and the code comment at
`estimates_summary()` says so: that one assigned `max(x_vec)` to a censored draw
and then used it as a number, which raised the point estimate without saying so.
D15 ruling 3 stands.

The uncensored path is the release's code verbatim rather than the new code with
an empty record. `median()` and `quantile(type = 7)` agree on the 0.5 quantile
algebraically but not necessarily in floating point, and a fit with no
beyond-range draw has to return the release's number to the last bit. That is the
regression guard and it has a test of its own.

### `xform`

A record is built on the scale the curve was searched on, which for `ecx()`,
`nsec()` and `expand_nec()`'s off-curve branch is the recorded predictor scale,
and is then remapped onto the fitted scale and again for any `xform` (§4.8).
`xform_censoring()` does both remappings: it restates the two bounds and, for a
decreasing transformation, swaps which end each censored draw is named at. The
same draws are censored either way.

Building the record against fitted bounds directly, which an earlier revision of
this branch did, leaves `above` and `below` naming the ends of the recorded
scale. Under `crf(-x)` a draw the search left beyond the top of the recorded
range lies below the foot of the fitted one, and was reported as at or above the
fitted top — the one value it is known not to exceed. `nsec.bayesnecfit()`
reached the right answer by accident, because it happened to pass a record whose
`upper` was below its `lower`; all three sites now state it.

### Readers

- `clean_nec_vals()` and `nice_ecx_out()` copy the record onto the matrix, which
  `as.matrix()` drops.
- `print_mat()` marks a censored entry `>=` or `<=`; `print_censoring_note()`
  prints the count and the bound beneath it.
- `bind_nec()` and `bind_ecx()` mark the `autoplot()` annotation, which is
  built from values `to_axis_scale()` has already put on the axis scale.
- `plot.bayesnecfit()` and `plot.bayesmanecfit()` mark their legends, so the two
  plotting paths and `summary()` cannot disagree about one object.
- `print.hurdlesummary()` stacks several estimates into one matrix with
  `rbind()`, which keeps the numbers and drops every attribute, so the records
  are collected by `row_censoring()` and reattached; `print_mat()` marks a
  stacked table row by row.
- `bnec()` reports once per fit through `report_ne_censoring()`, called from
  `expand_manec()` for the set and from `expand_nec()` for a threshold equation
  whose draw is beyond the grid. The per-equation report from `nsec_off_curve()`
  is unchanged in placement and now separates the two ends.

### Defects in the hurdle and ECx routes that the record exposes

`nec.bayesnechurdlefit()` combined its blocks with `pmin()`, which propagates the
`NA` an `ecx`-type block returns. A draw whose growth estimate was above the range
and whose survival estimate was well inside it therefore had no combined value and
was deleted. `combine_censored_min()` enters a beyond-range component draw at the
end it is known to be beyond, so the minimum is above the range only where both
blocks are and below it where either is. Two blocks both inside the range are
unchanged.

`ecx_from_posterior()` now separates the two kinds of `NA` as
`nsec_from_posterior()` already did. A below-range draw requires the target to sit
below the control, which is what distinguishes it from `type = "direct"` with a
response value above the whole curve — the case D15 ruling 3 is about.

### Deliberately not covered

`nsec.brmsfit()` summarises with `quantile()` and no `na.rm`, so a beyond-range
draw raises "missing values and NaN's not allowed" rather than being deleted in
silence. Deletion in silence is the defect this issue is about, and it has no
`bayesnecformula` to take a bound from. The branch does stop it returning the
per-draw `below_range` vector that `nsec_from_posterior()` now adds, which both
`bayesnec` classes already strip.

`average_estimates()` pools posteriors across separately fitted objects, whose
grids and formulas need not agree. `concat_censoring()` would answer the
combined-bound question, so the obstacle is not that one: it is that
`gm_mean()` divides by `length(x)` rather than by the count that is not
missing, so a component whose estimate is `NA` is already averaged in as 1.0.
`bayesnec:::gm_mean(c(4, NA))` returns 2 where the geometric mean of the
identified component alone is 4. That is a defect of its own, older than this
branch, and a censoring record cannot be attached over the top of it.

`compare_posterior()` and `compare_estimates()` take a posterior from `nec()`,
`nsec()` or `ecx()`, thin it, and compute a probability with
`mean(..., na.rm = TRUE)` over the draw pairs in which both estimates are
identified. The record does not survive the thinning, and what such a
comparison should report when one side is censored and the other is not is a
question §4 does not settle: a probability conditional on both being identified
is not the same quantity as one over the whole posterior, and neither is the
quantity the function currently names.

`bayesnecgroupfit`'s per-level estimate table, built by
`group_estimate_table()`, keeps the three numbers and drops the marks, so
`nec()` on a group fit returns unmarked numbers where the same estimate on the
level's own fit is marked. `summary()` on a group fit is unaffected: it prints
each level through the level's own method, and each of those tables is marked.
Adding a column to the per-level table changes a shape callers index by
position, so it is left to #394, which reworks that table for the
incomplete-design set.

`extrapolate` (§4.6) is #392.

### Tests

`tests/testthat/test-censoring.R`, 124 assertions. The rank treatment and that a
censored draw is given no value; both ends; a decreasing `xform`; the weighted
combination against a count over equations; the hurdle minimum with one block
censored; the printed bound and note; the plot annotation; `nsec()` agreeing with
what `summary()` prints for one object; a stored estimate unchanged by
`pull_out()`; and the uncensored regression guard on `manec_example`. The
integration tests re-expand packaged fits on a narrowed grid, so nothing is
compiled or sampled.

`tests/testthat/test-ecx.R:203` is updated: the posterior is still all `NA` and
still not the lowest concentration tested, and the summary now reports the bound
with its count instead of `NA`.

### Definition of done

Every reported no-effect and effect-concentration estimate states its censored
fraction and its bound; the numbers `summary()` prints agree with what `nec()`,
`nsec()` and `ecx()` return for the same object; nothing changes for a fit with no
censored draw; and `NEWS.md` records the measured change to the vignette figures.

</details>


beckyfisher:
## Outcome of the review

Five review cycles. Nine of the points raised were defects in this branch and
are fixed; the rest are answered below. Commits `80c06977`, `489f1356`,
`07c4733e`, `59999095`, `18d94a1a`, `7bb81e6a` and `4ac5c449`.

### Defects fixed

#### A quantile that straddled the censoring boundary

`summarise_censored()` placed a censored draw at `Inf` and took `quantile()` at
R's default `type = 7`, which interpolates between two adjacent order
statistics. An interpolation that starts at a finite draw and ends at a censored
one is infinite for any weight above zero, so a quantile lying only fractionally
inside the censored block came back as the bound. On ten draws with one censored
at 100, the 97.5 per cent quantile sits 0.775 of the way from the ninth draw
into the tenth and was reported as `>= 100`, although the smallest value
consistent with the sample is 9. The censored branch now takes `type = 1`, the
inverse empirical distribution function, so every reported entry is an order
statistic: either a draw that was identified, or a draw known to lie beyond an
end. The uncensored branch keeps `type = 7` and is untouched, which is what
keeps the regression guard exact; the consequence, that every entry shifts a
little once the first draw is censored, is now stated in the roxygen of `nec()`,
`nsec()` and `ecx()` and in `NEWS.md`.

#### The end named under a decreasing predictor transformation

`ecx()`, `nsec()` and `expand_nec()` search the curve on the recorded predictor
scale, so a draw the search did not identify is beyond an end of *that* grid.
The record was built against the fitted bounds instead, which leaves `above` and
`below` naming the ends of the recorded scale: under `crf(-x)` a draw beyond the
top of the recorded range lies below the foot of the fitted one, and was
reported as at or above the fitted top — the one value it is known not to
exceed. Records are now built on the scale the search ran on and remapped with
`xform_censoring()`, which swaps the two ends for a decreasing transformation.
The two `bayesnechurdlefit` sites were converted in the same way.

#### The bound each class of draw was named at

The message text chose its bound with `if (any(cens$above))`, which names the
opposite end where draws sit at both ends at once. `xform_censoring()` now
records a cumulative `"swapped"` flag, `censored_end()` reads it, and all six
message sites call it.

#### A warning and a note that contradicted each other

`warn_censored_draws()` keyed its wording on the geometry the search saw while
`print_censoring_note()` keyed its own on the record's fields, so a reversed
predictor had the two disagree about the same draws: "censored above -10"
printed over "lie below -10". Both now read the same two fields.

#### A draw the record explains at neither end

Such a draw could not be computed at all rather than being known to lie beyond a
bound. It is now reported on its own and left out of the denominator the other
two reports use, which is the denominator the note states, so every fraction
printed is over one sample.

#### A censored estimate printed unmarked on a hurdle summary

`print.hurdlesummary()` stacks several estimates into one matrix with `rbind()`,
which keeps the numbers and drops every attribute, and `print_mat()` marked only
a one-row matrix. A censored estimate therefore printed there as a bare number
while `nec()` on the same object returned it marked. `row_censoring()` collects
the per-row records and `print_mat()` marks a stacked table row by row, with one
note per censored row.

#### The grid the ECx block was computed over

`summary(x, ecx = TRUE)` called `ecx()` with no `x_range`, so the ECx was
rebuilt from the data while the no-effect estimate printed above it came from
the fit's own grid. With the marks this branch adds the two claims sit on one
screen: a note reading that the prediction range stops at 0.9 stood directly
above an unmarked ECx of 1.67. All three summary methods now pass the grid the
fit was predicted over. For a two-block fit that is the intersection of the two
component grids, which is the bound `combine_censored_min()` gives the combined
estimate.

#### A crash on the commonest hurdle summary

The first version of that fix read `pred_vals`, which is `NULL` on a
`bayesmanecfit`, so `range(c(NULL, NULL))` was `c(Inf, -Inf)` and
`summary(hurdle_fit, ecx = TRUE)` errored. `bnec_hurdle()` takes its model set
from `crf()`, so both components are routinely model-averaged and this was the
ordinary path. `hurdle_summary_range()` reads whichever slot the component class
uses, and has a test.

#### Dead code behind the plot annotation

`rescale_censoring_bounds()` was unread: `bind_nec()` and `bind_ecx()` build the
annotation from values `to_axis_scale()` has already rescaled. It and its four
call sites are removed. `bind_nec()` also lost an `xform` argument that was
always identity and would have mismatched the marks against the values, and
`nsec.brmsfit()` no longer returns the per-draw `below_range` vector that
`nsec_from_posterior()` adds for the record.

### Answered without a change

#### A target the curve declines away from

`ecx(type = "direct")` with a response value the curve never attains, and an absolute target above a control
that is itself negative, have the same geometry: there is no such concentration
at any predictor value. Specification 4.5 reports every draw the search did not
identify at the upper end, so both are reported there, where 2.1.3 returned
`NA`. This is the one case where the new report is less precise than the old and
the `ecx()` roxygen now says so.

#### The estimators left out of scope

`average_estimates()`, `compare_posterior()`, `compare_estimates()` and
`bayesnecgroupfit`'s per-level estimate table are named in the PR body, each
with the mechanism that keeps it out.

#### Residues that need a hand-assembled object to reach

`concat_censoring()` takes the first component's `"swapped"` flag and
`combine_censored_min()` takes the logical OR of the two; within one `bnec()`
call a single formula drives every component, so they cannot differ. And
`summary.bayesnechurdlefit()` passes `x_range` explicitly while still forwarding
`...`, so `summary(fit, ecx = TRUE, x_range = )` would be refused for a
duplicate argument — a call that was already incoherent, since
`nec.bayesnechurdlefit()` has no such formal. Both are worth an issue rather
than a change here.

### The conditions on merging

A fit with no beyond-range draw reports exactly what the release reported. On
the packaged fits, `ne`, `w_ne`, the stored posterior, `nec()`, `nsec()`,
`ecx()`, `summary()` and `print()` are identical between `eff06ea9` and this
branch.

The behaviour change is measured in `NEWS.md`: nil on the two vignette fits
rebuilt, and on the packaged `ecx4param` fit re-expanded over a range that stops
before its curve reaches the reference, 60 of 100 draws beyond the range with
the NSEC reported as `>= 0.9` (0.384 to `>= 0.9`) against a deleted-draw summary
of 0.801 (0.048-0.890). `example9`'s `fit_reduced` was started for the
transformed-predictor case and stopped after several hours without completing
its model set, so the measurement covers two vignettes rather than nine and says
so. Whether that meets specification 4.12 is @beckyfisher's call.

beckyfisher:
The body has been updated: *The measured behaviour change* now covers all nine
vignettes rather than the two it named before, and `NEWS.md` states the same
figures in commit 9f5aaf03. Nothing else in the body changed, and no code
changed with it.

What the measurement found, in short. 149 measurements over the estimates the
six vignettes that report one actually print; `example2b`, `example4` and
`example7` report none and are ruled out by inspection, with what was inspected
stated in the body; `example8` is not measurable without the
`grouping-structures` fit store and is named as such. No vignette estimate is
reported as a bound. Eight estimates change, all ECx on a hurdle fit in
`example6`, and only two of them differ once printed at the two decimal places
`summary()` uses.

PR #405, opened 2026-09-22
Add extrapolate to nec() and nsec(), reading the censoring record

beckyfisher:
## What

`nec()` and `nsec()` gain `extrapolate`, which chooses the bound the estimate is
reported as censored at. `FALSE`, the default, keeps the prediction range the
fit was built on and changes nothing. A single number is an upper limit, a pair
is a lower and an upper limit, and `TRUE` removes the bound at both ends.

## Why it matters

Once the tested-range truncation is removed (#393), a threshold beyond the
tested range is inside the prior support and the posterior can place mass there.
A user who wants that value needs a way to ask for it by name, and a user who
does not must keep getting an estimate inside the range they measured.

Both ends are in scope. #393 removes `lb` and `ub` in the same two lines, so a
threshold below the lowest concentration tested becomes admissible at the same
moment.

The two halves of a model-averaged N(S)EC cannot be extended the same way. A NEC
is a sampled parameter with no grid behind it, so releasing it is a comparison
and nothing more. An NSEC is read off a fitted curve, so extending it means
evaluating that curve further out, and no curve can be evaluated on an infinite
grid. `TRUE` is therefore accepted only where every component samples a NEC, and
on the default `bnec()` model set, which fits equations of both classes, it is
an error naming the finite form. A finite limit applies to both classes, by a
different route for each.

## Evidence

Measured on the packaged `manec_example` fits, re-expanded over a prediction
grid cut to 0.9 so that the stored estimate is censored. No model was compiled
or sampled.

On `nec4param` alone, every draw of the threshold posterior lies above the cut,
so the default reports `>= 0.9` at all three entries and states 100 of 100 draws
censored. `extrapolate = TRUE` returns 1.461 (1.361 to 1.528), which is the
summary the same fit gives over its full grid, to the last bit.
`extrapolate = 1.45` censors 63 of the 100 and reports the bound at 1.45.

On the mixed `nec4param` plus `ecx4param` set over the same cut grid,
`extrapolate = TRUE` is refused with an error naming `ecx4param` and the finite
form. `extrapolate = 3` releases the threshold half by comparison, re-evaluates
the smooth half out to 3, and reports 1.453 (0.723 to 1.528) with nothing
censored, against `>= 0.9` (0.709 to `>= 0.9`) by default.

`nsec(ecx4param, x_range = c(0.0324, 0.9))` censors draws above 0.9; adding
`extrapolate = 3` re-evaluates the curve out to 3 and the fraction falls to
none.

The re-evaluation route was checked against the stored posterior:
`nsec(ecx4param, resolution = 1000, posterior = TRUE)` reproduces
`manec_example$mod_fits$ecx4param`'s stored `ne_posterior` exactly at the grid
that built it.

Plan: `notes/tasks/386-incomplete-designs-human.md` phase 5, specification
§4.6.

Depends on #395 (PR #401), which had not merged into `predev` when this branch
was cut, so `issue-395-censored-estimates` is merged into it. Review the diff
against that branch rather than against `predev`.

<details>

### The new files

`R/extrapolate.R` is new and holds the machinery. `R/nec.R`, `R/nsec.R`,
`nec.bayesnechurdlefit()` and `nsec.bayesnechurdlefit()` wire it in.
`tests/testthat/test-extrapolate.R` is new, 83 assertions, and fits nothing.
`warn_censored_draws()` in `R/helpers.R` gains `range_label`, so a report of a
bound the caller named does not call it the end of the prediction range.

### The scale `extrapolate` is read on

The recorded predictor scale: the scale the data were supplied on and the scale
`x_range` takes, not the fitted scale the estimates are returned on. The
alternative was the fitted scale, which §4.8 points at by requiring the
comparison to happen there. The recorded scale fails safe. Under an increasing
`crf()`, a fitted-scale number passed where a recorded one is wanted is smaller
than the top of the prediction range and is caught by the "inside the range"
check; a recorded-scale number passed where a fitted one is wanted is larger and
would be accepted in silence.

The comparison is still made on the fitted scale.
`fitted_extrapolate_limits()` transforms the pair and then orders it, so a
decreasing `crf()` does not keep the names of the two ends with it, and an
infinite recorded limit stays infinite at the end the direction puts it. A limit
outside the domain of the transformation — 0 under `crf(log(x))` — is an error
rather than `-Inf` at the foot of a grid no curve can be read on.

### The routes a finite limit takes

A threshold component is released by recomputing `above` and `below` against the
new bounds. No curve is evaluated and no posterior is rebuilt.

A curve-read component is re-evaluated through `nsec(x_range = c(lower, upper),
posterior = TRUE)`, on the component classed for that call rather than through
`pull_out()`, which pays for a discarded `expand_nec()` per equation. The
resolution reuses the number of points in the grid the fit stored rather than
`nsec()`'s lighter default of 200, so over a wider range the grid is
proportionally coarser; `sig_val` and `resolution` can be passed through `...`.

Two requests re-evaluate nothing and return the stored posterior with its stored
record: a limit equal to the current bounds, and any limit on a fit with no draw
beyond either end, where every draw was identified inside the range the fit used
and a wider grid has none left to identify.

### The range a limit is measured against

The wider of two: the range the call will search, and the prediction range the
fit stores. They are the same for a fit built with no `x_range` of its own. They
differ for a fit built over a grid above the data, where `nsec()` with no
`x_range` returns to the observed range, and a limit between the two would
report the estimate as censored inside the grid the fit was built on. Taking the
wider of the two makes `nec()` and `nsec()` accept and refuse the same numbers,
which is what specification §4.6 asks for.

`ne_grid_bounds()` keeps a grid point only where it is finite on the fitted
scale as well as the recorded one, which is the filter `expand_nec()` applies to
its own bounds. Filtering the recorded grid alone keeps a zero under
`crf(log(x))`, and the lower limit of a single-number `extrapolate` defaults to
that bound, so such a fit refused every upper limit for naming a value the
caller had not given. It returns `NULL` where no grid is stored, and it is left
as a promise in both `nec()` methods, so a fit without `pred_vals` reaches
nothing on the default path.

### The model-averaged path

The threshold components are sliced straight out of `w_ne_posterior` by
position, which is free and exact, but only where `w_draw_index` is stored.
`manec_example` predates #216 and has an empty index, so `pull_draw_index()`
regenerates a different draw and slicing would put stored draws beside
recomputed ones taken under a different index: measured, the mixed lower bound
at the stored bounds read 0.81 against the stored 0.75. Where the index is not
stored, every component is rebuilt under the one regenerated index, which is
what `nsec.bayesmanecfit()` and `ecx.bayesmanecfit()` already do.

`manec_ne_types()` derives each component's class from the equation name, which
is how `expand_manec()` derives the label it puts on the set, so `nec()` and
`summary()` cannot disagree about what the set is. For a set of joint two-block
fits that name describes the response block alone, so `check_releasable()`
refuses on the posterior instead: a draw with no value cannot be released by
comparison, and the error says to read it with `nsec()`.

### Constraints reported rather than corrected

A `nec` prior bounded at or below the limit asked for means the posterior holds
no draw beyond it, so a wider limit returns the same truncated posterior. The
bound is read off the prior stored on the fit, so a user-supplied bound is
handled (specification §2.3). `brms` stores that bound as the character form of
the number at fifteen significant digits, so one taken from the grid parses back
a few units in the last place either side of it; both comparisons use a
tolerance, without which the exact test reported a lower bound that was not
reached and stayed silent about the upper bound that was.

A lower limit does not extend a curve-read search. `nsec_from_posterior()`
begins at the control, because the reference is a quantile of the control
posterior, and a draw already past the reference there has no identifiable
crossing at any lower concentration. Such a limit is accepted and a message
states that those components stay censored at the control, rather than being
refused: the threshold components of a mixed set are genuinely released to it,
and `concat_censoring()` already reports the bound true of every component.

### Refusals

- A limit inside the current range, at either end, naming `x_range`.
- `TRUE` where any component is curve-read, naming the finite form and the
  equations.
- `TRUE` on `nsec()` for every object, because every NSEC is read off a curve.
- `extrapolate` on `nsec.brmsfit()` and `nsec.drc()`, because neither stores a
  prediction range to measure a limit against. Both gained the argument so that
  the generic's is not silently absorbed into `...`.
- A pair given upper first, a pair naming one value at both ends, a length
  other than one or two, `NA`, and a non-numeric.
- A limit on an object that stores no prediction range at all.
- A draw with no value on a path that releases by comparison, which is the one
  case `manec_ne_types()` cannot see from the equation name.

### Unchanged behaviour

`ecx()` is unchanged. The censoring record, the censored summary and every
message #395 added are unchanged. Every call that does not name `extrapolate`
returns what it returned before, which the first integration test pins against
`manec_example`'s stored summary.

### Definition of done

`extrapolate = TRUE` returns the posterior unaltered on a pure NEC set and
errors on a mixed one; a finite limit extends both halves; a limit inside the
range is refused; and no output changes for a fit with no censored draw.

</details>


beckyfisher:
## The review cycle

Three rounds against an independent reviewer, closed by agreement. This comment
records the findings that were real and what each one changed. Commits
`25a2a3ae`, `b51908b6`, `15e5a4a9` and `2987c1e6`.

### Findings that produced a wrong number

1. The grid bounds were filtered on the wrong scale. `ne_grid_bounds()` dropped
   non-finite grid points on the recorded predictor scale, where `expand_nec()`
   drops them on the fitted one. A zero is finite on the recorded scale, so on a
   fit whose grid reaches zero under `crf(log(x))` the lower bound stayed at
   zero, and since a single-number `extrapolate` defaults its lower limit to
   that bound, every upper limit was refused for naming a value the caller had
   not given. The filter now requires a point to be finite on both scales.

2. A model-averaged set with no stored draw index was resampled on a call that
   should have changed nothing. `extrapolate` rebuilt the mixture through
   `pull_draw_index()`, which regenerates a different draw for an object saved
   before #216. On `manec_example`, which has nothing censored,
   `extrapolate = 5` returned 1.449 (0.808, 1.528) against a default of 1.450
   (0.749, 1.527) — a lower bound eight per cent away, on a fit where an
   extrapolation cannot report anything different. Both estimators now return
   the stored posterior unchanged where no draw lies beyond either end: every
   draw was identified inside the range the fit used, a wider bound censors none
   of them, and a wider grid has none left to identify.

3. `nsec()` and `nec()` measured a limit against different ranges. `nsec()` used
   the range it would search, which for a fit built over a grid above the data
   is the observed range and is narrower than the grid the fit stores. On a fit
   re-expanded to 8, `nsec(fit, extrapolate = 5)` was accepted and censored the
   estimate at 5, inside the fit's own grid, which is the silent tightening the
   argument exists to refuse. With no `x_range` the limit is now measured
   against the wider of the two.

4. A supplied `x_range` was then discarded. This was introduced by the fix to
   item 3 and caught in the second round:
   `nsec(x_range = c(0.5, 0.9), extrapolate = 9)` searched from 0.0324 rather
   than 0.5 without saying so, and `extrapolate = c(0.5, 9)`, the form that
   states the intent, was refused for naming a lower limit inside a range the
   caller had already narrowed. A named `x_range` is now the range at both ends,
   and the union in item 3 applies only where none was named.

5. A hurdle component's errors were all relabelled. The wrapper that names which
   component refused a limit was unconditional, so a default call naming no
   `extrapolate` at all reported "the growth component refused extrapolate: nec
   is not a parameter in ecx model types". It is now gated on the argument.

### Findings that produced misleading output

6. The censoring report named the wrong range. Under `extrapolate = 1.45` on a
   fit whose prediction range ends at 0.9, the warning read "1.45, the upper
   bound of the prediction range". `warn_censored_draws()` gained `range_label`,
   and the posterior is marked only by the branches that replace the record, so
   a limit equal to the current bounds still says "prediction range".

7. The truncated-prior message advised a refit on a call that recomputed
   nothing. It ran before the short-circuit in item 2, so
   `nec(manec_example, extrapolate = 5)` returned the default bit for bit and
   still said the `nec` prior was bounded and to refit with a wider one. It now
   runs after.

8. `extrapolate = FALSE` was not inert. `ne_grid_bounds()` was assigned eagerly,
   so a fit with no `pred_vals` raised two `min()`/`max()` warnings at each end
   of a call that had not named the argument. It is now a promise, forced only
   where the limits need it.

9. Two documentation claims were false. The roxygen said `TRUE` is gated on
   `ne_type`, which holds for a single fit and not for a set, where the gate
   reads the equation name. And the claim that `nec()` and `nsec()` refuse the
   same numbers fails with an `x_range` given, and also where the observed range
   is wider than the stored grid. Both are restated to what the code does.

### The cost finding

10. Each curve-read component paid a discarded `expand_nec()`. The re-evaluation
    reached the component through `pull_out()`, which rebuilt a prediction grid,
    a per-draw root search, `dispersion()` and the `loo` weights, all at the
    default grid rather than the one asked for, and discarded every one.
    Measured on the two-equation packaged fixture: 1.37 s of the 3.39 s a finite
    limit took. The component is now classed for the call instead, and the
    threshold rebuild route reads `b_nec_Intercept` off the fit directly. A test
    pins the classed component against `pull_out()` on the draws, the record and
    every attribute.

### Raised and not acted on

`extrapolate` sits before `which` in `nec.bayesnechurdlefit()`, so a positional
call binds differently than before. R's S3 method-consistency check requires the
generic's formals in the generic's order, so there is no other placement, and
the call fails loudly rather than silently.

A limit is not resolved once against the union of a hurdle fit's two component
ranges. The growth component's range can end below the survival component's, so
a limit between them is refused by one of them; the refusal names which, and the
asymmetry is in `@details`. Resolving it centrally would add arithmetic no
fixture can exercise.

An object fitted before this version records nothing about which of its draws
lie beyond the range, so `extrapolate` does nothing to it. Stated in the roxygen
and in `NEWS.md` rather than worked around.

### Coverage

`tests/testthat/test-extrapolate.R` is 103 assertions and fits nothing. The
branches the first round found untested — a pure-NEC set, the refusal of a draw
with no value, the rebuild under a regenerated index, and the short-circuit —
are all exercised, along with every finding above.

PR #406, opened 2026-09-22
Bound the nec and ec50 priors by their own support, not by the tested range

beckyfisher:
## What

The default `nec` and `ec50` priors are no longer truncated to the range of the
tested predictor. The bound now comes from the support of the prior distribution
instead: `lb = 0` on the `lognormal` branch, which `predictor_prior()` returns
for a predictor supplied as a recorded concentration, and no bound at all on the
`normal` branch, which it returns for a predictor already logged.

Closes #393. Phase 6 of #386.

#405 has since merged, so this now diffs cleanly against `predev` and the
"Files changed" tab shows this change alone. The branch was cut from `predev` at
`d6efdd8a` with `issue-392-extrapolate` merged in at `f16a7174`, because #393's
definition of done depends on the censoring of #395 and the `extrapolate`
argument of #392; `predev` at `2bb1f21a` has been merged back in since.

Twenty files. The package change is five lines of `R/define_prior.R`; the rest
is documentation, the tests that pinned the old bounds, and the measurement.

Three of the files are outside what specification §3 describes, and all three
are prose:

- `R/check_models.R` and `R/models.R` both said "`nec` is bounded to the
  predictor range, so the sampler is free to propose such a value" as the
  premise for excluding `nechormepwr` and `nechorme4pwr` from the 0-1 bounded
  families. That premise is a direct statement about the prior this change
  alters. The conclusion is untouched, and so is every function body: the
  equations `check_models()` drops are identical, and the only non-roxygen line
  either file changes is the sentence inside `unscaled_power_message()`.
- `vignettes/example3.Rmd.orig` is the vignette that documents the default
  priors, and it stated the truncation in six places. Seven prose lines change
  and no code chunk does.

None of the three is an improvement made in passing. Each would be a false
statement about `bayesnec`'s behaviour if this merged without it.

The `.Rmd.orig` changed and `vignettes/example3.Rmd` was not regenerated.
Precompiling refits every model and is a release operation. That is the
repository's normal state between a source edit and a release and is not a merge
precondition, but the exposure is larger than stale prose: the committed `.Rmd`
*demonstrates* the truncation as live output. Its stored `pull_prior()` tables
give the `nec` row `lb = 0.03234801324009`, which is `min(nec_data$x)`, and
`lb = 0, ub = 10` on a second fit, where the current code gives `0` and `NA`. A
user comparing the published vignette against their own `get_priors()` output
will see a discrepancy. Run `vignettes/precompile.R` for `example3` before the
release that includes this. Recorded under "Outstanding at merge" in
`notes/prior_audit.md` so it does not live only in this thread.

## Why it matters

A threshold above the highest concentration tested was outside the prior
support, not in its tail. The posterior could place no mass there, so it piled
against the bound and `nec()` returned the highest concentration tested with a
narrow credible interval around it. That reads as a precise estimate of
something the experiment did not measure.

Removing the truncation does not identify such a threshold. What it changes is
the shape of the posterior: mass spreads over the region the data cannot
distinguish and the interval widens, which is the correct statement. What holds
a reported estimate inside the tested range by default is now the censoring of
#395 and the `extrapolate` argument of #392, not the prior. That is why this
sits on top of both.

The lower bound goes with it, so a threshold below the lowest concentration
tested is admissible as well. On a predictor recorded as a concentration with a
zero control, `min(x)` was zero and the `lognormal` support is zero, so nothing
changes there; it binds on a logged predictor and on a design with no control.

How far into the tail a beyond-range threshold now sits is a property of the
branch and of `prior_type` together, and the four cells differ enough to state
here. Three of them set the spread by a coverage rule that reaches the farthest
concentration tested, so the mass above the highest concentration is 0.025 or
0.01 at least and larger where the series extends further below its median on
the log scale than above it. The fourth, `"uninformative"` on a predictor
supplied already logged, sets the spread to 10 times the standard deviation of
the *distinct* predictor values, which is the constant Fisher et al. (2024)
state rather than a coverage width, and is much the widest. Measured as the
share of the entry's mass between the lowest and the highest value tested, and
the share above the highest, on `nec_data` and on
`log(herbicide$concentration)`:

| predictor | `prior_type` | inside | above |
|---|---|---|---|
| recorded | `"uninformative"` | 0.755 | 0.220 |
| recorded | `"regularizing"` | 0.811 | 0.179 |
| logged | `"uninformative"` | 0.116 | 0.442 |
| logged | `"regularizing"` | 0.980 | 0.010 |

So removing the truncation does much more in one cell than in the other three,
and a design that has not reached its asymptote, on a predictor supplied
already logged, under `"uninformative"`, needs a prior supplied through `prior`.
Issue #393 and specification §3.2 both state the "one part in forty" figure
without a qualifier; it belongs to the coverage-rule cells. The table is written
out once, under `prior_type` in `?bnec`, and `NEWS.md`, the flatness report and
the vignette point there rather than restating it.

### The superseded acceptance criterion

Issue #393 asks for "the initial-value search taking no more proposals than
before on the complete designs". That is not met: the search takes more, by a
paired mean of 1.31 proposals [0.59, 2.04] on `nec4param`, and the measurement
is below. Specification §3.4, which is the authority for this phase, asks
instead that "a rise there must be quantified rather than assumed absent", and
that is what this pull request does. The rise is small, no complete design
reached even the audit's own 200-round cap, which is a fiftieth of the 10,000
`make_good_inits()` takes in a fit, and the number of cells that exhaust the
audit's cap falls. If the stricter reading is preferred, this needs a change to
`make_good_inits()` rather than to the prior, and that is a separate issue.

## Evidence

`notes/scripts/prior_audit.R` parts `sweep` and `reproduction`, run twice on
2026-09-22 on this branch, at the commit before and the commit after the change
to `R/define_prior.R`. R 4.6.1, brms 2.23.0. The script is unchanged between the
two runs and every cell seeds itself from its own index, so the 5,760 cells are
paired. The before run reproduces the 2026-09-19 figures in
`notes/prior_audit.md` part 3 exactly, which is what establishes that it
measured the released code. Full tables are in `notes/prior_audit.md` part 4.

### The prior support on an incomplete design

Of the 2,159 incomplete `ec50` cells that could be built, 1,439 had `p_truth` of
exactly 1 before and none do after. The largest remaining value is 0.99938, so the truth is
in the far upper tail and is no longer outside the support.

### The `nec` exclusion on #386's own design

The sweep cannot produce it: its completeness axis keeps a `nec4param` threshold inside the series, which is the
correction #391 recorded in its amendment to §6. Part 2 of the script measures
it on #386's design, a log-spaced series of eight concentrations from 0 to 40
with a true `ec50` of 45 and a true `nec` of 60.

| parameter | prior type | `ub` before | `p_truth` before | `ub` after | `p_truth` after |
|---|---|---|---|---|---|
| `ec50` | `uninformative` | 40 | 1 | none | 0.9795 |
| `ec50` | `regularizing` | 40 | 1 | none | 0.9923 |
| `nec` | `uninformative` | 40 | 1 | none | 0.9877 |
| `nec` | `regularizing` | 40 | 1 | none | 0.9962 |

### Complete-design cells

`top` and `bot` are untouched: every prior string and every `p_truth` is
identical over all 11,502 such rows. The `nec` and `ec50` prior strings are
identical on the complete cells too, over all 1,440. The *statistic* does
change, and §3.4's expectation that it does not is wrong for a reason worth
recording: `p_truth` is the CDF of the prior **after** truncation, so it is a
function of the bounds as well as of the distribution, and the truncation bound
the normalisation on every design rather than only on an incomplete one. The
fall is 0.0008 to 0.344, median 0.055, in every one of the 1,440 cells. The
property §3.4 was reaching for — that a complete design's truth is still well
inside the prior — holds: the count of complete `nec` and `ec50` cells outside
the central 95% of their own prior is 0 before and 0 after.

The sign of that fall is not guaranteed by algebra, only observed. Writing *a*
and *b* for the prior CDF at the old lower and upper bounds and *u* for the CDF
at the truth, the change is −(*u*(1 − *b*) − *a*(1 − *u*)) / (*b* − *a*), which
is negative for every *u* only where *a* = 0. That holds on the `lognormal`
branch with a zero control. On the `crf(log(x))` cells, where the largest falls
are, the old lower bound was binding too and the two terms oppose each other;
the fall there is the net. `notes/prior_audit.md` part 4 states this.

### The initial-value search

It takes more proposals on the complete designs, and the rise is confined to
`nec4param`. Paired by cell, one search per cell, capped at 200 rounds as part 3
was.

| equation, complete designs | median before | median after | mean before | mean after | paired mean difference |
|---|---|---|---|---|---|
| `nec4param`, n = 720 | 6 | 7 | 10.37 | 11.68 | +1.31 [0.59, 2.04] |
| `ecx4param`, n = 720 | 5 | 4 | 7.36 | 7.14 | −0.22 [−0.60, 0.16] |
| both, n = 1,440 | 5 | 6 | 8.86 | 9.41 | +0.55 [0.14, 0.96] |

An `ec50` drawn above the highest concentration still gives a curve that
declines across the series, so `check_init_predictions()` often accepts it; a
`nec` drawn above it gives a curve flat at `top` at every dose, which the band
rejects wherever the response has fallen. The rise does not reach the cap. No
complete-design cell fell back to Stan's own initialisation in either run, the
largest single count over the complete cells fell from 283 to 166, and over the
whole sweep the fallback count fell from 27 cells to 21. Those two counts are
at the audit's 200-round cap, not the 10,000 a fit takes, so they bound from
above how often a shipped fit reaches the fallback and are not a rate for it.

The per-chain acceptance rate was measured directly on one design as well, over
8,000 draws of `nec3param` initial values against the `alga` `c_proliferum`
contaminant A series: 0.188 with the `nec` prior truncated and 0.171 without it,
with 0.101 of that prior's mass now above the highest dose. The shipped search
fills one chain slot at a time, so it pays that rate once; the four-at-once rule
that #309 replaced pays its fourth power, which is why the comparison test in
`test-inits_functions.R` needed a larger cap.

### #392's bounded-prior messages

They stop firing on a new fit, which is intended and needed no code change. A
`nec3param` fit of `nec_data` under `Beta(link = "identity")` on this branch stores `lb = 0` and no `ub`,
`ne_prior_bounds()` returns `list(lb = 0, ub = NA)`, and
`nec(fit, extrapolate = 10)` emits no "bounded above" message.
`report_bounded_ne_prior()` reads the prior off the fitted object, so a fit
stored before this change still reports its bound, and the "bounded below"
message can still fire where a user asks for a limit below zero on a
concentration-scale predictor, which is a real bound.

<details>

### The change

`R/define_prior.R`, the two lines that built both entries with
`lb = min(prior_predictor), ub = max(prior_predictor)`:

```r
x_lb <- if (grepl("^lognormal", x_pr)) 0 else NA
pr_nec <- prior_string(x_pr, nlpar = "nec", lb = x_lb)
pr_ec50 <- prior_string(x_pr, nlpar = "ec50", lb = x_lb)
```

The branch test is on the distribution and not on the sign of the data. A
`lognormal` has zero density below zero, so without `lb = 0` brms declares an
unconstrained parameter whose every negative proposal Stan rejects. A `normal`
is supported on the whole line, and a concentration the user has already logged
may legitimately be negative, so it takes no bound.

`prior_string(lb = NA)` writes `NA` into the `lb` column, which is what
`define_prior()` already writes for an absent bound elsewhere and what
`blank_bounds_to_na()` normalises a fit's `""` to. `make_inits()` reads that
column: with `lb = 0` alone it enters the one-bound branch and redraws while
`draw <= 0`, which a lognormal draw never satisfies, so the bound itself draws
no extra proposals. The rise measured above comes from the widened draws reaching
`check_init_predictions()`, not from the redraw loop.

### Documentation

Three places asserted the truncation and are now wrong:

- `R/bnec.R`, the `prior_type` roxygen, said the threshold prior "excludes a
  threshold above the tested range". It now states where the mass is instead,
  by branch and `prior_type`.
- `R/models.R` and `man/models.Rd`, the `nechormepwr` and `nechorme4pwr`
  exclusion, said "Since \"nec\" is bounded to the predictor range, every such
  value is one the sampler is free to propose". Exported, and the most visible
  of these.
- `R/check_data.R`, both the docstring for the flatness report and the message
  that report prints, gave the truncation as one of three reasons it exists.
  The report still stands, for a different reason: a tail is not a statement
  that the design measured the threshold.
- `R/check_models.R`, the docstring and the message for the `nechormepwr` and
  `nechorme4pwr` exclusion, said "`nec` is bounded to the predictor range, so
  the sampler is free to propose such a value". The conclusion is unaffected and
  the premise is now the wrong way round: an unbounded `nec` makes the excluded
  case more reachable.
- `R/define_prior.R`, eleven passages of `predictor_prior()`'s and
  `define_group_prior()`'s own roxygen, which is where a maintainer looks.
  Those that asserted the truncation are corrected; those that reported a CDF
  computed under it are attributed to the entry as it was when the figure was
  measured, since the figures themselves are sound and are not recomputed here.
- `R/expand_classes.R` and `R/nec.R`, two comments from #395 and #392 that
  described this change as pending.
- `vignettes/example3.Rmd.orig` documents the default priors and stated the
  truncation in six places. Corrected there rather than left to go stale. The
  committed `example3.Rmd` is not regenerated, since precompiling refits every
  model and is a release operation; it needs to run before the release that
  includes this.

One measurement in the issue and in §3.2 needed correcting while doing this.
"Roughly one part in forty of the prior mass sits above the highest
concentration" holds only on the `lognormal` branch, and there only where the
highest concentration is the *farther* of the two ends from the prior location
on the log scale, because `sigma` is set from the larger half-width. Where the
series extends further below its median than above it the share is larger: 0.22
on `nec_data`. On the `normal` branch it is 0.44 on
`log(herbicide$concentration)`. `NEWS.md`, the roxygen and the vignette now
state both branches.

### Tests

Five assertions pinned `lb = min(x)` and `ub = max(x)` and are rewritten to
assert the new rule, in `test-define_prior.R` (two), `test-get_priors.R` (two)
and `test-fit_bayesnec.R` (one). One more, in `test-get_priors.R`, asserted only
that the `nec` bounds on a *stored* fit are not `NA`; that fixture predates the
change, so the assertion still passed while meaning something it no longer
means. It now compares against the bounds the fixture holds, put through
`blank_bounds_to_na()` first because a `brmsfit` writes an absent bound as `""`
and `usable_prior()` rewrites it to `NA`, which is the difference that test
(#141) exists to assert. A guard on the row count and on the bounds not all
being `NA` keeps it from becoming vacuous once the fixture is rebuilt.

Two more failed under the change for reasons of their own.
`test-hurdle_family.R` pinned both threshold bounds at `min(x)` and `max(x)` on
a predictor spanning negative values; it now asserts the property that test
exists for, which is that the two blocks build one entry from the whole
predictor, and that the entry is unbounded on that branch.
`test-inits_functions.R` capped its four-at-once comparison at 2,000 rounds,
which the lower per-chain rate now exceeds on one of its three seeds. The cap is
20,000, which leaves the chance of a capped seed below 1e-7. The three seeds
took 8.1 s, timed with `system.time()` in one `Rscript` process on one core;
not measured on CI, where the suite runs two workers. Raising the guard keeps
the comparison like for
like; the assertions the test exists for, the ratio and the cap on the shipped
search, are unchanged.

Four tests added at the end of `test-define_prior.R`:

- both entries on a concentration-scale predictor are given `lb = 0` and no `ub`,
  under both prior types and both equations;
- a logged predictor is given neither bound, reached both through the `auto`
  sign rule and through `predictor_scale = "log"`, so the branch is tested by
  the route as well as by the outcome;
- the prior CDF at a true `ec50` of 45 against a highest concentration of 40 is
  below 1, computed over the bounds the entry is given and stated beside the
  same quantity under the truncation, which is 1. The bound assertions are what
  make it load-bearing: the prior *string* is the same on both sides of this
  change, so a test reading the string alone would pass either way;
- a predictor with one distinct value builds and draws. The old two-bound
  branch of `make_inits()` ran `while (draw <= min(bounds) | draw >= max(bounds))`
  with `min == max`, which is a tautology, so that design hung the search rather
  than failing. With `lb = 0` and no `ub` the branch is unreachable from the
  default priors, and this pins it.

One assertion is added to `test-check_response_flattened.R`, which already pins
sixteen fragments of the message `check_response_flattened()` prints. The
sentence that message makes about the threshold priors was the only fragment
not pinned, and it is the one that went stale: it said they "exclude a threshold
above the tested range" and nothing failed when that stopped being true.

Full suite: `FAIL 0 | WARN 19 | SKIP 4 | PASS 5061` on `ubuntu-latest
(release)`, with `checking tests` taking 25 minutes. That is the four-platform
`R CMD check` matrix, which does run the suite under `NOT_CRAN=true` —
`r-lib/actions/check-r-package` sets it, although `R-CMD-check.yaml` does not,
and the 4 skips out of 5,065 confirm it. Every file this branch changes was also
run locally on its own under `NOT_CRAN=true` against an installed build, all
green.

Not reported: a local parallel full-suite run. Two attempts deadlocked, which I
confirmed rather than assumed — the stalled worker had accumulated zero seconds
of CPU time in 64 minutes while blocked in `poll_schedule_timeout`. Both that
and the `NOT_CRAN` question above are now recorded in
`notes/running_the_test_suite.md`.

### The precompiled vignette and the `manec_example` fixture

`vignettes/example3.Rmd` is not regenerated here, for the reason above.

#### The fixture

`manec_example` was fitted before #302 and still holds a `gamma(5, 2.283)` entry
for `nec` with `lb` and `ub` set. `test-extrapolate.R:670`, which belongs to
#405, asserts that the "bounded above" message fires on a fit derived from it,
and the round-trip assertion in `test-get_priors.R` reads its bounds. Rebuilding
that fixture breaks both together. This is older than either change and is not
#393's to fix, and editing a file under review in #405 from this branch would be
worse; recorded here so it is visible when the fixture is next regenerated.

### Stan's fallback initialisation

Where `make_good_inits()` exhausts its budget it hands the fit to Stan's own
initialisation, which draws uniform(−2, 2) on the unconstrained scale and
applies the declared constraint. A `nec` declared `<lower=a, upper=b>` therefore
started inside the tested series; one declared `<lower=0>` starts in
(0.135, 7.39) whatever the units of the series, and an unconstrained one — the
`normal` branch — starts in (−2, 2). A logged series spanning `log(10)` to
`log(10000)` now gets a fallback below every dose. The fallback is reached no
more often than before, and rather less: at the audit's 200-round cap the count
fell from 27 cells to 21, and a fit takes 10,000. What changed is where it
starts when it is reached. Recorded at the
fallback in `R/inits_functions.R`; supplying `init` through `bnec()` is the
remedy and the message there already names `get_priors()`.

### Behaviour changed for a stored fit

A fit made before this release keeps the bounds it was fitted with, and
`get_priors()`, `pull_prior()` and `update()` return them unchanged. `amend()`
rebuilds a prior only for an equation it adds, so adding one to such a fit
leaves a model-averaged set in which some equations have a truncated `nec`
posterior and some do not, and `ne_prior_bounds()` then reports the tightest
bound across the set. `NEWS.md` says so and advises refitting rather than
amending where the two halves must be on the same footing.

### Behaviour left unchanged

`bnec()`'s report on a design that has not reached its lower asymptote. The
`bot` prior is unaffected, and the threshold prior still places most of its mass
inside the tested range on the `lognormal` branch, so the design is still the
thing the priors describe.

`nsec()`, which reads its estimate off the curve and is already censored to the
prediction grid.

A fit made before this change. `get_priors()` and `pull_prior()` return what the
object stores, and `update()` reuses it.

### Rejected alternatives

Replacing `max(prior_predictor)` with a multiple of it. That puts an invented
number into the output where the truncation at least came from the design, and
it does not identify the threshold either; it changes which value the posterior
piles against.

Requiring the user to supply the bound on such a design. That makes the common
case an error, and a user-supplied bound is already available through the
`prior` argument.

Keeping the truncation and reporting the estimate as censored, which is what PR
#387 did. It leaves the posterior itself wrong, because a truncated prior does
not merely bound the estimate, it concentrates mass at the bound, so the
interval is narrow as well as displaced.

</details>


beckyfisher:
## The review findings that were real

Two rounds of review against this branch. Listing only the findings that were
real defects and are fixed, with the commit that fixed each. Everything else
raised was either already correct or declined with a reason, and those are not
repeated here.

### Wrong statements about the change's own behaviour

#### The prior width, stated at the wrong scope twice

It is a property of the branch *and* of `prior_type`, not of the branch alone.
First the roxygen said the threshold prior "places most of its mass inside the tested range" without qualification,
which is false where the predictor is supplied already logged: the
`"uninformative"` entry on `log(herbicide$concentration)` is
`normal(2.30, 31.5)` and places 0.116 of its mass inside the tested range and
0.442 above the highest concentration. Then the correction said "the normal
branch is far wider", which is false too, because `predictor_prior()` applies
the `"regularizing"` coverage rule outside its `is_logged` branch: the
`"regularizing"` entry on the same data is `normal(2.30, 1.98)`, 0.980 inside
and 0.010 above. The wide case is one cell of four.

| branch | `prior_type` | inside | above |
|---|---|---|---|
| `lognormal` (`nec_data`) | `"uninformative"` | 0.755 | 0.220 |
| `lognormal` (`nec_data`) | `"regularizing"` | 0.811 | 0.179 |
| `normal` (`log(herbicide$concentration)`) | `"uninformative"` | 0.116 | 0.442 |
| `normal` (`log(herbicide$concentration)`) | `"regularizing"` | 0.980 | 0.010 |

The table is now written out once, in `predictor_prior()`'s roxygen, and
`NEWS.md`, `?bnec` and `vignette("example3")` point at it. The vignette's advice
to supply a prior on the logged branch is scoped to `"uninformative"`, where it
is true. Issue #393 and specification §3.2 both state "roughly one part in
forty" without a qualifier; it belongs to the three coverage-rule cells, where
it is a lower bound rather than an approximation. `e869e3b8`, `8d67b564`.

#### The `NEWS.md` entry against the audit

The bullet reported
720 cells at a prior CDF of exactly 1 where the audit reports 1,439; said "no
complete-design cell changes" where the audit records a fall in all 1,440; and
said the median proposal count is unchanged where it rises from 6 to 7 on
`nec4param`. That sentence was written from the issue's definition of done
before the audit ran. `e869e3b8`.

#### Half a mechanism for the complete-design CDF fall

The fall was attributed to the upper tail returning to the normalising constant,
which is signed only where the prior CDF at the old lower bound is zero. On the
`crf(log(x))` cells, where the largest falls are, the lower bound was binding
too and the two terms oppose each other. `notes/prior_audit.md` part 4 now gives
the difference as (*u*(1 − *b*) − *a*(1 − *u*)) / (*b* − *a*) and says the
observed fall in all 1,440 cells is a measurement, not a consequence of the
algebra. `e869e3b8`.

#### Truncated sentences on the `?bnec` man page

`DESCRIPTION` sets `Roxygen: list(markdown = TRUE)`, and `%` is not an escapable
character in markdown, so `\%` in a roxygen block is a literal backslash
followed by a percent; roxygen2 then escapes the percent itself and emits
`\\%`. That is valid Rd, so `tools::checkRd()` passed it and so did four
`R CMD check` platforms. But a bare `%` opens a comment and the rest of the line
is discarded on render, so five sentences of `?bnec` ended mid-clause. One of
them was "at the 98% level for `"regularizing"` on either, so the mass above the
highest concentration", the clause stating the very fact the two rounds above
were spent getting right. Verified on a throwaway package: source `98\%`
generates `98\\%` and source `98%` generates `98\%`. All six occurrences in
`R/bnec.R` are now bare. `check_fit.Rd` and `nassarius.Rd` have the same defect
on `predev` and are not this change's to fix. The rule this leaves behind is
that `checkRd()` validates and only rendering reveals. `2f1d22f3`.

### Documentation the sweep missed

Eight further places asserted the truncation, in four files, including two
user-facing messages and one exported man page:

- the message `check_response_flattened()` prints, whose docstring twelve lines
  above it had been corrected in the same commit;
- `unscaled_power_message()` and `bounded_power_drops()` in `R/check_models.R`;
- `R/models.R` and `man/models.Rd`, which is exported and is the most visible;
- `predictor_prior()`'s and `define_group_prior()`'s own roxygen, in the file
  the change was made in.

Figures in those blocks that were computed under the truncation are attributed
to the entry as it was when they were measured rather than recomputed. Two of
them stated the rescaling factor as the mass above the highest dose where it is
one minus that mass. `2f8112f9`, `e869e3b8`, `8d67b564`.

### Tests

#### A new test that passed on the released code

It read the prior *string*, which this change does not alter; the truncation lived in `lb` and
`ub`. It now asserts both bounds and pins the truncated CDF at the truth to
0.9795, the value the audit reports for that cell. With `ub` reinstated at 40
the assertion fails, which is what makes it load-bearing. `e869e3b8`.

#### The round-trip assertion in `test-get_priors.R`

`test-get_priors.R`'s round trip compared `get_priors()` output
against a `brmsfit`'s stored `$prior`, where an absent bound is `""` on one side
and `NA` on the other — which is the difference that test exists to assert. It
now normalises with `blank_bounds_to_na()` first and has a guard against
becoming vacuous when `manec_example` is rebuilt. `e869e3b8`.

#### An unpinned message string

The advice sentence in `check_response_flattened()`'s report is the one fragment of that message
`test-check_response_flattened.R` did not assert, among sixteen it does, and it
is the fragment that went stale. Now pinned. `8d67b564`.

### Consequences that needed recording rather than fixing

#### Stan's fallback initialisation

It is no longer scale-aware on the logged branch. Where `make_good_inits()`
exhausts its budget the fit falls back to
Stan's default, which draws uniform(−2, 2) on the unconstrained scale. A `nec`
declared `<lower=a, upper=b>` started inside the tested series; `<lower=0>`
starts in (0.135, 7.39) whatever the units, and an unconstrained one in (−2, 2).
A logged series spanning `log(10)` to `log(10000)` now gets a fallback below
every dose. The fallback is reached no more often than before — at the audit's
200-round cap the count fell from 27 cells to 21 — and what changed is where it
starts. Recorded at the fallback; a scale-aware fallback is a change to
`make_good_inits()`'s contract and needs its own measurement. `e869e3b8`.

#### A mixed set from `amend()`

`amend()` on a fit made before this change produces a set in which some
equations have a truncated `nec` posterior and some do not, because it rebuilds
a prior only for an equation it adds. `NEWS.md` says so and advises
refitting. `e869e3b8`.

#### The unmet acceptance criterion

Issue #393 asks for no rise in the initial-value search; the rise is
+1.31 proposals [0.59, 2.04] on complete
`nec4param` designs. Specification §3.4 asks instead that a rise be quantified
rather than assumed absent, which is what the measurement does. Stated above the
fold in the body rather than left to be discovered. `e869e3b8`.

### Corrections to the plan rather than to the code

Specification §3.4 asks for the complete-design prior CDF to be unchanged. It is
not, and the reason is a property of the statistic: `p_truth` is the CDF of the
*truncated* prior, so removing a bound changes its normalisation on every
design, complete or not. What a complete design guarantees is that the truth is
inside the support, which holds before and after. And §3.2's "roughly one part
in forty of the prior mass sits above the highest concentration" is a lower
bound belonging to the three coverage-rule cells, not an approximation holding
everywhere. Both are recorded in `notes/prior_audit.md` part 4 rather than
propagated.

### An unclaimed fix

A predictor with a single distinct value used to hang the initial-value search
rather than fail it: the two-bound branch of `make_inits()` ran
`while (draw <= min(bounds) | draw >= max(bounds))`, a tautology when the bounds
are equal. With `lb = 0` and no `ub` that branch is unreachable from the default
priors. Pinned by a test, with a comment saying which assertion fails fast,
because a regression there hangs the suite rather than failing it. `e869e3b8`.

Separately, a constant predictor still stops in `check_data()` with "missing
value where TRUE/FALSE needed", from a comparison of the response means of the
two halves of the predictor. That refusal predates this change and is not fixed
here.

Issue #400, opened 2026-09-19
A bounded response at the bound in every observation gives NaN priors and an unattributable error


beckyfisher:
## What

`response_link_scale()` returns `NaN` for every element where every observation
of a bounded response equals 1. The fit then ends in `quantile.default()` with a
message that names neither the column nor the cause.

## Why it matters

A user meets this as a failed fit rather than as a refusal. The error is
"missing values and NaN's not allowed if 'na.rm' is FALSE", raised from inside
prior construction, and nothing in it points at the response, at the family, or
at what to do.

The data that produce it are ordinary. A `bernoulli` endpoint whose true mean is
0.95 at every dose gives a draw with no zero in it often enough to be met in
practice, and any bounded response recorded as a proportion where every replicate
scored 1 does the same.

## Evidence

`R/helpers.R:488` and `:519` compute `max(response[which(response < 1)])`. Where
no observation is below 1 that subset is empty, `max()` of it is `-Inf`, and
`linear_rescale()` then returns `NaN` for every element. `define_prior()` stops
on the first `quantile()` call that meets them.

Found by the default-prior audit of #391, on `dev` at `eff06ea9`: nine of 5,760
cells reached it, all `bernoulli`, all at the two flattest response settings.
`notes/prior_audit.md` records it under "Defects the run exposed".

<details>
<summary>Implementation detail</summary>

### The call sites

Both compute the same quantity, the largest observation strictly below the upper
bound, so that a response at the bound can be placed just inside it. The guard is
the same at both: where the subset is empty there is no such observation, so the
rescaling has nothing to anchor on.

The lower bound has the mirror case, a response where every observation is 0,
and it should be checked at the same time rather than left to be found later.

### The treatment

A refusal, not a silent fallback. A bounded response with no variation at all
does not identify a concentration-response curve, and a prior built from a
fabricated anchor would hide that. The refusal names the response column, says
that every observation is at the bound, and says which bound.

Raise it where the property is fixed for the whole call rather than per model,
following the rule in `bayesnec/CLAUDE.md`: in `bnec()` before the model loop,
beside `check_normalisation()` and `check_inline_boundary()`, and in
`bnec_group()` before its level loop. Keep a backstop in the prior construction
path for the routes that do not come through `bnec()`, which are `get_priors()`
and `amend()`.

### Tests

- A bounded response where every observation is 1, and one where every
  observation is 0, each refused by name through `bnec()`.
[...]

PR #407, opened 2026-09-22
A prior set for designs that do not reach the lower asymptote (#394)

beckyfisher:
## What

`bnec()`, `get_priors()` and `amend()` gain `asymptote_observed`, a declaration
of whether the highest predictor level reached the lower asymptote of the curve.
The default `TRUE` leaves every generated prior entry unchanged. `FALSE` spans
the central 95 per cent of the `bot` prior from the floor of the response to the
mean response at the highest predictor level, and extends the initial-value band
to the same floor.

Phase 7 of #386 and the last of the seven. Closes #394. Targets `predev`.

## Why it matters

Both default prior sets place `bot` at the observed response over the highest
concentrations. On a design that stopped before the asymptote that location sits
above the true value, and the spread narrows as the response flattens, so the
prior is at its narrowest exactly where `bot` is least identified. The fit
completes and `summary()` reports a credible interval with nothing recording
that the prior placed the true value outside its own central mass.

Selecting a prior type does not correct it. On the flattest of #386's designs the
prior CDF at the true `bot` is 6.43e-14 under `"uninformative"` and 2.30e-90
under `"regularizing"`. The roxygen at `R/define_prior.R` that named
`"uninformative"` as the set to use on such a design is corrected here.

## Evidence

`notes/scripts/prior_audit.R` gains a fifth part, `declaration`. Every cell is
run twice from one simulated response, under the default and under the
declaration, so the two arms are paired cell by cell and the difference is the
declaration alone. The factorial is 4 completeness settings x 5 designs x 3
predictor transforms x 12 families x 2 prior types, on `nec4param` and the
identity link: 1,440 cells and 2,880 prior builds. R 4.6.1, brms 2.23.0,
2026-09-22. `f92`, `f36` and `f02` name the fraction of the span from `top` to
`bot` the mean response has travelled at the highest predictor value.

Mean prior CDF at the true `bot`. A value near 0.5 means the prior is centred on
the truth:

| prior type and arm | complete | f92 | f36 | f02 |
|---|---|---|---|---|
| `uninformative`, default | 0.1804 | 0.0865 | 0.0444 | 0.0420 |
| `uninformative`, declared | 0.9273 | 0.7830 | 0.2178 | 0.1184 |
| `regularizing`, default | 0.3148 | 0.1642 | 0.0011 | 0.0009 |
| `regularizing`, declared | 0.9174 | 0.7951 | 0.2218 | 0.1204 |

Cells placing the true `bot` outside the central 95 per cent of its own prior,
out of 180 per prior type per setting:

| prior type and arm | complete | f92 | f36 | f02 |
|---|---|---|---|---|
| `uninformative`, default | 0 | 0 | 15 | 15 |
| `uninformative`, declared | 56 | 11 | 0 | 0 |
| `regularizing`, default | 1 | 26 | 179 | 176 |
| `regularizing`, declared | 41 | 10 | 0 | 0 |

The declaration removes the exclusion entirely at `f36` and at `f02`. At `f92`,
where the design is 8 per cent short of the asymptote, it leaves 11 and 10
cells, against 0 and 26 under the default; and at the complete setting it
introduces one, where the declaration is a false statement about the design.
That is the trade it makes, and it is the user's to make: the nearer the design
came to the asymptote, the nearer the endpoint mean is to `bot`, so a prior
whose central 95 per cent ends at that mean puts the truth towards the top of
its own interval.

The realised span of the declared entry, as multiples of the endpoint mean it is
built from, median over the 120 cells of each family. The rule asks for 0 and 1:

| family | 2.5th / endpoint | 97.5th / endpoint |
|---|---|---|
| `gaussian` | 0.0000 | 1.0000 |
| `bernoulli`, `Beta`, `binomial`, `beta_binomial`, `zero_inflated_beta` | 0.0084 to 0.0087 | 1.0000 |
| `Gamma`, `poisson`, `negbinomial`, `hurdle_gamma`, `zero_inflated_poisson`, `zero_inflated_negbinomial` | 0.0606 | 1.3929 |

The initial-value search, paired by cell, with both arms given the declared prior
so that what differs is the band. Median proposals to a full set of four chains,
and the paired mean difference of declared minus default:

| setting | default | declared | paired mean difference |
|---|---|---|---|
| complete | 7 | 6 | −1.18 [−1.57, −0.80] |
| f92 | 6 | 6 | −2.83 [−5.40, −0.27] |
| f36 | 7 | 6 | −2.19 [−2.84, −1.54] |
| f02 | 10 | 7 | −4.58 [−5.53, −3.63] |

The default path is unchanged. Parts 1 and 2 of the audit were run on this branch
and on `predev` at `5fee2e9a`, with the script itself identical between the two
runs and every cell seeded from its own index, so the two are paired cell by
cell. All 17,262 parameter-prior rows agree in every column: the prior string,
the bounds, the truncated CDF at the truth, the response statistics, the
initial-value proposal count and the flatness result. The printed summary tables
are identical line for line, and the fallback count of 21 of 5,751 cells
reproduces the figure part 4 records for #393. The test suite also pins 48 prior
strings captured from `predev` and compares them as text, over eight families,
both prior types and the `top`, `bot` and `nec` entries.

`notes/prior_audit.md` part 5 holds the tables in full, with the per-family
breakdown and the four cells the run could not build (#400, present in both
arms).

<details>

### The argument and where it goes

`asymptote_observed` is plumbed to `define_prior()` and `make_good_inits()`
through `fit_bayesnec()` and `add_brm_defaults()`, and to `get_priors()` and
`amend()` so that the entries can be inspected without fitting.
`define_hurdle_prior()` and `make_good_hurdle_inits()` pass it to both blocks,
because a design that stopped before the response block's asymptote stopped
before the survival block's as well.

It is not a third `prior_type` value, for the reason §5.2 records: `prior_type`
selects how strongly to regularise a design assumed complete, and whether the
design is complete is an orthogonal question. A user with an incomplete design
keeps whichever set they chose for `top`, `nec`, `ec50` and the group-level
entries. The `prior_type ==` control flow was read before anything was touched:
`R/define_prior.R` has `if (prior_type == "regularizing") ... else 1` branches
where the `else` means `"uninformative"`, and the new entry is written after both
branches rather than into either.

### The `bot` entry

One rule on every family: the central 95 per cent spans from the floor to the
mean response at the highest predictor level, *e*. The endpoint mean is an upper
bound on `bot` on such a design rather than an estimate of it, so it becomes the
top of the interval instead of its centre.

- gaussian: `normal(e/2, e/(2 * qnorm(0.975)))`, exactly `[0, e]`.
- beta: `beta(1, log(0.025) / log(1 - e))`, whose 97.5th percentile is exactly
  *e*.
- gamma: `gamma(2, 4/e)`, the `gamma(2, 2/q)` idiom at `q = e/2`.

*e* is `regularizing_location()`'s `"bot"` anchor, so the two prior sets, the new
entry and the initial-value band all read the endpoint the same way and the
treatment of zeros recorded there applies unchanged.

The cap in `regularizing_entry()` does not apply, because this is a separate
construction rather than a fourth branch of that function: the rule fixes both
ends of the interval, so the width follows from the endpoint mean rather than
from the width of an entry built on a different location.

The cap would bind on five of the eight families and not on the other three.
Standard deviations of the `bot` entry on the flattest design, `"uninformative"`
against declared: gaussian 0.107 against 0.220, `beta` 0.160 against 0.253,
`bernoulli` 0.160 against 0.297, `binomial` and `beta_binomial` 0.160 against
0.256. On the gamma branch the declared entry is narrower — `Gamma` 0.558
against 0.297, `poisson` 23.4 against 13.1, `negbinomial` 15.5 against 8.84 —
and narrower by exactly a half by construction, since a gamma at shape 2 has a
standard deviation of its mean over the square root of two and the rule halves
the mean. Narrower there is not worse: what the entry changes is where the mass
sits. A test asserts the gaussian case, which is the one the cap would bind on.

### The beta branch and the mechanism the specification suggested

§5.3 names `beta_from_mode_sd()`. That helper requires both shapes above 1, so
the density is zero at the floor. Applied with the mode at *e*/2 and the gaussian
branch's spread it returns a 2.5th percentile of 0.18*e*, which sat above the
true `bot` in every one of the 75 bounded cells at `f36`, and again at `f02`,
under each prior type — where the fixed `beta(2, 5)` of the `"uninformative"` set
excludes the truth in none of them. That measurement is what changed the
mechanism: fixing `shape1` at 1 leaves `shape2` determined by the 97.5th
percentile alone and excludes the truth in none of those cells. The governing
sentence of §5.3 is the contract and the helper was the mechanism it suggested.
Both runs are recorded in `notes/prior_audit.md` part 5.

### The floor and the responses that have none

`asymptote_floor()` is one function, read by `define_prior()` and by
`make_good_inits()`, so the prior and the band cannot disagree about where the
bottom of the curve is. It returns zero for every `prior_family_tag()` but
`gaussian`, and zero for a gaussian response that is non-negative throughout
under an identity link.

Two cases return `NA` and are refused, naming `bot` and the `prior` argument.
A gaussian response spanning negative values has no floor in its own units, which
§5.3 states; a log ratio or a growth increment is ordinary input here (#229).
A non-identity link is beyond §5.3's table and is called out here: a `log` or
`logit` link is rewritten onto the gaussian entries by `prior_family_tag()`, and
on that scale the floor of the mean is `-Inf`, so a response that happens to be
non-negative once logged states only that every observation exceeds one in
response units, which is a property of the units. The audit confirms all twelve
families are refused under their alternative link.

### The placement of the refusal

`bnec()` before its model loop and `bnec_group()` before its level loop, with
`define_prior()` keeping it as the backstop for `get_priors()` and `amend()`.
Both refusals the declaration can raise are hoisted: no derivable floor, and a
response holding no value above that floor.
This follows the rule in `CLAUDE.md`: from inside the model loop the refusal is
printed once per equation and the call ends on the generic all-models-failed
advice, and from inside the level loop it arrives only after the earlier levels
have compiled and sampled. Measured at 0.04 s to refuse a two-equation call, so
nothing is compiled. `bnec()` sets a private `.bayesnec_asymptote_checked`
marker, as it already does for the flatness report, so `bnec_group()` raises both
once over the whole response.

The refusal is raised only where a `bot` prior will be generated from the
declaration. `add_brm_defaults()` wraps the default build in `try()` so that a
user who supplied a complete set is never blocked by a default they will not use
(#207, #229), and a check before the loop would take that back, so it is gated
on `uses_response_range_defaults()` reading the `bot` row alone — the same test
that gates the flatness report above it. `define_prior()`'s backstop is gated
the same way, on whether the equation has a `bot` parameter at all. The
initial-value band is unaffected either way: with no floor to extend it to,
`make_good_inits()` leaves it where it was.

`check_asymptote_declaration()` builds the response the way `fit_bayesnec()` does
— divided by the trials or the `rate()` denominator, then on the link scale — and
reads it from the model frame, before `check_data()` has nudged a zero or a one
off a boundary. The two agree because a nudge leaves a value within its own
support and both divisors are positive, so neither changes the sign of the
response, which with the link is all the floor depends on.

### The initial-value band

`init_limits()` gains `floor_value`, applied through `min()` rather than by
assignment. The declaration states that the true asymptote may be lower than the
response shows and never that it is higher, so it may only widen the band: on a
gaussian response the spread already places the lower limit below zero, and
assigning the floor there would narrow the band and reject draws the released
search accepts. The upper limit is untouched.

The test for this is deterministic. It scores one fixed set of 200 proposals
against the two bands and asserts the declared one accepts more, rather than
comparing proposal counts from two searches: `refine_inits()` and the per-round
redraw of the empty chains alone both consume random numbers, so the two arms
diverge after the first acceptance differs and the round count is not monotone in
the width of the band. The audit supplies the effect on the search itself.

### The model set

Fourteen of the 23 equations have no `bot` parameter and so assert that the
response falls to the floor. A set holding both kinds is reported and left alone,
with the message naming both branches, because whether the response can reach the
floor is a property of the endpoint rather than of the data.

### Tests

`tests/testthat/test-define_prior.R` gains: the span rule asserted per branch on
eight families and both prior types, with the endpoint anchor pinned to the mean
at the highest predictor value; the coverage measurement on #386's flattest
design; the cap not binding on the gaussian branch; every refusal, including the
count of observations below zero; the bit-identity guard; the refusal raised
before `bnec()`'s model loop; the mixed-set report and its silence on a set on
one side of the divide; `chk_flag()` validation; the prior `brm()` is handed;
`amend()` through both its `define_prior()` and `fit_bayesnec()` calls; the
hurdle route on both blocks; a zero-inflated count family; a `rate()`
denominator; the supplied-prior gate; and the flatness report's silence where
the declaration would be refused. `tests/testthat/test-inits_functions.R` gains
the band tests above and the `add_brm_defaults()` plumbing.
`tests/testthat/test-bnec_group.R` gains the refusal before the level loop, an
abbreviated argument name, and the validation of an invalid declaration.

### Changes beyond the specification

`R/check_fit.R` and `R/data.R` wrote `\%` in roxygen under
`Roxygen: list(markdown = TRUE)`, which roxygen writes to the `.Rd` as `\\%`.
That is a literal backslash followed by an Rd comment, so three sentences were
truncated mid-line in the rendered documentation — `?check_fit` read "with a 95\
of". This is the defect #393 fixed in `man/bnec.Rd`, found in two more files.
`git grep '\\\\%' -- man/` is now clean.

The flatness report of #390 gains a closing sentence naming `asymptote_observed`,
which is the action it exists to prompt.

</details>


beckyfisher:
## Review round

The pull request went through five rounds with an independent reviewer. The body
above has been updated to match the current state: the `f92` counts, the reason
the `regularizing_entry()` cap does not apply, and the tests added since the
first push. The commits that followed the first review are in the branch
history.

These are the findings that were real and what each one changed.

1. The reason for not applying the `regularizing_entry()` cap was false for
   three families. The roxygen said the declared entry is wider than both
   released sets by construction. On the flattest design the declared standard
   deviation is 0.297 against 0.558 for `Gamma`, 13.1 against 23.4 for
   `poisson` and 8.84 against 15.5 for `negbinomial` — narrower by exactly a
   half, because a gamma at shape 2 has a standard deviation of its mean over
   the square root of two and the rule halves the mean. The decision stands;
   the reason is now the shape of the construction rather than its width, and
   the measurement is tabulated.

2. The refusal blocked a user who had supplied a complete prior.
   `add_brm_defaults()` wraps the default-prior build in `try()` so that such a
   user is never blocked by a default they will not use (#207, #229), and the
   check raised before the model loop took that back. It is now gated on the
   `bot` row alone, and `define_prior()`'s backstop on whether the equation has
   a `bot` parameter.

3. A supplied prior then silenced the mixed-set report as well, which was a
   defect introduced by the fix above. Whether a set mixes equations that
   estimate a lower asymptote with ones that do not is a property of the set and
   the design, so the report is now raised before the gate.

4. One of the two refusals was not raised before the model loop. A response with
   no value above the floor was refused from inside `define_prior()`, which on a
   model set prints it once per equation. It is now raised with the other.

5. An abbreviated argument name defeated that placement. `bnec_group()` read
   `asymptote_observed` from its dot list by exact name while `do.call()`
   matches an abbreviation, so `asymptote = FALSE` was declared for the fit and
   undeclared for the check. A `dots_arg()` helper now resolves
   `asymptote_observed`, `prior`, `model_survival` and `loo_controls` the way
   the forwarded call will. The last two were already wrong before this pull
   request.

6. The flatness report recommended a declaration that would be refused. The
   sentence naming `asymptote_observed = FALSE` was unconditional, so a gaussian
   response spanning negative values, or a non-identity link, was told to
   declare and would then have met a hard error.

7. "Assert that the response falls to the floor" is false for three of the
   fourteen. `neclin`, `neclinhorme` and `ecxlin` decay by subtraction and are
   unbounded below, and `check_models()` keeps all three for a gaussian family,
   which is the branch where the floor comes from the response being
   non-negative. The message, `?bnec` and `NEWS.md` now say the fourteen have no
   lower asymptote to estimate.

<details>
<summary>The smaller findings, and the ones answered without a change</summary>

Also fixed: the floor refusal now counts the observations below zero and reports
the smallest, so a single stray reading is distinguishable from a response that
is negative throughout; `?bnec` and `NEWS.md` describe the endpoint anchor as
the end of the predictor series rather than the highest predictor level, which
is what it is on a sparsely replicated design; the roxygen that credited a tenth
of the smallest positive observation to the released `"uninformative"` gamma
entry now records that the entry uses a hundredth of it, on the spread; the beta
branch tests its endpoint before taking a logarithm rather than its shape after
it; `unobserved_endpoint_mean()` tests for a positive value before reading the
anchor rather than wrapping the read in `try()`; the prior widths quoted in the
roxygen name the fixture they were measured on; and `man/bnec_group.Rd` was
deleted by a misplaced roxygen block in one of these commits and restored,
byte-identical to `predev`'s.

Tests added over the round: the hurdle route on both blocks, a zero-inflated
count family, a `rate()` denominator, the supplied-prior gate, the mixed-set
report under a supplied prior, the hoisted endpoint refusal, the count of
observations below the floor, the flatness advice, the endpoint anchor pinned to
the mean at the highest predictor value, and `bnec_group()` with an abbreviated
and with an invalid declaration.

Answered without a change: the gaussian entry keeps `lb = NA`, because the rule
is stated on the central interval and a hard bound at zero would be a stronger
claim than the declaration makes; the argument was inserted mid-signature as
`prior_type` and `predictor_scale` were in earlier phases, and none of the three
is on CRAN; and `mod_groups$bot_free` reading as "has a free `bot`" is worth its
own issue rather than a rename here.

</details>

